PR #17, opened 2018-10-23
Fix #16 and add unit test

### LINKED ISSUE #16: Could not find super type undefined when no system model
If you use the ModelManager without a system model specified you get a failure.

IllegalModelException: Could not find super type undefined File './models/domain.cto': line 9 column 1, to line 17 column 2.

mttrbrts:
Fixes #16 . 

This allows users to add Concerto to their project without specifying a system model. This implies that all base types (Transaction, Event, Asset & Participant) will need to specify an identifying field when a system model is not provided


dselman:
Merci!

PR #41, opened 2019-06-24
(feat) update dependencies and add a filename to the BaseFileException

dselman:

- Update the dependencies (also moved some to be devDependencies to slim things down).
- Added a fileName property to the BaseFileException for better error reporting inside VSCode


jeromesimeon:
With this PR, I _think_ this line should also go away?

https://github.com/hyperledger/composer-concerto/blob/7b89bae01b4cbcbad6a953bfba4c20eb5a7d931e/lib/introspect/parseexception.js#L57

dselman:
> With this PR, I _think_ this line should also go away?
> 
> https://github.com/hyperledger/composer-concerto/blob/7b89bae01b4cbcbad6a953bfba4c20eb5a7d931e/lib/introspect/parseexception.js#L57

You are right - that doesn't seem wonky. I'd forgotten that fileLocation includes the fileName.

dselman:
> With this PR, I _think_ this line should also go away?
> 
> https://github.com/hyperledger/composer-concerto/blob/7b89bae01b4cbcbad6a953bfba4c20eb5a7d931e/lib/introspect/parseexception.js#L57

It looks like this is still needed - it copies the filename into the file location if not specified.

PR #45, opened 2019-07-05
errors(models) remove modelfile from IllegalModelException

jeromesimeon:

## Checklist
 - [ ]  A link to the issue/user story that the pull request relates to
 - [ ]  How to recreate the problem without the fix
 - [x]  Design of the fix
 - [x]  How to prove that the fix works
 - [ ]  Automated tests that prove the fix keeps on working
 - [ ]  Documentation - any JSDoc, website, or Stackoverflow answers?

## Issue/User story

Exception returned by the API should not carry the whole model if possible. This removes the `modelFile` field from `IllegalModelExceptions`. `fileName` is used to retrieve the model file when needed instead.

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)
- [ ] [GitHub Issues](https://github.com/hyperledger/composer/issues)

## Validation of the fix

All tests are passing as before. At least one test uses access to the model file from their name when an illegal model error occurs.

### DIFF
--- a/lib/introspect/illegalmodelexception.js
+++ b/lib/introspect/illegalmodelexception.js
@@ -56,15 +56,6 @@ class IllegalModelException extends BaseFileException {
         messageSuffix = messageSuffix.charAt(0).toUpperCase() + messageSuffix.slice(1);
 
         super(message, fileLocation, message + ' ' + messageSuffix, fileName, component);
-        this.modelFile = modelFile;
-    }
-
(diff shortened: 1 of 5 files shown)


mttrbrts:
LGTM @jeromesimeon

PR #49, opened 2019-07-29
 refactor(scripts): Move dev scripts from runtime lib folder

### LINKED ISSUE #46: Move `parsejs` and dependencies to the `./scripts` directory
As far as I can tell, `./lib/codegen/parsejs.js` is only used as a developers dependency for generating the API documentation. It should be moved to a build-time directory such as `scripts`.

## Context

There are a number of dependencies e.g., `acorn` and `doctrine` which are not needed at runtime, but are placed in the `./lib` directory. This means any consumer of the `composer-concerto` package must also have those.

They also get bundled in through WebPack unnecessarily.

## Expected Behavior

Same behaviour as not but without the dependencies to `acorn` and `doctrine`, etc.

## Actual Behavior

Some current build-time dependencies are made necessary at runtime.

## Possible Fix

- Move the following: `./lib/codegen/parsejs.js` and `./lib/codegen/javascriptparser.js` to `./scripts`
- Move `acorn` and `doctrine` from `dependencies` to `devDependencies`.

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [ ] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)

This led to issue #43 

## Context

Forced to add unnecessarily `acorn` and `doctrine` upstream in a consumer of Composer Concerto (Accord Project's Cicero and related packages).

## Your Environment

* Version used: 0.71.6
* Link to your project:

mttrbrts:
Resolves #46 

- Moves modules that are only needed at design time to scripts folder. 
- Also moves associated dependencies to devDependencies


mttrbrts:
@jeromesimeon

test/codegen/javascriptparser.js
```diff
@@ -14,7 +14,7 @@
 
 'use strict';
 
-const JavascriptParser = require('./../../lib/codegen/javascriptparser');
+const JavascriptParser = require('./../../scripts/codegen/javascriptparser');
```
dselman: I think this is used at runtime by Cicero to introspect the generated JS code and to grab annotations on it.

jeromesimeon: Good catch. It's used right here (inherited from the ScriptManager):
https://github.com/accordproject/ergo/blob/4597c52f59f2824ced0a27410d66634a42065cfe/packages/ergo-compiler/lib/script.js#L18

I am unclear why this is part of Concerto though. I couldn't find a specific used for it. I would be ok moving that file to the Ergo repository (eventually this could be entirely removed, since we now always use JS files generated by the Ergo compiler).

jeromesimeon: > Good catch. It's used right here (inherited from the ScriptManager):
> https://github.com/accordproject/ergo/blob/4597c52f59f2824ced0a27410d66634a42065cfe/packages/ergo-compiler/lib/script.js#L18
> 
> I am unclear why this is part of Concerto though. I couldn't find a specific used for it. I would be ok moving that file to the Ergo repository (eventually this could be entirely removed, since we now always use JS files generated by the Ergo compiler).

I take it back, it's used at least in the scripts to generate the API...

dselman: Hi Jerome. I'd love to see this code removed from Concerto as it drags in a lot of dependencies (acorn etc). We should create a new module for it so we can use it as a `devDependency` for the build system.

PR #50, opened 2019-07-29
fix(47): Adds custom instanceof implementation to allow comparison of different module instances

### LINKED ISSUE #47: Avoid `instanceof` which leads to runtime errors with deep npm dependencies
The use of `instanceof` to check whether an object is of a class leads to unreliable check when importing Composer Concerto in projects with deep npm dependencies.

## Context

Composer Concerto uses `instanceof` checks e.g., here https://github.com/hyperledger/composer-concerto/blob/37189e52c4f52d13cbd862b38d36a64022976398/lib/serializer/resourcevalidator.js#L72 or here https://github.com/hyperledger/composer-concerto/blob/37189e52c4f52d13cbd862b38d36a64022976398/lib/serializer/jsonpopulator.js#L96

Those can be unreliable since it checks whether this is the same object from the exact same package installed in a node.js project.

The reason it is unreliable is that it is difficult to enforce that the corresponding version of `composer-concerto` is installed exactly once in the `./node_modules` directory through `npm install`.

This can lead to runtime errors when doing validation/serialization in some projects with deep npm dependencies.

## Expected Behavior

A project should be able to import Composer Concerto 0.7.1 and PROJECT 0.X.X depending on e.g., Composer Concerto 0.7.0 without leading to runtime errors.

## Actual Behavior

Currently a project has to ensure _every_ dependency which import Composer Concerto have the exact same version, and that `npm install` leads to only one copy of that same version, which is difficult to guarantee (due to the way `npm install` works).

## Possible Fix

Several fixes are suggested in: https://stackoverflow.com/questions/41587865/using-instanceof-on-objects-created-with-constructors-from-deep-npm-dependenci

The recommended fix is to use an explicit method to check whether an object has the expected class. So instead of writing `thing instanceof RelationshipDeclaration` one would write `isRelationshipDeclaration(thing)` where:

```js
RelationshipDeclaration.prototype._isRelationshipDeclaration = true;
var isRelationshipDeclaration = function isRelationshipDeclaration(obj) {
  return obj instanceof RelationshipDeclaration || Boolean(obj._isRelationshipDeclaration);
};

module.exports = {
  RelationshipDeclaration,
  isRelationshipDeclaration
};
```

## Steps to Reproduce

1.
2.
3.
4.

## Existing issues

- [x] [Stack Overflow issues](http://stackoverflow.com/tags/hyperledger-composer)

## Context

Importing Composer Concerto in various parts of the Accord Project (Ergo, Cicero, Cicero template library, template studio, etc.).

## Your Environment

* Version used: 0.70.*
* Link to your project:

mttrbrts:
Resolves #47 

Adds custom `instanceof` implementation through the use of [`Symbol.hasInstance`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance) to allow comparison of different module instances.

### DIFF
--- a/api.txt
+++ b/api.txt
@@ -9,6 +9,7 @@ class AssetDeclaration extends ClassDeclaration {
    + void constructor(ModelFile,Object) throws IllegalModelException
    + boolean isRelationshipTarget() 
    + string getSystemType() 
+   + boolean undefined(object) 
 }
 class ClassDeclaration extends Decorated {
    + void constructor(ModelFile,string) throws IllegalModelException
@@ -36,10 +37,12 @@ class ClassDeclaration extends Decorated {
    + Property[] getProperties() 
    + Property getNestedProperty(string) throws IllegalModelException
    + String toString() 
+   + boolean undefined(object) 
 }
 class ConceptDeclaration extends ClassDeclaration {
    + void constructor(ModelFile,Object) throws IllegalModelException
    + boolean isConcept() 
+   + boolean undefined(object) 
 }
 class Decorator {
    + void constructor(Object) throws IllegalModelException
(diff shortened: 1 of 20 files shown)


mttrbrts:
@jeromesimeon

api.txt
```diff
@@ -9,6 +9,7 @@ class AssetDeclaration extends ClassDeclaration {
    + void constructor(ModelFile,Object) throws IllegalModelException
    + boolean isRelationshipTarget() 
    + string getSystemType() 
+   + boolean undefined(object)
```
dselman: This is bizarre....

Issue #66, opened 2019-09-17
Support for offline download of external models


jeromesimeon:
**Is your feature request related to a problem? Please describe.**
The ability to resolve and download external model to a local directory.

**Describe the solution you'd like**
A command line tool which lets people resolve external model references (with `from`) and generate files in a given directory.

**Additional context**
In addition to being a useful tool for users, some of the Accord Project stack depends on external models at build time (`markdown-transform` requires CommonMark and CiceroMark ASTs, and `ergo-compiler` requires base models for the contract and runtime).

PR #70, opened 2019-09-27
feat(core): Store remote models on file system

mttrbrts:
# Issue #66
Support for offline download of external models

### Changes
- Adds `getModels` function to `ModelManager`
- Adds `writeModelsToFileSystem ` function to `ModelManager`
- Exposes `saveExternal` function in `concerto-cli`

### Flags
- Also fixes a bug with the `api.txt` generation when `hasInstance` was listed as `undefined`

### Related Issues
- Issue #66


packages/concerto-cli/index.js
```diff
@@ -54,6 +54,28 @@ require('yargs')
                 console.log(err.message + ' ' + err);
             });
     })
+    .command('saveExternal', 'save local copies of external model dependencies', (yargs) => {
```
jeromesimeon: @dselman and I had suggested `concerto get` which I think is what the `go` compiler uses. https://golang.org/cmd/go/

Other options might be `concerto install` `concerto download` ...

packages/concerto-cli/index.js
```diff
@@ -54,6 +54,28 @@ require('yargs')
...
+        yargs.option('ctoFiles', {
+            describe: 'array of local CTO files',
+            type: 'string',
+            array: true,
+            default: '.'
+        });
+        yargs.option('outputDirectory', {
```
jeromesimeon: Some of our other tools use `--out` (not always for directories though) so maybe we could try and align?

packages/concerto-core/lib/codegen/javascriptparser.js
```diff
@@ -251,12 +251,17 @@ class JavaScriptParser {
...
+                            if(!thing.key.name && thing.key.property.name){
+                                name = thing.key.property.name;
+                            }
+
                             commentData = commentData || [];
                             if (visibility === '+' || visibility === '~' || includePrivates) {
                                 const method = {
                                     visibility: visibility,
                                     returnType: returnType,
-                                    name: thing.key.name,
```
jeromesimeon: I'm unclear about those changes, do they have to do with the `api.txt` generation?

mttrbrts: Yes, that's right, this library is used to extract the signature of the API for generating `api.txt` automatically

Issue #74, opened 2019-10-02
Concerto Core has the wrong package name


jeromesimeon:
**Describe the bug**
Wrong package name.

The structure of the lerna packages in the code is:
```
concerto-core
concerto-tools
concerto-cli
```

But the package name (and the published package on npmjs) for `concerto-core` is still `concerto`. I think it would make sense to change that to avoid confusion, and align with other Accord Project packages structure/naming.

Corresponding line, I believe:
https://github.com/accordproject/concerto/blob/9f324a35c0d2f5f50ddeb64b04a1e819e7487143/packages/concerto-core/package.json#L2

Niladri24dutta:
@jeromesimeon created a PR to update the package name

jeromesimeon:
Thanks @Niladri24dutta ! That's great. I would love to get @mttrbrts 's feedback on this issue since he did the refactoring into a monorepo.

mttrbrts:
@jeromesimeon I've left some minor comments in the PR. otherwise it LGTM 👍

jeromesimeon:
@mttrbrts Great thanks! I wanted to make sure you agree with the issue in the first place!

jeromesimeon:
A few things that should be done for this:
- Make sure all references across packages use the new name. Most of them are in the `concerto-tools` package. Here is a grep on my local copy:
```

Here is a grep of the references on my local machine:
```
bash-3.2$ grep -r require packages/*/lib | grep concerto | grep accordproject | grep -v '@accordproject/concerto-tools'
packages/concerto-cli/lib/commands.js:const ModelManager = require('@accordproject/concerto').ModelManager;
packages/concerto-cli/lib/commands.js:const ModelFile = require('@accordproject/concerto').ModelFile;
packages/concerto-cli/lib/commands.js:const DefaultModelFileLoader = require('@accordproject/concerto').DefaultModelFileLoader;
packages/concerto-cli/lib/commands.js:const FileWriter = require('@accordproject/concerto').FileWriter;
packages/concerto-tools/lib/codegen/fromcto/plantuml/plantumlvisitor.js:const ModelFile = require('@accordproject/concerto').ModelFile;
packages/concerto-tools/lib/codegen/fromcto/plantuml/plantumlvisitor.js:const ModelManager = require('@accordproject/concerto').ModelManager;
[...]

PR #75, opened 2019-10-02
Update package.json for the package name concerto-core

Niladri24dutta:
### Fixes issue #74 
Updated package name as `concerto-core` 

### Changes
- Updated package name as `concerto-core` to avoid confusion in npm, and align with other Accord Project .


Niladri24dutta:
@jeromesimeon it seems there is a build issue in the travis pipeline. Can you let me know if I need to update the package name somewhere else as well?maybe the old name is still used by some other dependent packages.

mttrbrts:
@Niladri24dutta you're exactly right, you should update the dependencies for `concerto-cli` and `concerto-tools` too. For example, https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L43

@jeromesimeon this is a breaking change so we should bump to `0.81.0` on the next release.

Niladri24dutta:
> 
> 
> @Niladri24dutta you're exactly right, you should update the dependencies for `concerto-cli` and `concerto-tools` too. For example, https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L43
> 
> @jeromesimeon this is a breaking change so we should bump to `0.81.0` on the next release.

@mttrbrts 
Sure I will update the dependencies in `concerto-tools` package and others package.json if required. I was getting a strange issue in mocha during the test case execution so I updated the unit test cases as well. But still i think this packages needs to be renamed https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L48

Niladri24dutta:
@mttrbrts  However regarding the version of `sinon` which you mentioned above. I believe it's same across all the tools.

jeromesimeon:
> > @Niladri24dutta you're exactly right, you should update the dependencies for `concerto-cli` and `concerto-tools` too. For example, https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L43
> > @jeromesimeon this is a breaking change so we should bump to `0.81.0` on the next release.
> 
> @mttrbrts
> Sure I will update the dependencies in `concerto-tools` package and others package.json if required. I was getting a strange issue in mocha during the test case execution so I updated the unit test cases as well. But still i think this packages needs to be renamed https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L48

Yes this is absolutely correct, since that would be the dependency from `cicero-cli` to the new `cicero-core`. There should be a similar dependency in `cicero-tools`.

It is possible that there is a dependency to the package name in some of the tests for error messages. I can help figure that one out if need be. (a log for this if this happens would be helpful).

Niladri24dutta:
> 
> 
> > > @Niladri24dutta you're exactly right, you should update the dependencies for `concerto-cli` and `concerto-tools` too. For example, https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L43
> > > @jeromesimeon this is a breaking change so we should bump to `0.81.0` on the next release.
> > 
> > 
> > @mttrbrts
> > Sure I will update the dependencies in `concerto-tools` package and others package.json if required. I was getting a strange issue in mocha during the test case execution so I updated the unit test cases as well. But still i think this packages needs to be renamed https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/package.json#L48
> 
> Yes this is absolutely correct, since that would be the dependency from `cicero-cli` to the new `cicero-core`. There should be a similar dependency in `cicero-tools`.
> 
> It is possible that there is a dependency to the package name in some of the tests for error messages. I can help figure that one out if need be. (a log for this if this happens would be helpful).

Yeah I already fixed that it was in this commit https://github.com/accordproject/concerto/pull/75/commits/8f9c0fc73fdeffd3b319935672fae3db5510c49a#diff-4ffd2d6fd84fee54c94ed0ab58bca17cR21

jeromesimeon:
> > Yes this is absolutely correct, since that would be the dependency from `cicero-cli` to the new `cicero-core`. There should be a similar dependency in `cicero-tools`.
> > It is possible that there is a dependency to the package name in some of the tests for error messages. I can help figure that one out if need be. (a log for this if this happens would be helpful).
> 
> Yeah I already fixed that it was in this commit
> [8f9c0fc](https://github.com/accordproject/concerto/commit/8f9c0fc73fdeffd3b319935672fae3db5510c49a)

That's the one! great 👍

Niladri24dutta:
Hi @mttrbrts  and @jeromesimeon I have made the required changes as mentioned above and in this link https://github.com/accordproject/concerto/issues/74#issuecomment-537942799 . The build passed now. Let me know if any other change is require for this PR to be merged.

PR #80, opened 2019-10-03
(feat) typescript definitions

j4m3sb0mb:
Signed-off-by: Giacomo Minighin <giacomo.minighin@gmail.com>

### Changes
- created typescript definitions

### DIFF
--- a/packages/concerto-core/package.json
+++ b/packages/concerto-core/package.json
@@ -7,7 +7,7 @@
     "npm": ">=5"
   },
   "main": "index.js",
-  "typings": "index.d.ts",
+  "typings": "types/index.d.ts",
   "scripts": {
     "prepublishOnly": "webpack --config webpack.config.js --mode production",
     "prepare": "pegjs ./lib/introspect/parser.pegjs",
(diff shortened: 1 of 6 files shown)


mttrbrts:
Nicely done, thanks @j4m3sb0mb!

j4m3sb0mb:
I've followed your js-docs to create this, some of the functions were private and so I made them private as well in concerto-core types. do you think they should be public?

mttrbrts:
I think that marking them private is the correct thing to do. If users want to extend a built-in class, then they should know that they can use private functions.

jeromesimeon:
@mttrbrts @j4m3sb0mb How are those being tested? And somewhat related: what is the process to maintain them when the API changes?

mttrbrts:
No tests at the moment, and maintained manually. 

I wonder if we could extend https://github.com/accordproject/concerto/blob/master/packages/concerto-core/lib/tools/versionchecker.js to do the check / hash? At the very least we know when the API changes already so we can warn people to update the `types/index.d.ts` file too.

jeromesimeon:
@mttrbrts sounds like a good idea to start. Would you mind opening an issue for that?

Issue #89, opened 2019-10-03
Concerto CLI fails on some Accord Project models


jeromesimeon:
**Describe the bug**
The CLI (e.g., `concerto generate`) fails on well formed and valid models from the Accord Project.

**To Reproduce**

E.g., using examples from [cicero-models](https://github.com/accordproject/models/):
```
bash-3.2$ cd ~/git/cicero-models/src/
bash-3.2$ concerto --version
0.80.4-20191003155034
bash-3.2$ concerto generate --ctoFiles signature/signature.cto 
Class ContractSigned is not declared as abstract. It must define an identifying field. File 'signature/signature.cto': line 9 column 1, to line 11 column 2.  IllegalModelException: Class ContractSigned is not declared as abstract. It must define an identifying field. File 'signature/signature.cto': line 9 column 1, to line 11 column 2. 
bash-3.2$ concerto generate --ctoFiles cicero/contract.cto 
Generated JSONSchema code.
bash-3.2$ concerto generate --ctoFiles cicero/runtime.cto 
Class Request is not declared as abstract. It must define an identifying field. File 'cicero/runtime.cto': line 13 column 1, to line 13 column 23.  IllegalModelException: Class Request is not declared as abstract. It must define an identifying field. File 'cicero/runtime.cto': line 13 column 1, to line 13 column 23. 
bash-3.2$ 
```

**Additional context**
This is most likely due to distinct systems models. See also #62

vipul-vinsol:
I wanna work on this one if no one else is working on this @jeromesimeon

vipulbhj:
I wanna work on this, can I try this if no one else is working on it @jeromesimeon

dselman:
The issue is that the `ModelManager` is created without a system model:
https://github.com/accordproject/concerto/blob/master/packages/concerto-cli/lib/commands.js#L64

In AP we use this system model:
https://github.com/accordproject/ergo/blob/master/packages/ergo-compiler/lib/apmodelmanager.js#L21

A quick fix would be to use the same system model for `concerto generate` but we really should remove system models altogether to ensure that models are completely self-describing and reusable.

vipulbhj:
If that is the case, I should patch this first with the above suggested approach and then once the bug is fixed we can decide how to remove system model completely ??

jeromesimeon:
@vipulbhj Sounds like a pretty good plan.

[...]

PR #105, opened 2019-10-05
#89 

vipulbhj:
# Issue #89 
Some first attempt changes submitted for review

### Changes
Added command option for system and used default model system as a first step

### Related Issues

### DIFF
--- a/packages/concerto-cli/index.js
+++ b/packages/concerto-cli/index.js
@@ -43,7 +43,7 @@ require('yargs')
             Logger.info(`validate sample in ${argv.format} against the models ${argv.ctoFiles}`);
         }
 
-        return Commands.validate(argv.sample, argv.ctoSystem, argv.ctoFiles)
+        return Commands.validate(argv.sample, argv.ctoFiles)
             .then((result) => {
                 Logger.info(result);
             })
@@ -78,7 +78,7 @@ require('yargs')
             Logger.info(`generate code in format ${argv.format} from the models ${argv.ctoFiles} into directory ${argv.outputDirectory}`);
         }
 
-        return Commands.generate(argv.format, argv.ctoFiles, argv.outputDirectory)
+        return Commands.generate(argv.format, argv.ctoSystem, argv.ctoFiles, argv.outputDirectory)
             .then((result) => {
                 Logger.info(result);
             })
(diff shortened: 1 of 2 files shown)


packages/concerto-cli/index.js
```diff
@@ -20,13 +20,18 @@ const Commands = require('./lib/commands');
 
 require('yargs')
     .scriptName('concerto')
-    .usage('$0 <cmd> [args]')
+    .usage('$0 <cmd> [manager] [args]')
```
jeromesimeon: This change to `usage` is inconsistent with the changes below which use a `--ctoSystem` option as part of the arguments.

packages/concerto-cli/index.js
```diff
@@ -20,13 +20,18 @@ const Commands = require('./lib/commands');
...
     .command('validate', 'validate JSON against model files', (yargs) => {
         yargs.option('sample', {
             describe: 'sample JSON to validate',
             type: 'string',
             default: 'sample.json'
         });
+        yargs.option('ctoSystem', {
+            describe: 'system model to be used',
+            type: 'string'
+            default: 'org.accordproject.base.cto'
```
jeromesimeon: I'm unclear how the default can be a namespace, I was expecting it to be consistent with the option which should pass a file (or maybe a URL or either?) in a way similar to how other CTO files are passed on the command line.

Without a default, the value will be null which might be sufficient to decide whether a system model has been passed or not.

jeromesimeon:
Hi @vipulbhj Any progress on this? Any other help you might need? We've had a couple of requests for this fix 🙂

vipulbhj:
Hey, sorry for being unresponsive. Its a religion week in india right now celebrate at country scale, so I been a little occupied. I would love to get all the help I can for sure. Thank you so much :) and I be getting back to constant communications in a day or two at max

jeromesimeon:
@vipulbhj I need to push the timeline for this forward since a user is waiting for a fix. I've rebased your change to https://github.com/accordproject/concerto/tree/js-vi-fix-issue-89 which includes your initial code, and I'll be working in this. This should preserve your commit for posterity (and #hacktoberfest).

I'm closing this PR and will open a new one for #89

Issue #158, opened 2019-08-16
Serialization of Infinity is null


mttrbrts:
**Describe the bug**
When an expression sets a value in the state to `Infinity`, this gets serialized (through `JSON.stringify`?) to `null`. This has the effect of invalidating the state instance and rendering the contract unable to accept new requests.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description
> The numbers Infinity and NaN, as well as the value null, are all considered null.

**To Reproduce**
infinity.ergo
```js
namespace matt.inf

import org.accordproject.cicero.contract.*
import org.accordproject.cicero.runtime.*

contract Inf over InfContract {
    clause infTest(request: Request): Response {
        set state InfContractState {
            stateId: "inf0",
            value: 1.0 / 0.0
        };
        return Response{}
    }
}
```

model.cto
```
namespace matt.inf

import org.accordproject.cicero.contract.* from https://models.accordproject.org/cicero/contract.cto
import org.accordproject.cicero.runtime.* from https://models.accordproject.org/cicero/runtime.cto

asset InfContractState extends AccordContractState {
    o Double value
}

asset InfContract extends AccordContract {}
```

state.json
```json
{
    "$class": "matt.inf.InfContractState",
    "stateId": "state1",
    "value": 0.0
}
```

contract.json
```json
{
    "$class": "matt.inf.InfContract",
    "contractId": "contract1"
}
```

request.json
```
{
    "$class": "org.accordproject.cicero.runtime.Request"
}
```

When you run ...
```bash
mattmbp:minimal matt$ ergorun execute --contract contract.json --request request.json --state state.json  model.cto infinity.ergo
09:05:31 - info:
{
  "clause": "mattXinfXInf",
  "request": {
    "$class": "org.accordproject.cicero.runtime.Request",
    "transactionId": "54280e38-bfda-43f1-98d1-9048a64d7aeb",
    "timestamp": "2019-08-16T08:05:31.493Z"
  },
  "response": {
    "$class": "org.accordproject.cicero.runtime.Response",
    "transactionId": "994be5fe-39b9-451e-9aa1-5e6a0fc90e3e",
    "timestamp": "2019-08-16T08:05:31.504Z"
  },
  "state": {
    "$class": "matt.inf.InfContractState",
    "value": null,
    "stateId": "inf0"
  },
  "emit": []
}
```
Notice the value of `value` in the state response.

[...]

Issue #157, opened 2019-12-06
Request CTO validation is not working


dselman:
**Describe the bug**
It is possible to send an invalid request (with respect to the CTO model) and have the Ergo execute.

**To Reproduce**
1. Download the attached template
2. Using the request and state within the archive
3. `cicero execute --request monthlysales-invalid.json --state state.json`
4. Note that execution succeeds, when it should fail

**Expected behavior**
The execution should fail because `sales[0].price.doubleValue` is a string.

**Version**

Cicero 0.13.5

jeromesimeon:
Concerto is generally quite lose with validation. Bug or feature?

```
bash-3.2$ cat models/model.cto 
namespace org.accordproject.promissorynote

import org.accordproject.money.MonetaryAmount from https://models.accordproject.org/money.cto

transaction Payment extends Request {
  o MonetaryAmount amountPaid
  o Boolean isValid
	o Integer someNumber
}
bash-3.2$ cat request.json 
{
  "$class": "org.accordproject.promissorynote.Payment",
  "amountPaid": {
    "doubleValue": 1.0,
    "currencyCode": "USD"
  },
  "isValid" : "something",
  "someNumber" : "something else"
}
bash-3.2$ concerto validate --ctoFiles models/*.cto --sample request.json 
8:45:45 AM - info:
{
  "$class": "org.accordproject.promissorynote.Payment",
  "amountPaid": {
    "$class": "org.accordproject.money.MonetaryAmount",
    "doubleValue": 1,
    "currencyCode": "USD"
  },
  "isValid": false,
  "someNumber": null,
  "transactionId": "7a521051-6af6-4aca-8a26-199f66b09753",
  "timestamp": "2019-12-06T13:45:45.698Z"
}
```

jeromesimeon:
A proposed resolution is to change the validation semantics to be strict. This means atomic values in the JSON will have to be proper instances of the corresponding atomic type.

## Before

### Boolean
```
true validate Boolean -> true
false validate Boolean -> false
"true" validate Boolean -> true
3.14 validate Boolean -> false
null validate Boolean -> false
undefined validate Boolean -> false
{} validate Boolean -> false
[] validate Boolean -> false
```

## After

### Boolean
```
true validate Boolean -> true
false validate Boolean -> false
"true" validate Boolean -> ERROR
3.14 validate Boolean -> ERROR
null validate Boolean -> ERROR
undefined validate Boolean -> ERROR
{} validate Boolean -> ERROR
[...]

PR #161, opened 2019-12-12
feature(core) Stricter validation for atomic values

jeromesimeon:

# Issue #157 & #158 

A stricter validation/serialization semantics for atomic types and values in CTO models.

### Changes
- #157 Atomic values in the JSON have to be strictly of the type declared in the model. E.g., `true` validates against `Boolean` but `"true"` does not. E.g., `3.14` validates against `Double` but does not validate against `Integer` or `Long`. E.g., `"1"` validates against `String` but `1` does not.
- #158 Serializing non-finite Doubles (`NaN` or `Infinity` or `-Infinity`) throws a validation error
- Updates the tests accordingly

### Flags
- This is a breaking change
- The PR is against a new `release-0.83` branch

### DIFF
--- a/packages/concerto-core/lib/serializer/jsonpopulator.js
+++ b/packages/concerto-core/lib/serializer/jsonpopulator.js
@@ -222,17 +222,41 @@ class JSONPopulator {
             }
             break;
         case 'Integer':
-        case 'Long':
-            result = this.ergo ? parseInt(json.nat) : parseInt(json);
+        case 'Long': {
+            const num = this.ergo ? json.nat : json;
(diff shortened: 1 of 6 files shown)


jeromesimeon:
> We should add removing the "ergo" stuff that snuck into Concerto v1 imo...

We should definitely discuss how we want that to work, from a code-base / feature perspective!

jeromesimeon:
> We should add removing the "ergo" stuff that snuck into Concerto v1 imo...

This is somewhat orthogonal to this specific PR and issue. I opened a new issue to track the "ergo stuff removal" here:  #162

Issue #62, opened 2019-09-16
Remove support for system models


dselman:
Support for system models has complicated the codebase and makes it hard to reuse models, as they now have an *implicit* dependency on a system model.

On balance I think we should make a breaking change and remove the support for system models, forcing people to explicitly declare dependencies on their own models using an `import` with a `from`.

jeromesimeon:
A few comments:
1. We need a definition for what the structure of a `transaction` `event` `participant` etc is. Those need to be built in in some sense
2. One of the motivations for the current structure is to be able to switch those systems model (I think two main instances are for hyperledger composer and for Cicero). There are a set of defaults for hyperledger. This could be simplified if this gets removed (not sure if it's acceptable).
3. If we revise the top of the type hierarchy, it might be nice to decide on a super type for all of classes (e.g., `concept` could be the super class for everything or it could be `object` or `class`)

jeromesimeon:
Why having a top type would be useful: https://github.com/accordproject/ergo/issues/420

dselman:
1. Agreed
2. Yes, I think this is acceptable. We can preserve the hyperledger system model as a "standard" model on models.accordproject.org so that people can import it into their models and refactor if required.
3. Agreed. `concept` seems like a natural candidate, but that may have some API implications.

jeromesimeon:
Small fact finding mission. I can't seem to find a clear dependency from Concerto to HL:
```
bash-3.2$ grep -r concerto composer
Binary file composer/packages/composer-admin/test/data/businessnetwork.zip matches
composer/packages/composer-common/test/data/zip/test-npm-archive/node_modules/animaltracking-model/.npmignore:!chaincode/src/concerto
composer/packages/composer-common/test/data/zip/test-npm-archive/node_modules/animaltracking-model/models/mozart.cto:namespace com.ibm.concerto.mozart
Binary file composer/packages/composer-common/test/data/zip/test-archive.zip matches
Binary file composer/.git/objects/pack/pack-19a8cd93ba0365a189b8f632a9e5f3440428f9ae.pack matches
Binary file composer/.git/index matches
```

With that project being deprecated, can we assume supporting the hyper ledger composer system model is not a requirement?

dselman:
> Small fact finding mission. I can't seem to find a clear dependency from Concerto to HL Composer:
> 
> ```
> bash-3.2$ grep -r concerto composer
> Binary file composer/packages/composer-admin/test/data/businessnetwork.zip matches
[...]

PR #165, opened 2020-02-03
Remove Ability to Create Custom System Models + Simplify Definitions for Core Classes

jeromesimeon:
# Issue #62 

### Changes
- Remove ability to customise system models
  - Remove notion of systems table
  - Always uses a fixed built-in system models for `Asset` `Transaction` `Event` and `Participant`
- Some changes to criteria allowing relationships (now relationship can be created to any class with an identifier)
- Numerous fixes to tests, notably since now hyper ledger CTO in tests is no longer added as a systems model.
- Changes namespace for core classes `Transaction` `Participant` `Event` `Asset` to `system`
- Core classes now have neither identifier nor any attributes

### Flags
- `Concept` is still a special thing, and not declared as the top of the type hierarchy. See https://github.com/accordproject/concerto/issues/180
- This does not address all issues and questions around: identifiers, relationships. See https://github.com/accordproject/concerto/issues/181


packages/concerto-core/lib/systemmodel.js
```diff
@@ -0,0 +1,29 @@
...
+abstract asset Asset {  }
+abstract participant Participant {  }
+abstract transaction Transaction identified by transactionId {
+  o String transactionId
+}
+abstract event Event identified by eventId {
+  o String eventId
+}`;
+
+module.exports = { systemNamespace, systemFileName, systemModel };
```
mttrbrts: Why do we need this file? I thought that this would be explicitly added by users and hosted in a models repo?

packages/concerto-core/lib/systemmodel.js
```diff
@@ -0,0 +1,29 @@
...
+
+'use strict';
+
+const systemNamespace = 'org.accordproject.base';
+const systemFileName = '@' + systemNamespace + '.cto';
+const systemModel = `namespace org.accordproject.base
+abstract asset Asset {  }
+abstract participant Participant {  }
+abstract transaction Transaction identified by transactionId {
+  o String transactionId
```
mttrbrts: Didn't we discuss removing these built-in transactionIds?

packages/concerto-core/test/composer/composermodel.js
```diff
@@ -14,8 +14,7 @@
 
 'use strict';
 
-const SYSTEM_MODELS = [
-
+const COMPOSER_MODEL =
```
mttrbrts: Can we remove this file entirely?

packages/concerto-core/test/composer/composermodelutility.js
```diff
@@ -14,22 +14,16 @@
 
 'use strict';
```
mttrbrts: Redundant too?

packages/concerto-core/test/composer/i18n.js
```diff
@@ -23,7 +23,7 @@ const Factory = require('./../../lib/factory');
 const Serializer = require('./../../lib/serializer');
 const ModelManager = require('./../../lib/modelmanager');
 const ModelUtil = require('./../../lib/modelutil');
-const Util = require('../composer/systemmodelutility');
+const Util = require('../composer/composermodelutility');
 const Moment = require('moment-mini');
 
 describe('Globalization', function() {
```
mttrbrts: Redundant?

packages/concerto-core/test/composer/composermodel.js
mttrbrts: Ah, this file is only used as a utility to make the tests work unchanged, right?

packages/concerto-core/lib/systemmodel.js
jeromesimeon: Sorry 😦 I had forgotten to push my latest changes to the branch on the main repo. It should be good now. The "system" model (which is now empty as per our discussion) is:
```
namespace system
abstract asset Asset {  }
abstract transaction Transaction { }
abstract event Event { }
abstract participant Participant {  }
```

We still need to declare the names `Asset` `Transaction` `Event` and `Participant` so users can write `concept Cat { o Participant owner }` having that file is a convenient way to do that (it would be done purely in JavaScript).

packages/concerto-core/test/composer/i18n.js
jeromesimeon: > Redundant?

Redundant with what?

jeromesimeon: To clarify: `composermodelutility` is now a user model as any other. It's been used throughout all the tests though, so kept it for testing.

mttrbrts: Ah, I wrote this comment before I realised that this was just a utility file for tests

PR #169, opened 2020-02-26
feature(core) Add convenience function to load model manager from model files objects

jeromesimeon:


jeromesimeon:
This also addresses #170

Issue #172, opened 2020-03-01
Remove Winston Logger


mttrbrts:
**Is your feature request related to a problem? Please describe.**
The concerto logger module (`require('@accordproject/concerto-core').Logger`) is used extensively by the rest of the AP stack, however, in most cases, only the console output is actually used. The logger is implemented using the Winston library which introduces an additional 130KB to the concerto bundle size; over 30% of the total bundle size.

**Describe the solution you'd like**
Remove the dependency on Winston but maintain the same interface of the Logger module to avoid having to refactor the rest of the stack. The new logger module would use the Node `console` module directly.

**Additional context**
This should be a Concerto v1.0 feature and it potentially changes the semantics of logging and would remove logging to disk.

mttrbrts:
Fixed in the `release-1.0` branch

PR #175, opened 2020-03-01
refactor(logger): Remove winston dependency …

mttrbrts:
# Issue #172 
Remove Winston dependency and rewrite the logger class, while retaining backwards compatibility with commonly used functions. 

### Changes
- Removes `winston`, `tripe-beam` and `fast-safe-stringify` as NPM dependencies
- Adds `silly`, `verbose`, `log`, `http`, `info`, `warn`, `debug` and `error` functions to the Logger class that closely follow the behaviour of the Node Console module
- Adds `add` operation to allow the use of custom transports
- Removes other Winston operations from the module's signature.

### Flags
- Although this is a private API, it has been used extensively by downstream libraries and systems. While care has been taken to ensure coverage and maintain API compatibility, this should be considered a breaking change and released as part of Concerto v1.0
- Retains the `@private` annotation in JSDocs.
- This new version deliberately flattens Error objects nested inside a log object. This is a change in behaviour. It was assumed that the previous behaviour was incorrect. See the output of the test `should call logger.info with an Error object in the message`

Comparison of previous console output (left), with proposed console output (right)

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "concerto",
-	"version": "0.82.0",
+	"version": "0.82.6",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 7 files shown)


mttrbrts:
> 1. What's the rationale for having `info` return a JSON structure. Is that going to help with downstream log processing or hinder it? What is the error message contains JSON, won't that make it harder to read?

In my screenshot the left-hand side is the current behaviour, this PR fixes that.

> 2. For inclusion in Concerto 1.0, I'm afraid this PR (and the corresponding changes) should be made against the `release-1.0` branch. That branch already contain quite a number of breaking changes, could you please reopen a PR against it?

Thanks, I'll rebase against that and try again.

jeromesimeon:
> > 1. What's the rationale for having `info` return a JSON structure. Is that going to help with downstream log processing or hinder it? What is the error message contains JSON, won't that make it harder to read?
> 
> In my screenshot the left-hand side is the current behaviour, this PR fixes that.
> 

oh! great, actually.

> > 1. For inclusion in Concerto 1.0, I'm afraid this PR (and the corresponding changes) should be made against the `release-1.0` branch. That branch already contain quite a number of breaking changes, could you please reopen a PR against it?
> 
> Thanks, I'll rebase against that and try again.

👍

Issue #171, opened 2020-02-27
Instance Generation should provide Regex compatible values for fields


mttrbrts:
**Is your feature request related to a problem? Please describe.**
When generating a JSON instance for a model with fields that have regex constraints, the generated instances are not valid instances of the model.

**Describe the solution you'd like**
We should use a library such as https://github.com/fent/randexp.js, to safely create sample values that are compatible with regex annotations on models.

This will allow upstream tools (such as concerto-ui) to generate values that are valid instances of the model

jeromesimeon:
@mttrbrts Is this issue closed by #178 ?

mttrbrts:
Yes, thank you!

PR #178, opened 2020-03-07
feat(valuegenerator): Generate valid values for Regex and Range meta properties

mttrbrts:
# Issue #171 
Adds functions that generate valid range and regex values when generating sample and empty instance JSON values.

### DIFF
--- a/packages/concerto-core/lib/serializer/valuegenerator.js
+++ b/packages/concerto-core/lib/serializer/valuegenerator.js
@@ -16,6 +16,7 @@
 
 const loremIpsum = require('lorem-ipsum');
 const Moment = require('moment-mini');
+const RandExp = require('randexp');
 
 /**
  * Empty value generator.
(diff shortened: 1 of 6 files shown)


packages/concerto-core/lib/serializer/valuegenerator.js
```diff
@@ -95,6 +96,56 @@ class EmptyValueGenerator {
...
+    }
+
+    /**
+     * Get a default range value.
+     * @param {number} lowerBound the lower bound on the range.
+     * @param {number} upperBound the upper bound on the range.
+     * @param {string} type the number type for the range
+     * @return {number} a String value.
+     */
+    getRange(lowerBound, upperBound, type) {
```
mttrbrts: Is there another sensible thing to do for the `empty` case? Default to the `lowerBound` perhaps?

packages/concerto-core/lib/serializer/valuegenerator.js
```diff
@@ -16,6 +16,54 @@
 
 const loremIpsum = require('lorem-ipsum');
 const Moment = require('moment-mini');
+const RandExp = require('randexp');
+
+const randomNumberInRangeWithPrecision = function (userMin, userMax, precision, systemMin, systemMax) {
```
jeromesimeon: Comments for what this function calculates, please.

packages/concerto-core/lib/serializer/valuegenerator.js
mttrbrts: > I actually like how you did it. Do we check that lowerBound and upperBound are in the proper integer/long range? I don't even know if we check those limits on validation.

Yes that happens in the `randomNumberInRangeWithPrecision` function

I initially tried this, but this causes `NaN` answers when performing precision calculations. For this use case of generating a sample number, I feel that we can restrict the range to lower default extremes, I chose to give Decimals the same range as Longs for convenience

PR #186, opened 2020-03-18
refactor: All typed classes use methods with `_` prefix

### LINKED ISSUE #121: Cannot have a property on a request called validate
Get a runtime failure because when we try to call the validate function on the Concerto object we are actually accessing the field.

mttrbrts:
# Issue #121
In order to fix the issue where all methods in the `Typed` class and its subclasses (such as `.validate`) are reserved model properties, this PR renames all of the relevant methods to include a `_` prefix.
Fixes #121 

### Changes
- This is a *BREAKING CHANGE* and is therefore targetted for the 1.0 release.
- Changes to the API are documented in this diff, https://github.com/accordproject/concerto/compare/release-1.0...mttrbrts:mr-fix-121?expand=1#diff-46b73b28ea461a5927c3cbfe28a21bac
- I've also changed the parser to prevent identifiers starting with `_`, this guarantees that we won't have any collisions in the future.

### Flags
- @jeromesimeon and I discussed several options, listed #121, and we concluded that this change would introduce the fewest breaking changes to API clients while keeping the natural object modification behaviour for user-defined model properties, e.g. `myConcept.name = "Foo"`
- This approach also means that we don't have to maintain an extensive list of reserved property names and the difficulty of new methods introducing a breaking change in the future
- The only reserved property is `toString` which is retained to maintain compatibility with the built-in JS function.


packages/concerto-core/api.txt
```diff
@@ -136,44 +136,44 @@ class TransactionDeclaration extends ClassDeclaration {
...
-   + Relationship fromURI(ModelManager,String,String,String) 
+   + boolean _isRelationship() 
+   + Relationship _fromURI(ModelManager,String,String,String) 
 }
 class Resource extends Identifiable {
    + String toString() 
-   + boolean isResource() 
-   + Object toJSON() 
+   + boolean _isResource() 
+   + Object _toJSON()
```
dselman: Intentional? JS calls `toJSON` automatically when you call JSON.stringify.

packages/concerto-core/lib/model/typed.js
```diff
@@ -188,7 +188,7 @@ class Typed {
      * without using the Serializer.
      * @private
      */
-    toJSON() {
+    _toJSON() {
         throw new Error('Use Serializer.toJSON to convert resource instances to JSON objects.');
```
dselman: This code will now no longer get called if you try to call JSON.stringify.

packages/concerto-core/api.txt
mttrbrts: Thanks. No this wasn't deliberate

lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.82.7",
+  "version": "1.0.0-alpha.1",
```
jeromesimeon: Please undo, we can do an alpha release using scripts and GitHub tagging.

package.json
```diff
@@ -15,7 +15,7 @@
   },
   "name": "concerto",
   "description": "You must install [Lerna](https://lernajs.io) to build this multi-package repository.",
-  "version": "0.82.7",
+  "version": "1.0.0-alpha.1",
```
jeromesimeon: Please undo, we can do an alpha release using scripts and GitHub tagging.

packages/concerto-core/test/data/parser/classdeclaration.badidentifierprefix.cto
```diff
@@ -11,16 +11,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-'use strict';
 
-/*eslint-disable no-unused-vars*/
```
jeromesimeon: That's a weird change. Git hiccup? Maybe it's just a wrongly identified file rename.

packages/concerto-core/test/data/parser/functiondeclaration.bracesmissing.js
```diff
@@ -1,26 +0,0 @@
-/*
```
jeromesimeon: Are those files dead code?

lerna.json
mttrbrts: I'm happy to back it out however, please note that:
1. This change was done to satisfy the API diff validation. If I don't incremement the version, the changelog.txt gives misleading information. i.e. breaking changes are recorded against the wrong version
2. This change was done using the scripts

packages/concerto-core/test/data/parser/classdeclaration.badidentifierprefix.cto
mttrbrts: Yes, this is weird. This wasn't a rename, I deleted `functiondeclaration.throwsexception.js` and created a new file `classdeclataion.badidentifierprefix.cto`, they just happen to have lines in common.

packages/concerto-core/test/data/parser/classdeclaration.badidentifierprefix.cto
```diff
@@ -11,16 +11,11 @@
...
- * @throws {Error}
- * @exception {Error}
- */
-function onTestTransactionLulz() {
-    console.log('onTestTransaction');
+namespace com.testing
+
+asset Asset identified by id {
+  o String id
+  o String validate // This is ok
```
mttrbrts: @jeromesimeon This file is used in unit tests to prove that previously reserved values can now be used.

packages/concerto-core/test/data/parser/functiondeclaration.bracesmissing.js
mttrbrts: Yes, these are legacy from when Composer supported "transaction processor functions". This makes me wonder whether there is other dead code anywhere, maybe even in the parser?

Issue #187, opened 2020-03-27
Functional API and Plain JS Objects


dselman:
Concerto would benefit from a non class based API - allowing Concerto objects to be plain old JS objects. You would no longer need to use the `Factory` to create them, or the `Serializer` to convert them to/from plain JS objects.

**Describe the solution you'd like**
Examine all the methods on Resource, Identifiable, Typed etc and move them into a set of functions. The functions could be static methods on a class, or just exported as a library of functions.

E.g.

```
const modelManager = new ModelManager()
...
const myObject = { ... }
Concerto.validate( myObject, modelManager );
```

**Describe alternatives you've considered**

Messing with the prototype: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object

**Additional context**

Removing the need to use the Factory and Serializer to create Concerto objects would simplify the APIs, learning curve and could lead to performance improvements.

> This new API could be developed in parallel to the existing API, allowing for a progressive migration strategy.

Note that we don't necessarily need to change the core internal APIs (things like `ModelManager`).

Here are the instance-based APIs which would need to be re-implemented as functions.

```
class Concept extends Typed {
   + boolean isConcept() 
}
class Identifiable extends Typed {
   + string getIdentifier() 
   + void setIdentifier(string) 
   + string getFullyQualifiedIdentifier() 
   + String toString() 
   + boolean isRelationship() 
   + boolean isResource() 
   + String toURI() 
}
class Relationship extends Identifiable {
   + String toString() 
   + boolean isRelationship() 
   + Relationship fromURI(ModelManager,String,String,String) 
}
class Resource extends Identifiable {
   + String toString() 
   + boolean isResource() 
   + Object toJSON() 
}
class Typed {
   + string getType() 
   + string getFullyQualifiedType() 
   + string getNamespace() 
   + void setPropertyValue(string,string) 
   + void addArrayValue(string,string) 
   + boolean instanceOf(String) 
}
class ValidatedConcept extends Concept {
   + void setPropertyValue(string,string) throws Error
   + void addArrayValue(string,string) throws Error
   + void validate() throws Error
}
class ValidatedResource extends Resource {
   + void setPropertyValue(string,string) throws Error
   + void addArrayValue(string,string) throws Error
   + void validate() throws Error
}
```

jeromesimeon:
I love this idea. Some fairly high-level questions.

[...]

PR #188, opened 2020-03-27
(feat) Functional API

dselman:

# Issue #187 

Create a functional API for concerto objects.

### Changes
- 
### Flags
- 

### Related Issues
- Issue #<NUMBER HERE>
- Pull Request #<NUMBER HERE>


dselman:
> Most immediate suggestion: please do the work on top of the `release-1.0` branch.
> 

I like the idea of releasing this first for `0.82.x` as I think it will help people migrate progressively to `1.0`.

> I think we may need the validate method to return a new JS object (we are filling in default values, turning ISO 8601 dates into moment objects, etc). -- Alternatively we could separate those things into two passes: one which changes the data representation (e.g., string to date) the other which validates and only fails/succeeds without changing anything).
> 
> I would love if those methods would be functional (i.e., validate would not change the input JSON but construct a new one).

Yes, good idea. Should not be hard to implement.

dselman:
Already merged into release-1.0 branch.

Issue #181, opened 2020-03-10
Revisions to identifiers & relationships


jeromesimeon:
(This is a place-holder issue, need more details and explanations).

Current status:
- Identifiers can be and have to be declared on transactions and events, but no other classes
- Relationships are distinguished from simple `has a` properties of a class, and have to be to an identifiable class
- The relationship can be to a resource or to an inline object

Some changes in `release-1.0`:
- Any class can have an identifier
- Relationship can be to any class which has an identifier
- Internally identifiers are both kept as a special `$identifier` property and the user-defined property declared in the `identified by` part of a class declaration

Some further possible changes:
- Distinguish between internal identifiers (necessary for relationships) and user-defined properties

jeromesimeon:
_Notes from discussion (@dselman @mttrbrts @jeromesimeon)_

## User stories

- When creating a model, one wants to distinguish "inline" data (e.g., company containing an address) from "relationship" data (e.g., purchase order referencing a company)
- When creating a model, one wants to distinguish external identifiers (e.g., SSN, DB key) from identifiers used for the sole purpose of creating a relationship
- When editing a contract, one wants to simply refer to the relationship e.g., party, but just want to edit the relationship identifier

## Current Issues

- When creating writing contract logic, one does not want to have to invent identifiers (See #...)
- Current "hack" that identifiers from the base model are hidden from logic does not apply now that systems models are gone
- Unhydrated relationships are unsound (Ergo compiler assumes the relationship is "hydrated")

## Technical proposal

1. New relationship / identifier model

   1. concepts can be identified or not
   2. a concept `C` can be used through a relationship (`--> C name`) iff it is identified
   3. a concept `C` can be used inline (`o C name`) iff it is not identified
   4. identifiers for identified concepts can be managed by the system (declared as `identified` and serialized as property `$identifier`) or by the users (declared as `identified by myId` and serialized as property `myId`)
   5. identifiers managed by the system are created through a `generate` function parameter whose signature is from object to a `String` identifier.
      The function should be complete and deterministic. A default function is provided which returns a stable hash based on the fields in the object.
      identifiers managed by the system do not appear (cannot be accessed or specified) in the contract logic.
   6. identifiers managed by the user have to be provided in the instance objects and are not automatically created.
      identifiers managed by the system always appear (can be accessed and have to be specified) in the contract logic.
   7. asset/participant/event/transaction are special concepts that are always identified

2. Resolving Relationships

[...]

PR #196, opened 2020-06-10
IDs & relationships for #181

dselman:
# Issue #181
Rework identifiers and relationships. See issue for details and design.

### Changes
- PEG grammar updated 
   - allow `identified by` for Concepts
   - add support for `identified` —  automatically create an identifying field called `$identifier`
- Remove `Concept` from the mode (`Resource` and `Concept` are now functionally identical)
- Update `Factory` to always delegate to `newResource` when `newConcept` is called, checking whether the `Concept` has an identifying field
- Allow relationships to any class that has an identifying field (reject otherwise)

### Flags
- Nothing yet to automatically generate IDs. Should that be a feature of the `Factory`?


packages/concerto-core/lib/introspect/classdeclaration.js
```diff
@@ -116,6 +120,20 @@ class ClassDeclaration extends Decorated {
...
+     * This method should only be called by system code.
+     * @private
+     */
+    addIdentifierField() {
+        const definition = {};
+        definition.id = {};
+        definition.id.name = '$identifier';
+        definition.propertyType = {};
+        definition.propertyType.name = 'String';
+        this.properties.push(new Field(this, definition));
```
jeromesimeon, line 134: I am unclear if this should be added to the properties, or remain internal.

packages/concerto-core/lib/introspect/parser.pegjs
```diff
@@ -833,7 +844,7 @@ Decorators
   = decorators:(d:Decorator __ {return d;})*
 
 AssetDeclaration
-  = decorators:Decorators __ abstract:AbstractToken? __ AssetToken __ id:Identifier __ idField:IdentifiedByField? __ classExtension: ClassExtension? __
```
jeromesimeon, line 836: I thought the design was to always have asset/participant/transaction/events identified, so maybe we disallow it in the grammar?

packages/concerto-core/lib/serializer/jsongenerator.js
```diff
@@ -86,14 +84,14 @@ class JSONGenerator {
     visitClassDeclaration(classDeclaration, parameters) {
 
         const obj = parameters.stack.pop();
-        if (!((obj instanceof Resource) || (obj instanceof Concept))) {
```
jeromesimeon, line 89: 👍

packages/concerto-core/lib/serializer/jsongenerator.js
```diff
@@ -86,14 +84,14 @@ class JSONGenerator {
...
-            throw new Error('Expected a Resource or a Concept, but found ' + obj);
+        if (!((obj instanceof Resource))) {
+            throw new Error('Expected a Resource, but found ' + obj);
         }
 
         let result = {};
         let id = null;
 
-        if (obj instanceof Identifiable && this.deduplicateResources) {
+        if (obj.isIdentifiable() && this.deduplicateResources) {
```
jeromesimeon, line 94: 👍

packages/concerto-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@accordproject/concerto-core",
-  "version": "0.82.7",
+  "version": "1.0.0",
```
jeromesimeon, line 3: This should not be done by hand (this will prevent the use of the publish scripts). Please revert.

packages/concerto-core/lib/introspect/parser.pegjs
dselman: I think what we should ensure is that in a type hierarchy the first non-abstract asset (etc) has an identifier declared. I will tighten up the programmatic validation and add tests for that.

packages/concerto-core/package.json
dselman: We will have to remember to version in the change log by hand.

dselman:
> Looks great so far. Only one requested changes (version in package.json) and a few questions/comments.
> 
> I think some more testing, etc. but a great start. Couple of open questions:
> 
> * allow asset/participant/events/transactions without identifier or not?

I think we should prevent non abstract non identified asset/participant/events/transactions. It is a breaking change...

> * prevent "inlined" form for identified classes? (i.e., force having a relationship to them?).

Here I am inclined to say that we should allow containment of identified classes. If person A creates a model using `asset` they can't know whether person B would like to use it with a containment or using a relationship. I've seen quite a few models that would like to store a collection of transactions, for example.

dselman:
Not sure why ObjectValidator coverage has been lowered by these changes. I need to look into that.

packages/concerto-core/test/introspect/identifieddeclaration.js
```diff
@@ -0,0 +1,215 @@
+/*
```
jeromesimeon, line 1: Looks great.

Issue #205, opened 2020-09-15
Investigate: switch to different DateTime library than moment (e.g., dayjs)


jeromesimeon:
moment.js is nice but big and with lots of features we do not use.

We may want to consider switching to a different library which is smaller, if we can. e.g., https://github.com/iamkun/dayjs

https://moment.github.io/luxon/

### Also

We should probably do this anyway, considering the recently posted status of development for moment.js:
https://momentjs.com/docs/#/-project-status/

DianaLease:
Slides comparing Moment vs Dayjs vs Luxon: https://docs.google.com/presentation/d/1bGSR076MzrLslwMMdGPkEg4Hyf_PZDad5kEeV5g5r-E/edit?usp=sharing

Repl comparing usage & performance: https://repl.it/@DianaLease/Comparing-Moment-Alternatives#perf-tests.js

irmerk:
Went into the call favoring Luxon by a lot, but I think the case was made for Dayjs and we should definitely switch to that. That is, given that we feel the plugins cover our use cases and vanilla JS isn't our pick.

jeromesimeon:
> Went into the call favoring Luxon by a lot, but I think the case was made for Dayjs and we should definitely switch to that. That is, given that we feel the plugins cover our use cases and vanilla JS isn't our pick.

Thanks for the analysis. I think after today's presentation I agree.

I gave a quick look at https://date-fns.org It looks nice, but is bigger than day.js -- haven't gotten a chance to look at performances and features. One possible benefit is that it relies on JavaScript Date objects so we could, I believe, completely remove dependency to date/time libraries in Concerto, and use it only for evaluation in Ergo.

DianaLease:
List of `moment` methods used in the `concerto-core` package of the release-1.0 branch:
- `Moment().utc()`
- `Moment.parseZone(...)`
- `Moment.isMoment(...)`
- `moment(..., moment.ISO_8601).isValid()`
- `new Moment()`

jeromesimeon:
> List of `moment` methods used in the `concerto-core` package of the release-1.0 branch:
> 
> * `Moment().utc()`
> * `Moment.parseZone(...)`
> * `Moment.isMoment(...)`
> * `moment(..., moment.ISO_8601).isValid()`
> * `new Moment()`

A small test to explain each of those:
```
bash-3.2$ node test.js
new Moment()                                  : 2020-11-04T15:54:40-05:00
[...]

PR #218, opened 2021-01-07
chore(datetime): migrate from moment to dayjs

DianaLease:
Signed-off-by: Diana Lease <dianarlease@gmail.com>

# Issue #205 

### Changes
-  Migrates concerto to use dayjs instead of moment due to the reasons described in #205 

### Flags
- We no longer preserve the timezone offset as we did with `Moment.parseZone()` bc dayjs does not offer equivalent functionality.
- Dayjs is less strict about what it will parse as a date. To combat this, this PR adds validation to check if the value is undefined or not a string in `serializer/objectvalidator.js`
- I'll keep this PR open to discuss on next weeks's WG call

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1715,6 +1715,38 @@
 				"@types/node": ">= 8"
 			}
 		},
+		"@sinonjs/commons": {
+			"version": "1.8.1",
+			"resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.8.1.tgz",
+			"integrity": "sha512-892K+kWUUi3cl+LlqEWIDrhvLgdL79tECi8JZUyq6IviKy/DNhuzCRlbHUjxK89f4ypPMMaFnFuR9Ie6DoIMsw==",
+			"requires": {
+				"type-detect": "4.0.8"
+			}
+		},
+		"@sinonjs/formatio": {
+			"version": "3.2.2",
+			"resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-3.2.2.tgz",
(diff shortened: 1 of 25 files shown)


mttrbrts (approved):
Good work, @DianaLease 👏 

Just one question on this statement:

> We no longer preserve the timezone offset as we did with Moment.parseZone() bc dayjs does not offer equivalent functionality.

Does this imply that all parent libraries will now become responsible for TZ / Offset normalization before using Concerto? If not, the Dayjs behaviour could cause errors where the offset is silently dropped (which actually changes the timestamp)?

DianaLease:
> Does this imply that all parent libraries will now become responsible for TZ / Offset normalization before using Concerto? If not, the Dayjs behaviour could cause errors where the offset is silently dropped (which actually changes the timestamp)?

The offset will be used when it gets converted into utc, so the _actual_ time shouldn't change. However, once it's in utc, we won't have a way to know which timezone it was originally in when passed to dayjs.  (cc @jeromesimeon)

jeromesimeon (approved):
Looks great! Unless anybody has further objections, I would vote for merging this PR and make this switch to `day.js` part of the upcoming Concerto 1.0 release.

jeromesimeon:
> > Does this imply that all parent libraries will now become responsible for TZ / Offset normalization before using Concerto? If not, the Dayjs behaviour could cause errors where the offset is silently dropped (which actually changes the timestamp)?
> 
> The offset will be used when it gets converted into utc, so the _actual_ time shouldn't change. However, once it's in utc, we won't have a way to know which timezone it was originally in when passed to dayjs. (cc @jeromesimeon)



> > Does this imply that all parent libraries will now become responsible for TZ / Offset normalization before using Concerto? If not, the Dayjs behaviour could cause errors where the offset is silently dropped (which actually changes the timestamp)?
> 
> The offset will be used when it gets converted into utc, so the _actual_ time shouldn't change. However, once it's in utc, we won't have a way to know which timezone it was originally in when passed to dayjs. (cc @jeromesimeon)

Couldn't have said it better.

PR #219, opened 2021-01-19
feature(build) Switches from TravisCI to GitHub actions

jeromesimeon:

- Switches from TravisCI to GitHub Actions
- Includes:
   - Unit tests (includes linter, license check and actual tests)
   - Micro-publish to `npm` on every commit to the repo, with a timestamp
   - Publish to `npm` on release
   - Code coverage reporting through coverall
   - Tests on both Ubuntu and Windows

### Perceived Benefits

- Tests seem to run _much_ faster than Travis, but in some cases we do see timeouts on concerto model http requests
- A lot of scripts used to determine what kind of git event triggered the build and when to publish are now gone. This looks to me like a big improvement and should facilitate maintenance.
- Windows build seems a lot more reliable than in TravisCI and adding different versions of nodes or other OS (e.g., MacOS?)  should now be much easier.
- One less third party service to keep track of. See also issue https://github.com/accordproject/technical-steering-committee/issues/4

### Flags

- Very first repo with this approach -- it certainly would benefit from a thorough review before we go ahead and update all our other repos
- The following specific issue in coveralls actions (used to publish code coverage results) made me pause. Is this a concern? https://github.com/coverallsapp/github-action/issues/68
- Change in process: the publish workflow will open a PR with a change to the package.json version rather than push that change as a commit directly to master (I personally prefer, but would love other's feedback on this)
- We are losing caching -- I did not attempt to port that part of the TravisCI workflow (but it's quite fast, so maybe we don't need it)


.github/workflows/build.yml
```diff
@@ -0,0 +1,94 @@
...
+      job-status: ${{ job.status }}
+
+    steps:
+      - name: git checkout
+        uses: actions/checkout@v2
+
+      - name: lerna install
+        run: |
+          npm install lerna@^3.22.0
+          ./node_modules/.bin/lerna bootstrap 2>&1
```
irmerk, line 81: Can we not use the result of `L28-L31`?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+          branch: clausehq-publish-${{ github.event.release.tag_name }}
+          delete-branch: true
+          title: 'chore(actions): publish ${{ github.event.release.tag_name }} to npm'
+          body: |
+            # NPM Publish
+
+            Publish build to NPM
+          assignees: ${{ github.actor }}
+          reviewers: ${{ github.actor }}
+          draft: false
```
irmerk, line 52: Is this what you mentioned here?

> Change in process: the publish workflow will open a PR with a change to the package.json version rather than push that change as a commit directly to master (I personally prefer, but would love other's feedback on this)

Seems like an extra step that can be a useful sanity check before finalizing.

irmerk (commented):
A couple thoughts on specific code. Will follow up with comments on your PR description.

irmerk:
> **[BENEFIT]** Tests seem to run much faster than Travis, but in some cases we do see timeouts on concerto model http requests

What does this mean, that we would need to manually re-run the workflow to resolve these timeouts?

irmerk:
> **[FLAG]** The following specific issue in coveralls actions (used to publish code coverage results) made me pause. Is this a concern? coverallsapp/github-action#68

Yeah this seems like a concern to me.

irmerk:
> **[FLAG]** We are losing caching -- I did not attempt to port that part of the TravisCI workflow (but it's quite fast, so maybe we don't need it)

Along with the benefits you mentioned: 

- _One less third party service to keep track of_
- _Tests seem to run much faster than Travis_

I find this tradeoff fine, especially when this isn't a service that has a high frequency of development.

.github/workflows/build.yml
jeromesimeon: apparently not. I think every job is its own space. Good question though and yes I tried. Maybe it can be cached, but for now that's the solution I found.

jeromesimeon: I can do a bit more investigation on that before we roll out.

.github/workflows/publish.yml
jeromesimeon: 👍 Yes it is. and agreed.

jeromesimeon:
> > **[BENEFIT]** Tests seem to run much faster than Travis, but in some cases we do see timeouts on concerto model http requests
> 
> What does this mean, that we would need to manually re-run the workflow to resolve these timeouts?

Correct. Same as in Travis (it happened in Travis too to be fair -- sometimes calls to `https://models.accordproject.org` do not resolve). We can change the timeouts, but I thought I would still flag it.

Issue #228, opened 2021-02-03
Root types have disappeared


jeromesimeon:
**Describe the bug**

Types `Participant`, `Event`, `Transaction` have disappeared in the `1.0.0` release.

**To Reproduce**

CTO file:
```
namespace org.test

transaction T {
}

participant P {
}

event E {
}

concept Foo {
  o Transaction t
  o Participant p
  o Event e
}
```

JSON:
```
{
  "$class": "org.test.Foo",
  "t": { "$class": "org.test.T" },
  "p": { "$class": "org.test.P" },
  "e": { "$class": "org.test.E" }
}
```

Should validate, but loading the model throws an error:
```
zsh-5.8$ ./index.js validate --ctoFiles test.cto --sample test.json
3:37:50 PM - ERROR: Undeclared type Transaction in property org.test.Foo.t File 'test.cto': 
```

dselman:
The linked PR adds the abstract types to the concerto namespace. You will have to import the types/ns to reference them in a CTO file.

dselman:
Fixed.

jeromesimeon:
I'm unclear that this is properly fixed. A few questions/issues.

1. it's confusing that `Concept` is automatically imported but not the other root types (that's also an unnecessary breaking change imo)
2. it's unclear that the root types for `Transaction` etc are properly placed in the type hierarchy. (See example below)

### Example

**CTO file*
```
namespace org.test

// Should not be needed
import concerto.Transaction
import concerto.Participant
import concerto.Event

concept C {
}

transaction T {
}

participant P {
}

event E {
}

concept Foo {
	o Concept c
	o Transaction t // Validation error when containing a class T
	o Participant p
	o Event e
}
```

**data**

```
{
  "$class": "org.test.Foo",
  "c": { "$class": "org.test.C" },
  "t": { "$class": "org.test.T" },
  "p": { "$class": "org.test.P" },
  "e": { "$class": "org.test.E" }
}```

**Error**

```
zsh-5.8$ ./index.js validate --ctoFiles test.cto --sample test.json 
2:38:20 PM - ERROR: Instance org.test.C#null has property t with type org.test.T that is not derived from concerto.Transaction
```

dselman:
I can pick this one up.

jeromesimeon:
> I can pick this one up.

[...]

PR #230, opened 2021-02-09
feat (root types) : add abstract root types in concerto ns

dselman:

# Issue #228

Adds abstract types to the system `concerto` namespace:

```
namespace concerto

abstract concept Concept {}
abstract asset Asset {}
abstract transaction Transaction {}
abstract participant Participant {}
abstract event Event {}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1660,6 +1660,38 @@
 				"@types/node": ">= 8"
 			}
 		},
+		"@sinonjs/commons": {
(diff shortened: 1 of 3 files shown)


jeromesimeon (changes requested):
Excellent! Could we add something based on the example in the original issue as a test?

jeromesimeon (approved):
👍

PR #236, opened 2021-03-17
chore(CI) Align github workflows with cicero

jeromesimeon:

### Why

We need to do a second round of design on CI migration from Travis to GitHub Actions because:
1. we have to variants now with slightly different takes in https://github.com/accordproject/concerto/tree/master/.github/workflows and https://github.com/accordproject/cicero/tree/master/.github/workflows
2. We need to decide which node version we support
3. Better to maintain one well understood version to make our lives easier
4. Travis is being _very_ capricious and the matter feels more urgent than it was a year ago

### Changes

- Consolidate GitHub action workflows for build/testing/publication to npm

### Flags

- Still draft, one key question is when is `lerna` installed and how, but want this PR ready for discussion on Tech WG call

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,7 +76,7 @@ jobs:
       - notify
 
     name: Publish to npm
-    if: ${{ success() && github.event_name == 'push'}}
+    if: ${{ success() && github.event_name == 'push' && github.repository_owner == 'accordproject' }}
     runs-on: ubuntu-latest
 
     outputs:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: build
+name: Build
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


irmerk:
Is this off of https://github.com/accordproject/technical-steering-committee/issues/4?

jeromesimeon:
> Is this off of [accordproject/technical-steering-committee#4](https://github.com/accordproject/technical-steering-committee/issues/4)?

Indeed it is

jeromesimeon:
@mttrbrts Update on this one:

1. Fixed a few things in the script, they work for building, testing and timestamp deployment on Windows/MacOS/Unbuntu
2. Caching seems to work (new feature)
3. Added logic to disable npm publication in forks (so people forking the repo should be safe now)
4. Previous features still working include code coverage
5. Cleaner than previous script by using `npx lerna`
6. Still need to re-test publication on release (I will need to merge this PR to test that)

There are other improvements we could consider, but to me this feels like addressing our main requirements and a good initial "blueprint" we can use in other projects.

Once merged / tested I'll update markdown transform and cicero with the same scripts, but you're very welcome to do further improvements and we still need to switch the remaining projects.

mttrbrts (commented):
Thanks @jeromesimeon . Adding a retrospective approval

PR #237, opened 2021-03-19
fix(core) Root hierarchy should have proper inheritance

jeromesimeon:

### Changes

- squeeze core classes `Event` `Transaction` `Asset` `Participant` back in the type hierarchy
- fixes tests
- add missing `timestamp` to `Event` and `Transaction`
- strangely sometimes we had identifiers generated even when not declared, maybe this was still what we wanted but I was recalling the opposite
- fixes an unrelated bug in `isExplicitlyIdentified` API call sometimes returning undefined

### Flags

- This needs quite a bit more testing especially of the base types and new type hierarchy
- I'm quite confused about `timestamp` why are `event` an `transaction` have a timestamp but the other classes do not? what's the criteria?
- Somewhat mixed confidence that this is really what we intended to do (but I believe it's at least closer to that goal)

### DIFF
--- a/packages/concerto-core/lib/modelmanager.js
+++ b/packages/concerto-core/lib/modelmanager.js
@@ -66,10 +66,10 @@ class ModelManager {
     addRootModel() {
         this.addModelFile( `namespace concerto
         abstract concept Concept {}
-        abstract asset Asset {}
-        abstract transaction Transaction {}
-        abstract participant Participant {}
-        abstract event Event {}`, 'concerto.cto');
+        abstract concept Asset {}
+        abstract concept Transaction { o DateTime timestamp }
+        abstract concept Participant {}
+        abstract concept Event { o DateTime timestamp }`, 'concerto.cto');
     }
 
     /**
--- a/packages/concerto-core/test/introspect/identifieddeclaration.js
+++ b/packages/concerto-core/test/introspect/identifieddeclaration.js
@@ -52,7 +52,7 @@ asset Order {
             mm.addModelFile( `
 namespace test
 
-asset Order {
+asset Order identified {
     o Double price
 }
             `, 'test.cto');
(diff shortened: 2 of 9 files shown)


jeromesimeon:
@mttrbrts : this is a good test of the new build script, with successful linting and unit testing but regression on coverage.

packages/concerto-core/lib/modelmanager.js
```diff
@@ -66,10 +66,10 @@ class ModelManager {
     addRootModel() {
         this.addModelFile( `namespace concerto
         abstract concept Concept {}
-        abstract asset Asset {}
-        abstract transaction Transaction {}
-        abstract participant Participant {}
-        abstract event Event {}`, 'concerto.cto');
+        abstract concept Asset {}
```
dselman, line 69: I don't think we want to convert all these to Concepts - they should retain their asset etc stereotypes.

packages/concerto-core/test/introspect/identifieddeclaration.js
```diff
@@ -52,7 +52,7 @@ asset Order {
             mm.addModelFile( `
 namespace test
 
-asset Order {
+asset Order identified {
```
dselman, line 55: To be discussed I think. This now allows assets to not be identifiable - which seems odd.

dselman (commented):
Nice. Probably best that we have a quick chat, just to ensure we are aligned.

packages/concerto-core/lib/modelmanager.js
jeromesimeon: The are concepts: the new type hierarchy has Assets, Participant, Transaction and Event extensions of Concept. This could be written in two ways:
1. as above
2. as `abstract transaction Transaction extends Concept {}`

I don't think this makes any difference (recall no instance can be created for those since they are abstract) but with the definition nb 2. I was having some weird errors, possibly due to the fact that we are using `transaction` before it's actually defined.

packages/concerto-core/test/introspect/identifieddeclaration.js
jeromesimeon: I just never know which of those are meant to be `identified` as a built-in thing or as a user decision. Is it just assets? I'm a little unclear there is a benefit those distinctions, but it may be because I forgot earlier discussions.

I'll play a bit by adding `identified` to assets and report.

PR #240, opened 2021-03-19
fix(core) Root hierarchy should have proper inheritance (v2)

jeromesimeon:
### Changes

A variant of #237 where assets are always identified. I'm quite confused by the number of test fixes, wondering why those weren't failures before.


jeromesimeon:
Reason for test failure is this "bug" https://github.com/accordproject/concerto/issues/241

PR #265, opened 2021-04-16
chore(deps): Upgrade dependencies. Node 12+ only

mttrbrts:
This PR upgrades the majority of the node dependencies for each of the packages to the latest available version.

This is a *BREAKING CHANGE*. 

## Flags
- This change means that Concerto is now only supported on Node 12 and above (Node 10 goes EOL at the end of the month, https://nodejs.org/en/about/releases/).
- This addresses a couple of security issues in dependencies (that also get propagated throughout the AP stack).
- There are a few noteable exceptions to the upgrade:
  - `uuid`, the next major version of `uuid` switches to ES Modules which we are not able to mock in our tests
  - `acorn`, the plugin framework has been completely rewritten in the latest releases, and I don't understand the context enough to upgrade it right now.


jeromesimeon (approved):
nice!

PR #267, opened 2021-04-18
feature(ergo) Add missing bits of ergo runtime validation

jeromesimeon:

### Changes

This is one of three PRs refactoring the following:
- moving run-time validation for the ES6 compilation target from `ergo-compiler` to `ergo-runtime` (PR: https://github.com/accordproject/ergo/pull/818)
- moving run-time validation for the ES6 compilation target from `ergo-runtime` to `concerto` (this PR)
- adjust Cicero to changes in the logic manager and model manager APIs (PR: https://github.com/accordproject/cicero/pull/644)

This has the following benefits

- at long last, we now having run time validation consistent with the Q*cert code generation for ES6 all in one place (rather than in 2 places) which should make it easier to test systematically
- this is the proper architecture to start adding other backends
- we can finally disentangle the ergo engine from the compiler, with almost no dependency between both except for the model manager


mttrbrts (changes requested):
LGTM, although can you address the drop in coverage, please?

PR #287, opened 2021-06-29
Documentation & Types: Updates to JSDoc and TypeScript interface for Concerto 1.0

jeromesimeon:
### Changes

- Align typescript interface with Concerto 1.0 changes
- Some fixes / alignment to JSDoc for consistency
- Remove some unused parameters left from when we used to have systems models
- Fixes isIdentifiable call so it properly returns a boolean
- Update `index.js` so it aligns with the typescript definition to simplify future maintance


jeromesimeon:
CC: @sstone1

Issue #293, opened 2021-07-03
Update API


mttrbrts:
# Feature Request 🛍️
Today, creating new types programmatically requires string manipulation. There is an internal AST for a Concerto definition, but it is not documented nor designed for manipulation.

## Use Case

With an update API, a user could add a new declaration to an existing model, or even build models completely from scratch (e.g. from a graphical tool)

## Possible Solution

As well as defining a public API for model manipulation, we should define a JSON DOM (similar to Cicero Mark DOM) of a model definition to allow efficient serialisation and deserialization.

Possible API signatures
```
modelFile.addConcept(name, properties, [superType]);
modelFile.updateConcept(name, [properties], [superType]);
modelFile.deleteConcept(name);
classDeclaration.addProperty(name, type, optional, validators);
classDeclaration.updateProperty(name, type, optional, validators);
classDeclaration.deleteProperty(name);
```

## Context

Changing a model definition dynamically could lead to breaking changes in client code. We could chose to reject, warn, or allow breaking changes.

dselman:
@jeromesimeon @mttrbrts now that we have moved Concerto to the published metamodel I think we can close this.

PR #297, opened 2021-07-09
feat(codegen): Add JSON Schema to CTO utility

### LINKED ISSUE #294: Model generation from JSONSchema definition
# Feature Request 🛍️

Generate Concerto models from JSONSchema definitions.

## Use Case

This would allow round-tripping between Concerto and (a suitable sub-set of) JSONSchema which would be useful for contexts where client applications standardise on this. For example in microservices applications using OpenAPI definitions.

## Context

JSONSchema is a document validation tool rather than a type system. This makes it difficult to perform operations such as code generation from models. Furthermore, JSONSchema does not support type extension, at least not in the usual OO sense.

However, JSONSchema is pervasive and well adopted across platforms.

Bi-directional support for JSONSchema in Concerto would broaden it's reach to more systems, while retaining the benefits of a strongly types schema language.

## Detailed Description

1. Implement a JSONSchema-to-Concerto visitor in `concerto-tools` to complement the existing Concerto-to-JSONSchema code generation that we already have.
2. Identify a suitable sub-set of JSONSchema that can be mapped to Concerto

mttrbrts:
# Closes #294
This PR adds support for Concerto model generation from a small subset of JSON Schema.

### Changes
- Adds a new CodeGen module in the style of the JSON document inference utility.
- Includes support for top-level schema definitions
- Includes support for primitive types, `date-time` formats and objects with `properties`.

### Flags

- This feature is implemented using string manipulation to create the model. It would be greatly improved through the use of an #293. This would also allow us to better handle sub-schemas.
- This feature should be considered experimental, as such I do not propose to document this publicly or add it to the Concerto CLI yet. 
- Requires further unit test for exception paths to satisfy code-coverage requirements

### Related Issues
- Issue #294 

### Author Checklist


dselman (approved):
Very nice. We should be able to infer string regex and range constraints as well?

Issue #301, opened 2021-07-27
jsome > yargs > yargs-parser vulnerability reported by npm audit


sstone1:
Installing `@accordproject/concerto-core@1.0.4` causes npm audit to report vulnerabilities.

Two of them are caused by the dependency on the `jsome` library:

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ yargs-parser                                                 │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=13.1.2 <14.0.0 || >=15.0.1 <16.0.0 || >=18.1.2             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @ds/flows-cli                                                │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @ds/flows-cli > @ds/flows-engine >                           │
│               │ @accordproject/concerto-core > jsome > yargs > yargs-parser  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1500                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ yargs-parser                                                 │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=13.1.2 <14.0.0 || >=15.0.1 <16.0.0 || >=18.1.2             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @ds/flows-plugin-basic                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @ds/flows-plugin-basic > @ds/flows-engine >                  │
│               │ @accordproject/concerto-core > jsome > yargs > yargs-parser  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1500                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

The `jsome` library appears to be abandonware, with no update since 2018:
- https://www.npmjs.com/package/jsome
- https://github.com/Javascipt/Jsome

It seems unlikely that they will publish a new version with the dependency changes in.

[...]

PR #306, opened 2021-08-06
chore(build) Upgrade dependencies, fixes some vulnerabilities, switch to json-colorize

jeromesimeon:

### Changes for #301 

- Switch from `jsome` to `json-colorize` in logger
- Upgrade and cleanup other dependencies

### Notes

- `json-colorize` seems more used, I liked the idea of no-dependencies but the dependencies are still small here
- local `npm audit` still shows some dependencies to old `minimatch` which seems to come from `Istanbul-combine` another project that seems unmaintained. I have not attempted to fix that yet
- it's possible that this still addresses #301 since Istanbul is only a dev dependency?

### Example

Output seems fine:

Screen Shot 2021-08-06 at 4 33 53 PM


jeromesimeon:
A quick follow up: indeed we get a clean `npm audit` if we remove `Istanbul-merge` and `Istanbul-combine` from the dependencies, but this is used in some of the current GitHub workflows so we need to figure that part out.

PR #312, opened 2021-08-12
chore(deps): move to uuid@8

sstone1:
Currently installing the `concerto-core` module results in the following warning:

```
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

This is because it depends on `uuid@3.4.0`. This change moves to `uuid@8.3.2`.

Unfortunately the way the `uuid` module is written means that it is no longer possible to stub. To get the affected tests working again, I have added a new `Factory.newId` function that can be stubbed instead.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2802,42 +2802,6 @@
 				"@octokit/openapi-types": "^9.4.0"
 			}
 		},
-		"@sinonjs/commons": {
-			"version": "1.8.3",
-			"resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.8.3.tgz",
-			"integrity": "sha512-xkNcLAn/wZaX14RPlwizcKicDk9G3F8m2nU3L7Ukm5zBgTwiT0wsoFAHx9Jq56fJA1z/7uKGtCRu16sOUCLIHQ==",
-			"requires": {
-				"type-detect": "4.0.8"
-			}
-		},
-		"@sinonjs/fake-timers": {
-			"version": "6.0.1",
-			"resolved": "https://registry.npmjs.org/@sinonjs/fake-timers/-/fake-timers-6.0.1.tgz",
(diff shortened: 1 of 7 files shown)


jeromesimeon (commented):
Thanks for the PR! Are you aware of any change in behavior between those two versions of `uuid` that we should be aware of or concerned about? [Since we mock the calls in test, not sure what we actually test there]

jeromesimeon (changes requested):
Looks great. Could you please add the corresponding method in the typescript interface as well? https://github.com/accordproject/concerto/blob/cd3de6b83cbf7708b2c56eca375eaa54d3056b50/packages/concerto-core/types/index.d.ts#L218

sstone1:
> Looks great. Could you please add the corresponding method in the typescript interface as well?
> 
> https://github.com/accordproject/concerto/blob/cd3de6b83cbf7708b2c56eca375eaa54d3056b50/packages/concerto-core/types/index.d.ts#L218

@jeromesimeon I did! https://github.com/accordproject/concerto/pull/312/files#diff-e8048dc5a91e53453da6b9fc03e13c5a41d61435bc2e24fb3d05691999118b6cR219

I'm not aware of any differences in behaviour between the two versions, other than "more randomness".

jeromesimeon:
> > Looks great. Could you please add the corresponding method in the typescript interface as well?
> > https://github.com/accordproject/concerto/blob/cd3de6b83cbf7708b2c56eca375eaa54d3056b50/packages/concerto-core/types/index.d.ts#L218
> 
> @jeromesimeon I did! https://github.com/accordproject/concerto/pull/312/files#diff-e8048dc5a91e53453da6b9fc03e13c5a41d61435bc2e24fb3d05691999118b6cR219
> 
> I'm not aware of any differences in behaviour between the two versions, other than "more randomness".

duh 🤦 😀 

this is going in!

jeromesimeon (approved):
💯

PR #327, opened 2021-09-01
feat(models): allow prerelease concerto versions to match (resolves #326)

### LINKED ISSUE #326: Unstable builds difficult to use due to Concerto versions in model files
When you use an unstable build to test a change that has been merged but not released, the validation for the Concerto version specified in the model file is slightly too strict - in my opinion at least!

For example, with a model file that contains the following:

```
concerto version ">= 1.0.0"
```

You get the following error:

```
> concerto compile --model models/model.cto --target TypeScript --output types

10:34:19 - ERROR: ModelFile expects Concerto version >= 1.0.0 but this is 1.1.3-20210826140311
```

Which means that you have to go and edit your model file, which seems like unnecessary pain.

By changing the options to the `semver` library, we can allow pre-release versions to match.

sstone1:
Update model file parsing to enable the specified Concerto version range (for example, `concerto version "^1.0.0"`) to match a pre-release version of Concerto (for example, `1.1.3-20210826140311`). This fixes the pain point in issue #326 by making it possible to install `@accordproject/concerto-core@unstable` and use it without changing your model files to match.

### DIFF
--- a/packages/concerto-core/lib/introspect/modelfile.js
+++ b/packages/concerto-core/lib/introspect/modelfile.js
@@ -564,7 +564,7 @@ class ModelFile {
      */
     isCompatibleVersion() {
         if (this.ast.version) {
-            if (semver.satisfies(packageJson.version, this.ast.version.value)) {
+            if (semver.satisfies(packageJson.version, this.ast.version.value, { includePrerelease: true })) {
                 this.concertoVersion = this.ast.version.value;
             } else {
                 throw new Error(`ModelFile expects Concerto version ${this.ast.version.value} but this is ${packageJson.version}`);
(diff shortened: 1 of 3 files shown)


mttrbrts (approved):
This looks reasonable to me.

To be explicit for future reviewers, the risk is that a prerelease version might not satisfy the version semantics of the normal (non-prerelease version), which could lead to unexpected behaviour. I believe that this is acceptable, because users of unstable concerto releases should be doing so explicitly, and therefore are aware of the risks.

jeromesimeon (approved):
Yes I think I like it. I wasn't expecting users to really be so close to the development process that they'd want to match against pre-release Concerto, but I don't see a problem if they do (at their own risks as @mttrbrts justifiably points out).

PR #338, opened 2021-09-27
fix(metamodel) Switches to Concerto metamodel 0.3

jeromesimeon:

### Description

Switches metamodel support to Concerto Metamodel `0.3` which includes terminology and some structural changes (no more abstract `ClassDeclaration`.

### Changes

- `field` renamed to `property`
- no longer use `class` using `concept` instead
- toplevel declarations are now `enums` and `concepts`

### Flags

- This is fundamentally a breaking change for anyone using the metamodel API. Releasing a version `2.0` seems somewhat overkill though. Keeping a backward compatible version would be quite a lot of work. Anyone has suggestions?
- The new model reference definition is in https://github.com/accordproject/models/pull/146


jeromesimeon:
@stefanblaginov Feedback welcome. Those are changes to the metamodel API for Concerto based on https://github.com/accordproject/models/pull/146

dselman (approved):
LGTM. We will still have the round-trip ordering issue with enums / concepts to/from CTO - but perhaps we live with that for now?

jeromesimeon:
> LGTM. We will still have the round-trip ordering issue with enums / concepts to/from CTO - but perhaps we live with that for now?

Yeah, there are a few things that we could consider depending on our release timing:
- revert the enums / concepts declaration separation (it does round trip but loses the order atm). it's been done this way mostly for UI because in the UI you it is more readable to separate enums from concepts.
- should fully qualified names just have a namespace instead? (a user would have to split the FQN to get the namespace atm)
- version? I think this one is likely out of reach

jeromesimeon:
@mttrbrts @dselman Made the changes we discussed today. A follow up review would be appreciated. New metamodel is available here: https://github.com/accordproject/models/pull/147

mttrbrts (approved):

PR #350, opened 2021-11-17
feature(parser) Switch model creation from old ast to metamodel

jeromesimeon:

### Changes

- Parser generates metamodel instance rather than custom AST (written by @dselman )
- Model creation from metamodel rather than custom AST (written by @jeromesimeon )
- Old AST is gone
- Code that used to transform old AST to meta model is gone

### Flags

- Further post-PR cleanup may be useful
- This PR requires changes to the latest `0.3` Concerto metamodel. Changes are in a separate PR: https://github.com/accordproject/models/pull/149


dselman (approved):
Looks awesome!

PR #352, opened 2021-11-22
Remove instanceof symbol

### LINKED ISSUE #47: Avoid `instanceof` which leads to runtime errors with deep npm dependencies

dselman:
# Closes #47
- Removes custom `hasInstance` code, which is incompatible with Babel
- Refactor code and tests to use new methods to determine the runtime types
- Removes support for Node 12, adds support for Node 16

### Flags
- API breaking change

### Author Checklist


packages/concerto-core/lib/introspect/classdeclaration.js
```diff
@@ -65,6 +65,7 @@ class ClassDeclaration extends Decorated {
         this.idField = null;
         this.timestamped = false;
         this.abstract = false;
+        this.type = this.ast.type;
```
jeromesimeon, line 68: oh no! that one's going to be  😭

packages/concerto-core/lib/introspect/metamodel.js
```diff
@@ -19,1069 +19,1069 @@ const ModelManager = require('../modelmanager');
 const Factory = require('../factory');
 const Serializer = require('../serializer');
 
+const metaModelCto = `/*
```
jeromesimeon, line 22: Why changes to this file? That seems unrelated to `instanceof` ? I would remove those changes to avoid conflicts with the other PRs on the new parser.

jeromesimeon (changes requested):
Looks great! Just one change request on the `metamodel.js` file.

packages/concerto-core/lib/introspect/metamodel.js
dselman: Ok, I will remove

jeromesimeon (approved):
LGTM!

PR #356, opened 2021-11-23
refactor(imports) clean up import handling from parser

jeromesimeon:

### Changes

This PR cleans up import handling which can be made simpler with the new parser using the Concerto Metamodel. Notably:

- Builds simpler import tables for wildcard and non-wildcard
- Avoid using fully qualified names which have to be parsed again when possible
- Also cleans up what looks like obsolete code in `ModelUtil`


dselman (approved):
Lovely!

PR #390, opened 2022-03-04
refactor(errors): improve error messages

### LINKED ISSUE #389: Error messages need consistency improvements
# Bug Report 🐛

The current error messages originating from `concerto-core` have a few inconsistencies that need to improved. Often it's unclear if a part of the error message description refers to a natural language word or a term or value used by Concerto.

## Possible Solution

The following principles should be followed when in comes to error messages:
- Terms should be surrounded by double quotation marks, common to US English. In this context, terms are all words that represent code, parts or snippets of Concerto and differ from the complementing explanatory text in the message.
- Full stops at the end of error message (detail) sentences should be added in order to provide a consistent error format. The reasoning behind this is that error messages (details) are proper sentences, rather than just titles. Some discussions on the topic:
https://ux.stackexchange.com/questions/18671/periods-at-the-end-of-a-sentence-in-alert-message
https://stackoverflow.com/questions/1136829/do-you-end-your-exception-messages-with-a-period
- Occurrences of "super type" should be replaced with "supertype", which seems to be widely used.
- Consistency should be improved when it comes to terms surrounded by parentheses.
- The "telegraph-ese" style of messages should be foregone and grammatical articles should be used.
- Sentence ambiguity should be removed.

stefanblaginov:
# Closes #389

### Changes
- Terms are surrounded by double quotation marks, common to US English. Terms are all words that represent code, parts or snippets of Concerto and differ from the complementing explanatory text in the error message.
- Full stops at the end of error message (detail) sentences have been added in order to provide a consistent error format. The reasoning behind this is that error messages are proper sentences, rather than just titles. Some discussion surrounding this:
https://ux.stackexchange.com/questions/18671/periods-at-the-end-of-a-sentence-in-alert-message
https://stackoverflow.com/questions/1136829/do-you-end-your-exception-messages-with-a-period
- Replace occurrences of "super type" with "supertype", which seems to be widely used.
- Improved consistency when it comes to terms surrounded by parentheses.
- Forego the "telegraph-ese" style of messages and use grammatical articles.
- Remove sentence ambiguity. The last thing that developers would like to do while debugging is do meaning guesswork.
- Some tests were improved to include more concrete error text checks (rather than partial regex matching) and to have a more useful failed test output.

### Flags
Code dependant on parsing `concerto-core` error messages is likely to break from this change.

### Author Checklist


packages/concerto-core/test/models/test.js
```diff
@@ -190,35 +190,35 @@ describe('Test Model', function(){
...
 
             // set model to a double
             cObject.model = 42.05;
-            ( function() {serializer.toJSON(cObject);}).should.throw(/.+expected type String/);
+            ( function() {serializer.toJSON(cObject);}).should.throw('Model violation in the "org.acme.Vehicle#CAR_123" instance. The field "model" has a value of "42.05" (type of value: "number"). Expected type of value: "String".');
 
             // set model to a Boolean
             cObject.model = true;
-            ( function() {serializer.toJSON(cObject);}).should.throw(/.+expected type String/);
+            ( function() {serializer.toJSON(cObject);}).should.throw('Model violation in the "org.acme.Vehicle#CAR_123" instance. The field "model" has a value of "true" (type of value: "boolean"). Expected type of value: "String".');
```
jeromesimeon, line 201: Am I the only one finding those changes confusing, surely `"1"` is a String!

packages/concerto-core/test/serializer/objectvalidator.js
```diff
@@ -233,7 +233,7 @@ describe('ObjectValidator', function () {
 
             (function () {
                 objectValidator.visit(concerto.getTypeDeclaration(data), parameters);
-            }).should.throw(/Instance undefined has property previousOwners with type undefined that is not derived from test.Person\[\]/);
+            }).should.throw('Instance "undefined" has a property "previousOwners" with type "undefined" that is not derived from "test.Person[]".');
```
jeromesimeon, line 236: I'm wondering if the `undefined` here is a bug? the sentence doesn't read quite right.  Could we double check?

packages/concerto-core/test/serializer/objectvalidator.js
```diff
@@ -264,7 +264,7 @@ describe('ObjectValidator', function () {
             parameters.stack = new TypedStack(data);
             (function () {
                 objectValidator.visit(concerto.getTypeDeclaration(data), parameters);
-            }).should.throw(/expected a Relationship/);
+            }).should.throw('Model violation in the "undefined" instance. Class "test.Person" has a value of "[object Object]". Expected a "Relationship".');
```
jeromesimeon, line 267: I'm wondering if the `undefined` here is a bug? the sentence doesn't read quite right. Could we double check?

jeromesimeon (commented):
Looks good @stefanblaginov !

I've placed a couple of comments for things that should be doublechecked.

Also, I feel there is more work we should be doing on error messages:
1. Check that all those errors are actually used, some of them may be obsolete and could be removed
2. Check for use of other errors that aren't listed in the `message/en.json` and use a different  approach e.g., `throw new Error("This is some error which is thrown by hand")`

So I wouldn't mind leaving your issue open and listing those are additional work that would be useful to do.

mttrbrts (approved):

PR #403, opened 2022-03-22
fix(types): update index.js for better type generation (resolves #402)

### LINKED ISSUE #402: TypeScript types are difficult to use
The TypeScript types are generated as follows:

```typescript
export var ModelManager: typeof import("./lib/modelmanager");
```

This means that you end up having to write verbose code to use the types:

```typescript
import {
  ModelManager
} from '@accordproject/concerto-core';

// This is OK.
const modelManager = new ModelManager();

someFunc(
  // This is way too verbose!
  modelManager: InstanceType<typeof ModelManager>
): void {
  // TODO: implementation
}
```

The problem stems from the way they are exported from `index.js`:

```javascript
module.exports.ModelManager = require('./lib/modelmanager');
```

Switching them to this pattern:

```javascript
const ModelManager = require('./lib/modelmanager');

module.exports = {
    ModelManager
};
```

Results in more common TypeScript types:

```typescript
import ModelManager = require("./lib/modelmanager");
export { ModelManager };
```

And easier usage of those types:

```typescript
import {
  ModelManager
} from '@accordproject/concerto-core';

// This is OK.
const modelManager = new ModelManager();

someFunc(
  // This is better!
  modelManager: ModelManager
): void {
  // TODO: implementation
}
```

sstone1:
https://github.com/accordproject/concerto/issues/402

Change `index.js` so that instead of requiring and exporting in one shot, like this:

```javascript
module.exports.SecurityException = require('./lib/securityexception');
```

It is separated, like this:

```javascript
const SecurityException = require('./lib/securityexception');

module.exports = {
    SecurityException
}
```

This results in better, more usable type definitions.


mttrbrts (approved):
I believe that this is only a breaking change for TypeScript clients?

sstone1:
> I believe that this is only a breaking change for TypeScript clients?

Correct!

PR #405, opened 2022-03-25
fix(types): typescript type generation improvements (contributes to #402)

sstone1:
Lots of changes to improve TypeScript generation across all modules:

- Install TypeScript as a development dependency into all projects
- Remove `npx -p typescript tsc` from `build:types` target and just use `tsc` instead
- Force `build:types` to be run during CI by adding it as the `postdoc` target (it validates all types are correct)
- Remove JSON requires that are exported and replace with `JSON.parse(fs.readFileSync(...))` equivalent
   These don't work well with TypeScript generation; the compiler omits them altogether 🤦 
- Remove `const Thing = require('module').Thing` and use `const { Thing } = require('module')`
   This style of imports doesn't work well with TypeScript generation, especially when multiple imports of the same module occur - it's also redundant to type `Thing` twice
- Import types that are used in JSDoc comments but not elsewhere in the code to give TypeScript everything it needs
   These imports are wrapped in `eslint-disable` and `eslint-enable` blocks because they are unused in code.
   These imports are wrapped in a `if (global === undefined) {` (never true) block because otherwise we get circular requires and these cause the code to fall over in a heap
   These imports are wrapped in a `istanbul ignore next` block because the `if` condition will never be true
- Fixed a bunch of `@private` vs `@protected` errors which the TypeScript compiler caught
- Enable TypeScript generation for `concerto-tools`


mttrbrts (approved):
The following pattern is incredibly ugly, but I don't see another way around it beyond migrating the whole project to TypeScript.

```
// Types needed for TypeScript generation.
/* eslint-disable no-unused-vars */
/* istanbul ignore next */
if (global === undefined) {
  ...
}
/* eslint-enable no-unused-vars */
```

I'd be happy to accept it as a short-term measure to unblock release of v2. Can you open an issue for porting to TypeScript as a long-term solution for v3+?

Issue #392, opened 2022-03-07
Provide TypeScript type definitions for built-in models


sstone1:
# Feature Request 🛍️
Please provide TypeScript type definitions for built-in models, for example for the metamodel and the new decorator sets being introduced under https://github.com/accordproject/models/pull/154

These TypeScript type definitions could be exported from one of the Concerto modules, or a new Concerto module.

## Use Case
Make it easier to work with the built-in models from a TypeScript application, without having to introduce code generation on the "client side".

dselman:
Should models.accordproject.org publish Typescript models as part of its release workflow?

jeromesimeon:
> Should models.accordproject.org publish Typescript models as part of its release workflow?

I think it already does?

https://models.accordproject.org/address.html
<img width="1374" alt="Screen Shot 2022-03-07 at 12 32 50 PM" src="https://user-images.githubusercontent.com/670099/157087026-b4ebc1de-65b8-4f7f-b180-bec1ff24bb23.png">

jeromesimeon:
A further note: all the models in `models.accordproject.org` are not _built in_. They are user code.

PR #437, opened 2022-06-01
feat(*): produce a new package with metamodel types

sstone1:
https://github.com/accordproject/concerto/issues/392

Introduce a new package, `@accordproject/concerto-types`, that provides TypeScript type definitions that are automatically generated from the metamodel. These TypeScript types can then be used by TypeScript applications that need to consume the metamodel.

This change introduces a little bit of groundwork for moving to TypeScript - there is a new top level `build` target.


mttrbrts (approved):
😎

PR #444, opened 2022-06-17
(feat) support namespaces with semver

### LINKED ISSUE #426: Model versioning
## Introduction

This is a feature proposal for introducing explicit model versioning into the Concerto modelling language. Today, it is only possible to implement model versioning by convention, by inserting a version number into the namespace of a model. This makes it hard to understand what model versions exist in a set of models and how those model versions depend on each other.

We believe in the use of semantic versioning for models, which would mean that any instance created with a 1.x.y version (for example, 1.0.0) model should be compatible and validate against the latest 1.x.x version (for example, 1.2.3) model. 

```
namespace org.example@1.0.0

concept Person {
  o String name
}
```

A given model manager would be able to load multiple versions of a model, so it could also load a new major version of that model with an incompatible change:

```
namespace org.example@2.0.0

concept Person {
  o String name
  o Integer age 
}
```

## Namespaces
Namespaces are versioned, which implicitly versions all the declarations within the namespace.

Version numbers are composed of:
1. Major number
2. Minor number
3. Patch number
4. Pre-release identifier

Version numbers are encoded within namespaces using the following scheme:

```
namespace org.acme@1.2.3-alpha.0
```

## Imports
A model may reference an explicit version of a namespace - and import types from that namespace. A model must only reference a single version of a namespace.

An import that does not specify a version identifier uses the latest “stable” version of a namespace.

```
import org.acme.Customer@1.2.3-alpha.0 // pre-release
import org.acme.Product // uses "stable"
import org.acme.Product@2.3.4 from https://acme.org/models/product.cto // stable, loaded from a URL
import org.acme.Customer@unstable // use latest prerelease
```

If a version number is specified if must be explicit - i.e. version ranges are not supplied (we do not have a “lock file” concept).

## Instances
JSON serialisation encodes the version of a namespace used for a type. Validation checks that the instance conforms to the version of the namespace. An instance can be verified against the latest (compatible) version of a namespace. On save an instance may optionally be updated to reference the latest compatible version.

An instance would be able to refer to a model version using the model version number that the application used to create that instance:

```
[
  {
    "$class": "org.example.Person@1.0.0",
    "name": "Simon"
  },
  {
    "$class": "org.example.Person@2.0.0",
    "name": "Simon",
    "age": 99
  }
]
```

## Model Manager
Model Manager manages a set of (versioned) namespaces. A model manager may contain version N as well as version N+1 of a namespace, as well as pre-release versions of the namespace.

## Migration
Migration functions may be registered to migrate instances from version N to version N+1 of a namespace.

dselman:

# Closes #426 

### Changes
- Modify the CTO grammar to support an optional semantic version number after the name of a namespace
- Updates parser generation to use Peggy.js v2.0.1
- Adds ability to import types from versioned namespaces `import test.Foo@1.0.0`
- Adds a new syntax for importing multiple types from a versioned namespace: `import {Foo,Bar} from test@1.2.3`
- Add a `versionedNamespacesStrict` option for the `ModelManager` to enforce that namespaces and imports are versioned
- Added tests
- Added a `concerto` namespace as well as a `concerto@1.0.0` namespace. When `versionedNamespacesStrict` is true only `concerto@1.0.0` is added to the model manager. When `versionedNamespacesStrict` is false both `concerto` and `concerto@1.0.0` are added
- Metamodel AST is now versioned

### Flags

- All existing models stored as JSON are invalid with respect to the versioned metamodel
- All existing decorator sets are invalid with respect to the versioned metamodel

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist


packages/concerto-core/lib/basemodelmanager.js
```diff
@@ -93,13 +95,33 @@ class BaseModelManager {
...
+        return this.versionedNamespacesStrict;
+    }
+
     /**
      * Adds root types
      * @private
      */
     addRootModel() {
+        // iff we allow unversioned namespaces we should *also*
+        // allow people to import from concerto@1.0.0 namespace
```
mttrbrts, line 112: I think that this comment is incorrect? 
```
        // iff we allow unversioned namespaces we should *also*
        // allow people to import from concerto namespace
  ```

File: packages/concerto-core/lib/rootmodel.json [hunk 2]
```diff
@@ -1 +1,118 @@
...
+    "$class": "concerto.metamodel@1.0.0.Model",
+    "namespace": "concerto@1.0.0",
+    "imports": [],
+    "declarations": [
+        {
+            "$class": "concerto.metamodel@1.0.0.ConceptDeclaration",
+            "name": "Concept",
+            "isAbstract": true,
+            "properties": [],
+            "location": {
```
mttrbrts, line 11: Can we drop the locations?

packages/concerto-core/api.txt
```diff
@@ -237,13 +240,14 @@ class ValidatedResource extends Resource {
    + void validate() throws Error
 }
 class ModelLoader {
-   + ModelManager[] loadModelManager(string[],object,boolean?,number?) 
-   + ModelManager[] loadModelManagerFromModelFiles(object[],string[],object,boolean?,number?) 
+   + ModelManager[] loadModelManager(string[],object,boolean?,boolean?,number?) 
+   + ModelManager[] loadModelManagerFromModelFiles(object[],string[],object,boolean?,boolean?,number?)
```
mttrbrts, line 244: These signatures don't match the JS API. Perhaps it's a bug in the JS/jsdoc parser?

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -726,6 +725,66 @@ relative_part
...
   = scheme ":" hier_part ("?" query)?
 
+/**
+ * SemVer.org v2
+ * https://semver.org/spec/v2.0.0.html
+ */
+semver
+  = versionCore:versionCore
+    pre:('-' @preRelease)?
+    build:('+' @build)?
```
mttrbrts, line 735: Can you add a test for a version string that uses preRelease and build components, please?

packages/concerto-cto/lib/printer.js
```diff
@@ -180,27 +180,27 @@ function declFromMetaModel(mm) {
         result += 'abstract ';
     }
     switch (mm.$class) {
-    case 'concerto.metamodel.AssetDeclaration':
+    case 'concerto.metamodel@1.0.0.AssetDeclaration':
```
mttrbrts, line 183: It would be convenient to have `concerto.metamodel@1.0.0` refactored somewhere to avoid touching all of these files when we need to bump the version in future. Is there any benefit to making it dynamic?

packages/concerto-metamodel/lib/metamodel.json
```diff
@@ -1 +1,2340 @@
...
+            "$class": "concerto.metamodel@1.0.0.ConceptDeclaration",
+            "name": "Position",
+            "isAbstract": false,
+            "properties": [
+                {
+                    "$class": "concerto.metamodel@1.0.0.IntegerProperty",
+                    "name": "line",
+                    "isArray": false,
+                    "isOptional": false,
+                    "location": {
```
mttrbrts, line 16: Can we dump all of the location information here too, please?

packages/concerto-tools/lib/codegen/fromcto/golang/golangvisitor.js
```diff
@@ -97,8 +97,8 @@ class GoLangVisitor {
     visitModelFile(modelFile, parameters) {
         // we put all the code into the main package, but we
         // seperate out into multiple files using the namespaces
-        let packageName = this.toGoPackageName(modelFile.getNamespace());
-        parameters.fileWriter.openFile(packageName + '.go');
+        const { mangledNamespace } = ModelUtil.parseNamespace(modelFile.getNamespace());
```
mttrbrts, line 100: I find `mangled` an odd name. What's the purpose of this? To escape `@` for targets that don't allow it?

mttrbrts: `escapedNamespace`?

mttrbrts (changes requested):

packages/concerto-core/lib/basemodelmanager.js
dselman: I've cleaned up the comment.

packages/concerto-core/api.txt
dselman: Yes - it doesn't handle the nested options syntax correctly. I don't think it matters too much - it is only used to detect breaking API changes.

mttrbrts (approved):

sstone1 (changes requested):
The inlined CTO here https://github.com/accordproject/concerto/blob/0bc1ad09dd085486b446920b5d6d6ff050a3236d/packages/concerto-metamodel/lib/metamodelutil.js#L181 needs updating.

PR #448, opened 2022-06-28
feat(*): add custom resource factories

sstone1:
Introduce a new facility that allows a user to extend Concerto's built in resource type (as a "custom" resource), and have instances of that type created by Concerto's built in factory methods.

For example, given a model:

```
namespace com.example

asset MyAsset {
  o String value
}
```

... a user can define a TypeScript class, with their own property definitions and custom functions:

```
import { CustomResource } from '@accordproject/concerto-core';

class MyAsset extends CustomResource {
  value: string;

  getValue() {
    return this.value;
  }

  setValue(value: string) {
    this.value = value;
  }
}
```

... a user can register that TypeScript class with a factory:

```
import { Factory } from '@accordproject/concerto-core';

const factory = new Factory(/* ... */);
factory.addResourceFactory('com.example', 'MyAsset', opaque => new MyAsset(opaque));
```

... and then when a user uses that factory, they get instances of that TypeScript class instead of the built-in Resource class:

```
const resource = factory.newResource('com.example', 'MyAsset', '1234');
if (resource instanceof MyAsset) {
  console.log('woohoo!', resource.getValue());
}
```

### DIFF
--- a/packages/concerto-core/lib/model/customresource.js
+++ b/packages/concerto-core/lib/model/customresource.js
@@ -1,0 +1,131 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const { TypedStack } = require('@accordproject/concerto-util');
+const Resource = require('./resource');
+
+/**
+ * CustomResource is a Resource that is designed to be extended by
+ * generated client classes.
+ * @abstract
+ */
+class CustomResource extends Resource {
+    /**
+     * Constructor.
+     * @param {unknown} opaque Opaque data
+     */
+    constructor(opaque) {
+        // These properties are internals and should not be exposed to the caller.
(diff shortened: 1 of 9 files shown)


packages/concerto-core/lib/model/customresource.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Sets a property. If validation is enabled, then the property value is first validated
+     * to ensure that it does not violate the model.
+     * @param {string} propName - the name of the field
+     * @param {string} value - the value of the property
+     * @throws {Error} if the value is not compatible with the model definition for the field
+     */
+    setPropertyValue(propName, value) {
```
mttrbrts, line 44: This feels a little like we're doing optional inheritance between `Resource` and `ValidatedResource`. Is this the best pattern, here?

mttrbrts (approved):

packages/concerto-core/lib/model/customresource.js
sstone1: @mttrbrts I think if I had the opportunity to make breaking changes I would have merged `Resource` and `ValidatedResource` (so that `Resource` simply has optional validation), changed the constructor signature for `Resource` (so that it accepts an "opaque" object) and just used `Resource` instead. 

Maybe a v3 item to revisit? Any other suggestions?

dselman: I'm torn on this one... I think regretting those runtime Resource domain objects, and forcing people to use Factory and Serialiser to manipulate them. I think it is steepens the learning curve.

I think the [Concerto](https://github.com/accordproject/concerto/blob/master/packages/concerto-core/lib/concerto.js) functional class shows that we can accomplish almost everything with a functional API that sits on-top of JSON data, without paying the cost of JSON serialisation to/from domain objects.

Irrespective of this PR, what do you think would be the preferred way forward for Concerto v2?

sstone1:
Parking this one in favour of #449

PR #456, opened 2022-07-19
feat(tools): improved C# code generation

mttrbrts:
Extends the C# code generation to:
1. optionally support NewtonsoftJson as a serialization / deserialization library
2. add `useNewtonsoftJson` and `useSystemTextJson` as flags to the `concerto compile` CLI.
3. prefix C# namespaces with `Concerto.Models` to avoid clashes with other packages.
4. make `enum` declarations public
5. switch from `new` modifier to use `virtual`/`override` modifiers for classes. See https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/knowing-when-to-use-override-and-new-keywords
6. make Enum fields nullable to prevent C# from making an assumption about a default value
7. escape field names that are reserved keywords in C#

### Related Issues
- This change is used by the new `concerto-dotnet` library, https://github.com/accordproject/concerto-dotnet

### Author Checklist

### DIFF
--- a/packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
+++ b/packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
@@ -78,15 +78,31 @@ class CSharpVisitor {
      * @private
      */
     visitModelFile(modelFile, parameters) {
+        // If no serlialization library is specified we default to the .NET one.
+        // However, we also allow both options to be specified
+        if (!parameters.useSystemTextJson && !parameters.useNewtonsoftJson){
+            parameters.useSystemTextJson = true;
+        }
+
         parameters.fileWriter.openFile(modelFile.getNamespace() + '.cs');
         parameters.fileWriter.writeLine(0, 'using System;');
-        parameters.fileWriter.writeLine(0, 'using System.Text.Json.Serialization;');
-        parameters.fileWriter.writeLine(0, `namespace ${modelFile.getNamespace()} {`);
+
+        if (parameters.useSystemTextJson){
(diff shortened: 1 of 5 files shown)


mttrbrts:
FYI @rkotangoor @bergeron

packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
```diff
@@ -78,15 +78,31 @@ class CSharpVisitor {
...
+            parameters.fileWriter.writeLine(0, 'using System.Text.Json.Serialization;');
+            parameters.fileWriter.writeLine(0, 'using Concerto.Serialization;');
+        }
+
+        if (parameters.useNewtonsoftJson){
+            parameters.fileWriter.writeLine(0, 'using NewtonsoftJson = Newtonsoft.Json;');
+            parameters.fileWriter.writeLine(0, 'using NewtonsoftConcerto = Concerto.Serialization.Newtonsoft;');
+        }
+
+        parameters.fileWriter.writeLine(0, `namespace Concerto.Models.${modelFile.getNamespace()} {`);
```
dselman, line 100: Can we please make the namespace prefix optional and passed in as an option? This will allow CLI users to decide if they would like to use a prefix and what it should be.

dselman (changes requested):

packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
mttrbrts: Done

dselman (approved):

PR #466, opened 2022-07-26
(fix) CSharp code gen with versioned namespaces

dselman:

Fixes the escaping of versioned namespaces.

### Changes
- Improve namespace escaping
- Manual test to ensure that the output from CSharp code generation compiles with .NET compiler

### Flags

- <ONE>
- <TWO>

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist

### DIFF
--- a/packages/concerto-core/api.txt
+++ b/packages/concerto-core/api.txt
@@ -79,6 +79,7 @@ class ClassDeclaration extends Decorated {
    + string getName() 
    + string getNamespace() 
    + string getFullyQualifiedName() 
+   + string getEscapedFullyQualifiedName() 
    + Boolean isIdentified() 
    + Boolean isSystemIdentified() 
    + Boolean isExplicitlyIdentified() 
--- a/packages/concerto-core/changelog.txt
+++ b/packages/concerto-core/changelog.txt
@@ -23,7 +23,7 @@
 #
 # Note that the latest public API is documented using JSDocs and is available in api.txt.
 #
-Version 2.3.0 {774980a857090905fe276b6e94f1dbb1} 2022-07-21
+Version 2.3.0 {2fcea15dff01ec08b79e6b71115e84e9} 2022-07-26
 - Versioned namespaces
 
 Version 2.3.0 {28dd70fd922b3b8b491216661796b95b} 2022-06-21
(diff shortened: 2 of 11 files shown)


mttrbrts (approved):

sstone1 (commented):
I have some questions on the developer workflow here.

I would expect that a developer would want C# classes generated for the major version(s) of the model that they want to use - e.g. if their application needs to work with v1 and v2 of `org.example`, then they would want two sets of C# classes to accomplish that. 

I wouldn't think that the developer would care about the minor/patch version(s) of the model, as these should be compatible changes and they shouldn't need to change their application to work with later minor or patch versions - unless they want to access new fields.

It looks like with these changes I would end up with `Org.Example_1_2_3` and `Org.Example_2.3.4`. Should we just generate `Org.Example_1` and `Org.Example_2`? Are there more appropriate mechanisms for .NET that we should be using - should these be v1 and v2 of the Org.Example assembly (using the `Org.Example` namespace)?

mttrbrts:
> It looks like with these changes I would end up with `Org.Example_1_2_3` and `Org.Example_2.3.4`. Should we just generate `Org.Example_1` and `Org.Example_2`? Are there more appropriate mechanisms for .NET that we should be using - should these be v1 and v2 of the Org.Example assembly (using the `Org.Example` namespace)?

@bergeron @rkotangoor any thoughts?

dselman:
> I have some questions on the developer workflow here.
> 
> I would expect that a developer would want C# classes generated for the major version(s) of the model that they want to use - e.g. if their application needs to work with v1 and v2 of `org.example`, then they would want two sets of C# classes to accomplish that.
> 
> I wouldn't think that the developer would care about the minor/patch version(s) of the model, as these should be compatible changes and they shouldn't need to change their application to work with later minor or patch versions - unless they want to access new fields.
> 
> It looks like with these changes I would end up with `Org.Example_1_2_3` and `Org.Example_2.3.4`. Should we just generate `Org.Example_1` and `Org.Example_2`? Are there more appropriate mechanisms for .NET that we should be using - should these be v1 and v2 of the Org.Example assembly (using the `Org.Example` namespace)?

Yes, good questions. I think we'd expect versioning to be handled at the .NET package level — but we still need to be able to generate code with multiple versions of the same namespace in the model manager. I can imagine a couple of options:
1. We drop the version numbers from the namespace, only generating code for the latest version of a namespace
2. We drop *some* of the version number (minor/patch) and generate code for the latest version

PR #469, opened 2022-07-27
fix(*): expose model util methods, remove redundant type lookup

sstone1:
### Changes

- Expose certain `ModelUtil` methods and correct the TypeScript types for `parseNamespace`
- Remove a seemingly redundant codepath from `ModelFile.getFullyQualifiedTypeName`

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/modelfile.js
+++ b/packages/concerto-core/lib/introspect/modelfile.js
@@ -393,13 +393,7 @@ class ModelFile {
                 }
             }
             else {
-                // check whether type is defined in another file
-                const fqn = this.resolveImport(type);
-                const modelFile = this.getModelManager().getModelFile(ModelUtil.getNamespace(fqn));
-                if (!modelFile) {
-                    return null;
-                }
-                return modelFile.getLocalType(fqn).getFullyQualifiedName();
+                return this.resolveImport(type);
             }
         }
         else {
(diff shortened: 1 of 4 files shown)


packages/concerto-core/lib/introspect/modelfile.js
```diff
@@ -393,13 +393,7 @@ class ModelFile {
                 }
             }
             else {
-                // check whether type is defined in another file
-                const fqn = this.resolveImport(type);
```
sstone1, line 397: Here, we get the FQN of the type. For `import { Thing } from org.example` and `import org.example.Thing` this is simple, as all the information to identify that the FQN of `Thing` is `org.example.Thing`.

For wildcard imports `import org.example.*` (which we want to drop), we have to visit the model manager and look through all wildcard imports for the namespace that contains the type we want.

Either way, we end up with the FQN of the type or we throw an error as the type cannot be resolved.

packages/concerto-core/lib/introspect/modelfile.js
```diff
@@ -393,13 +393,7 @@ class ModelFile {
...
-                const modelFile = this.getModelManager().getModelFile(ModelUtil.getNamespace(fqn));
-                if (!modelFile) {
-                    return null;
-                }
-                return modelFile.getLocalType(fqn).getFullyQualifiedName();
```
sstone1, line 402: All of this code is seemingly redundant - we use the FQN that we already have to locate the model file of the type, then we get the type, and then we get the FQN of the type (which we already know!) and return that.

sstone1: Maybe worth adding the motivation for this actually - yes, it will have a small performance benefit, but it actually enables you to get fully qualified names out of a `ModelFile` instance without having to have the imported models present in your `ModelManager` now that we have versioned imports and are looking at dropping wildcard imports.

mttrbrts (approved):

PR #478, opened 2022-08-08
fix(*): use webpack version for browsers

sstone1:
The Concerto modules cannot currently be used in a Node.js application:

```
{"$class":"concerto.metamodel@1.0.0.TypeIdentifier","name":"Model"},"isArray":true,"isOptional":false}]}]}')}},t={},function r(n){var o=t[n];if(void 0!==o)return o.exports;var u=t[n]={exports:{}};return e[n](u,u.exports,r),u.exports}(10);var e,t}));
                                                                                                                                                                                                                   ^

ReferenceError: self is not defined
    at Object.<anonymous> (/home/jenkins/workspace/models-registry_PR-111/node_modules/@accordproject/concerto-cto/dist/concerto-cto.js:2:212)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/home/jenkins/workspace/models-registry_PR-111/fixtures.js:1:20)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
```

This change updates our `package.json` files to use the Webpack'ed versions of the code for browsers only, and restores the original value of the `main` property for Node.js and other runtimes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
 				"@babel/preset-env": "7.16.11",
 				"@supercharge/promise-pool": "1.7.0",
 				"@types/jest": "28.1.1",
+				"@types/semver": "7.3.10",
 				"@typescript-eslint/eslint-plugin": "5.27.1",
 				"@typescript-eslint/parser": "5.27.1",
 				"acorn": "8.5.0",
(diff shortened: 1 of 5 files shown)


mttrbrts (approved):

dselman (commented):
Thank you!

PR #481, opened 2022-08-09
feat(*): extend concerto compare to include field validators

sstone1:
Extend `concerto compare` so that it can compare number and string validators attached to a field.

This will catch:
- Number validator added to a field (incompatible change)
- Lower bound of a number validator increased (incompatible change)
- Lower bound of a number validator decreased (compatible change)
- Upper bound of a number validator decreased (incompatible change)
- Upper bound of a number validator increased (compatible change)
- String validator added to a field (incompatible change)
- String validator regex pattern changed (incompatible change)
- String validator regex flags changed (incompatible change)

Note that for regexes, it may be possible to make a regex more lenient than it was before, but there is no way for us to programmatically assert that one regex is more lenient than another regex.

### DIFF
--- a/packages/concerto-analysis/src/compare-utils.ts
+++ b/packages/concerto-analysis/src/compare-utils.ts
@@ -12,7 +12,7 @@
  * limitations under the License.
  */
 
-import { ClassDeclaration, EnumValueDeclaration, Field, Property, RelationshipDeclaration } from '@accordproject/concerto-core';
+import { ClassDeclaration, EnumValueDeclaration, Field, NumberValidator, Property, RelationshipDeclaration, StringValidator, Validator } from '@accordproject/concerto-core';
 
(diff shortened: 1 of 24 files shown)


packages/concerto-analysis/src/compare-utils.ts
```diff
@@ -43,3 +43,13 @@ export function getPropertyType(property: Property) {
         throw new Error(`unknown property type "${property}"`);
     }
 }
+
+export function getValidatorType(validator: Validator) {
+    if (validator instanceof NumberValidator) {
```
dselman, line 48: Probably not applicable here - but we've been bitten by using instanceof for exported types, as instanceof also checks the module for the type.

dselman (approved):
A thing of beauty. 👍

PR #483, opened 2022-08-10
chore(jsonpopulator): remove data values from validation error messages

mttrbrts:
This change removes the data values from error messages when model validation fails.  Instead, error messages now report the path in the object being validated where the error occurs. A client application could choose to parse the JSONPath expression to find the offending value.

### Before
```
ERROR: Expected value "2022-08-04T10:07asda" to be of type DateTime
```

### After
```
ERROR: Expected value at path `$.identityVerification.date` to be of type `DateTime`
```

I also add a test that demonstrates the more advanced error output, previously we had full coverage, but the tests only demonstrated base cases. Surprisingly few tests failed due to this change!

### DIFF
--- a/packages/concerto-core/lib/serializer/jsonpopulator.js
+++ b/packages/concerto-core/lib/serializer/jsonpopulator.js
@@ -14,6 +14,7 @@
 
 'use strict';
 
+const { TypedStack } = require('@accordproject/concerto-util');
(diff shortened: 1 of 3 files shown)


sstone1 (approved):
LGTM, shame that `??=` isn't available in Node.js 14!

PR #484, opened 2022-08-12
BREAKING CHANGE(core): limit regex to redos safe patterns

mttrbrts:
This PR changes the Regular Expression engine for String Validators to use [re2](https://www.npmjs.com/package/re2). re2 is a drop-in replacement that prohibits back-tracking. This protects us from Regular Expression Denial of Service (ReDos) attacks.

### Changes
- Changes the implementation of the RegExp engine in `StringValidator` in `concerto-core`
- Fixes a vulnerable regex in the metamodel! 😱
- Refactors use of backtracking in the property name regex in the metamodel to the core runtime instead.
- Improves the error message for Validator failures.
- Fixes a bug with incorrect pattern code generation for JSON Schema targets.

### Flags
- This is a breaking change because existing models with regex string validators that use backreferences or lookahead assertions are no longer supported. 
- `re2` also uses Unicode mode by default.  

### Author Checklist


packages/concerto-tools/test/codegen/fromcto/jsonschema/jsonschemavisitor.js
```diff
@@ -215,7 +215,7 @@ describe('JSONSchema (samples)', function () {
             modelManager.addCTOModel( MODEL_BOUNDS );
             const visitor = new JSONSchemaVisitor();
             const schema = modelManager.accept(visitor, { rootType: 'test.Test'});
-            expect(schema.properties.myString.pattern).equal('^abc.*$');
+            expect(schema.properties.myString.pattern).equal('abc.*');
```
dselman, line 218: This got looser? Is that what we want?

dselman (approved):
One question about the JSON Schema regex - but otherwise looks good.

packages/concerto-tools/test/codegen/fromcto/jsonschema/jsonschemavisitor.js
mttrbrts: Yes, I believe so. Previously we would always add these start/end tokens in the codegen, that modifies the declared regex from the model.
https://github.com/accordproject/concerto/blob/28304562ca32b15a39349f29c3c35e1ed66182b5/packages/concerto-tools/lib/codegen/fromcto/jsonschema/jsonschemavisitor.js#L299

PR #487, opened 2022-08-16
feat(stringvalidator): client provided regex

mttrbrts:
This PR reverts the changes introduced in #484 to adopt the RE2 engine. The engine has some constraints (no browser support, memory limits etc.), so I'd like to simplify Concerto, but allow consumers to supply their own engine where they can make own judgements on the tradeoffs.

### Changes

- Revert back to the default JavaScript `RegExp` engine by default for all platforms.
- Add a new options parameter to the `ModelManager` constructor to allow a client to provide their own RegExp engine.

```
const XRegExp = require('xregexp')
XRegExp.install('astral');
const modelManager = new ModelManager({ regExp: XRegExp });
```

### Flags

- This change makes Concerto vulnerable to ReDoS attacks if client applications do not introduce further checks. This strategy is consistent with the approach that other validation libraries take (see [AJV](https://ajv.js.org/security.html#redos-attack)).

### Author Checklist


dselman (approved):

sstone1 (commented):
LGTM

PR #492, opened 2022-08-18
feat(*): add decorators to model files (resolves #491)

### LINKED ISSUE #491: Top level decorators for model files or namespaces
# Feature Request 🛍️

I would like to have some way of decorating a model file or a namespace with additional metadata for that model file/namespace.

## Use Case

I would like to have a model file/namespace be able to declare metadata that configures how the C# code generator generates C# code. For example, I would like to have a `@DotNetNamespace('Org.Acme')` decorator that tells the C# code generator to put the generated code in the `Org.Acme` namespace. This is similar to how Protocol Buffers handles these things: https://github.com/hyperledger/fabric-protos/blob/main/msp/identities.proto#L7

## Possible Solution

```
@DotNetNamespace('Org.Acme')
namespace org.acme

concept Foo {
  ..
}
```

## Context

See above use case.

sstone1:
Allow decorators to be attached to the namespace element in a model file:

```
@DotNetIgnore
@DotNetNamespace('Org.Acme')
namespace org.acme@1.2.3

concept Foo {
  ...
}
```


packages/concerto-core/lib/introspect/decorated.js
```diff
@@ -38,13 +38,14 @@ class Decorated {
      * Create a Decorated from an Abstract Syntax Tree. The AST is the
      * result of parsing.
      *
-     * @param {ModelFile} modelFile - the model file
+     * @param {ModelFile|undefined} modelFile - the model file
```
dselman, line 41: Do we really want `|undefined` here?

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1435,9 +1435,10 @@ Version
      }
 
 Model
-  = version:Version? ns:Namespace imports:Imports? body:SourceElements? {
+  = version:Version? decorators:Decorators? ns:Namespace imports:Imports? body:SourceElements? {
       const result = {
         $class: "concerto.metamodel@1.0.0.Model",
+        decorators: optionalList(decorators),
```
dselman, line 1441: I *think* that if you omit this when the list is empty then you won't have to update all the test JSON files... We also need to update the metamodel itself in models.accordproject.org repo (bump the version?) to add this as an optional attribute. I think the metamodel is also duplicated in @accordproject/concerto-metamodel repo?

dselman (changes requested):

packages/concerto-cto/lib/parser.pegjs
sstone1: Given I've already updated all the test JSON files, I'll remember that for next time! I've updated the @accordproject/concerto-metamodel package (which is in this repo), but I will also update the models.accordproject.org repo in a separate PR.

packages/concerto-core/lib/introspect/decorated.js
sstone1: I was hoping that `ModelFile` could be a `Decorated` as per other decorated objects, however in the constructor for `ModelFile` you cannot pass `this` into the `super(modelFile, ast)`. 

The original approach was if you do not pass a `ModelFile` into the constructor (passing `undefined`), you can pass it later via `Decorated.setModelFile`, and if you've done neither of those things `Decorated.process` will throw an error.

I think this is actually cleaner if `Decorated.getModelFile` is abstract and must be implemented by the subclass, where `ModelFile.getModelFile` can return `this`. See new commit!

PR #539, opened 2022-10-21
feat(proto3): implement a Concerto to Proto3 converter

### LINKED ISSUE #418: Code Gen support for Protobuf
# Feature Request 🛍️

Add code to convert a Concerto model to a [Protobuf](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec) file.

## Use Case
Would like to use Protobuf to represent Concerto models.

## Possible Solution
Extend the code generation package and CLI to support Protobuf.

stefanblaginov:
# Closes #418

### Changes
- Added a way to convert Concerto *.cto files to Proto3 *.proto ones.

### Author Checklist

# Concerto to Proto3 mapping proposal

# Mappings

## Primitive types

| Concerto | Proto3 |
| --- | --- |
| String | string |
| Double | double |
| Integer | sint64 |
| Long | sint64 |
| DateTime | google.protobuf.Timestamp |
| Boolean | bool |

Concerto:

```
concept AllTypes {
  o String foo
  o Double bar
  o Integer baz
  o Long qux
  o DateTime quux
  o Boolean corge
}
```

```protobuf
message AllTypes {
  string foo = 1;
  double bar = 2;
  sint64 baz = 3;
  sint64 qux = 4;
  google.protobuf.Timestamp quux = 5;
  bool corge = 6;
}
```

## Meta properties

`optional` fields remain `optional` in a Proto3 message.

Arrays are marked as `repeated` in a Proto3 message and are implicitly `optional` as well.

Concerto:

```
concept Person {
  o String firstName
  o String middleName optional
  o String lastName
  o String[] phoneNumbers
}
```

Proto3:

```protobuf
message Person {
  string firstName = 1;
  optional string middleName = 2;
  string lastName = 3;
  repeated string phoneNumbers = 4;
}
```

`default`s, `range`s and `regex`s are discarded when converting to Proto3 fields.

Concerto:

```
asset Vehicle {
  o String model default="F150"
  o String make default="FORD"
  o Integer year default=2016 range=[1990,]
  o String V5cID regex=/^[A-z][A-z][0-9]{7}/
}
```

Proto3:

```protobuf
message Vehicle {
  string model = 1;
  string make = 2;
  sint64 year = 3;
  string V5cID = 4;
}
```

## Classes

The following Concerto concepts map to `message` in Proto3:

- `concept`
- `asset`
- `participant`
- `transaction`
- `event`

When converting to Proto3 abstract classes are removed and inherited fields are explicitly stated in parent classes.

Identified class markers (`identified by <identifier>` , `identified`) are discarded.

If a class is identifiable in Concerto using `identified`, it gets an extra `_identifier` string field in the resulting Proto3.

Relationship identifier fields are simple `string`s.

Concerto:

```
concept Person {
  o String firstName
  o String lastName
}

concept Product identified {
  o String name
  o Double price
}

asset Vehicle identified by vin {
  o String vin
}

participant Customer identified by email {
  o String email
}

transaction Order identified by orderId {
  o String orderId
}

asset Orders {
  --> Order[] orders
}

event LateDelivery identified by eventId {
  o String eventId
}
```

Proto3:

```protobuf
message Person {
  string firstName = 1;
  string lastName = 2;
}

message Product {
  string _identifier = 1;
  string name = 2;
  double price = 3;
}

message Vehicle {
  string vin = 1;
}

message Customer {
  string email = 1;
}

message Order {
  string orderId = 1;
}

message Orders {
  repeated string orders = 1;
  string _identifier = 2;
}

message LateDelivery {
  string eventId = 1;
}
```

## Abstract classes

When converting to Proto3, abstract classes are discarded and their properties are placed into their child classes.

Concerto:

```
abstract concept Animal {
  o DateTime dob
}

concept Dog extends Animal {
  o String breed
}
```

Proto3:

```protobuf
message Dog {
  google.protobuf.Timestamp dob = 1;
  string breed = 2;
}
```

## Inheritance

Inheritance does not exist in the Proto3 world, however we can achieve similar functionality by:
- copying over the fields of parent `class`es to their children when converting them to Proto3 `message`s.
- having a utility Proto3 `message`, which can be referenced in other `message`s.

Concerto:

```
concept Animal {
   o String name
}

abstract concept Quadrupeds extends Animal {
   o Boolean tail
}

concept Tiger extends Quadrupeds {
   o Integer humansEaten optional
}

concept Zebra extends Quadrupeds {
   o Integer numStripes
}

concept Zoo {
   o Animal[] animals
}
```

Proto3:

```protobuf
message Animal {
  string name = 1;
}

message _Subclasses_of_class_Animal {
  oneof _class_oneof_Animal {
    Animal _subclass_of_class_Animal_Animal = 1;
    Tiger _subclass_of_class_Animal_Tiger = 2;
    Zebra _subclass_of_class_Animal_Zebra = 3;
  }
}

message _Subclasses_of_class_Quadrupeds {
  oneof _class_oneof_Quadrupeds {
    Tiger _subclass_of_class_Quadrupeds_Tiger = 1;
    Zebra _subclass_of_class_Quadrupeds_Zebra = 2;
  }
}

message Tiger {
  optional sint64 humansEaten = 1;
  bool tail = 2;
  string name = 3;
}

message Zebra {
  sint64 numStripes = 1;
  bool tail = 2;
  string name = 3;
}

message Zoo {
  repeated _Subclasses_of_class_Animal animals = 1;
}
```

## Enumerations

When converting from Concerto to Proto3, enumeration options need to be prepended with the name of the enumeration declaration, as they must be unique to the namespace.

Concerto:

```
enum AddressType {
  o residential
  o business
}
```

Proto3:

```protobuf
enum AddressType {
  AddressType_residential = 0;
  AddressType_business = 1;
}
```

## Namespaces

Concerto `namespace`s can be mapped to Proto3 `package`s, however the version needs to be transformed to be compliant with Proto3.

Concerto:

```
namespace org.accordproject.address@1.0.0
```

Proto3:

```protobuf
package org.accordproject.address.v1_0_0;
```

## Imports

You can import other Proto3 files using `import`. This differs to how Concerto `import`s work, as Proto3 needs a file specified.

Concerto:

```
import org.accordproject.address@1.0.0.{PostalAddress}

concept Office {
  o PostalAddress address
}
```

Proto3:

```protobuf
import "org.accordproject.address.v1_0_0.proto"

message Office {
  PostalAddress address = 1;
}
```

# Losses

The conversion to a Proto3 model is lossy i.e. you can’t convert a Proto3 model back to the same Concerto model that you started from. This is due to Proto3’s limitations stemming from the need to convert to a wide array of languages. As the primary reason why Proto3 is needed in our case is smaller packet sizes when communicating between services, the Concerto constraints and safeguards lost do not seem to be of high importance and can be implemented in the consumer code as validation manually as needed.

- Identifiable classes without an identifier (`identified`), get an `_identifier`
`string` inserted.
- The specified identifier in the identifiable classes (`identified by`) is lost.
- Relationship references are represented as `string`s. This is lossy, as the information about a field being an identifier is lost
- Decorators are lost.
- Concerto `default` s are lost.
- Inheritance of Concerto classes is lost. Instead we have explicit listing of properties inherited from the parent class, into our child class. References to parents are facilitated my a utility Proto3 `message`.
- Proto3 `repeatable` fields are also `optional`  i.e. you can’t have a required list (be it empty) constraint in Proto3, something which is present in Cicero.
- All the types of Concerto classes map to `message` in Proto3. The class distinctions are lost.
- Abstract classes are dropped.


packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
```diff
@@ -0,0 +1,349 @@
...
+
+    /**
+     * Transform a Concerto meta property into a Proto3 field rule.
+     * @param {Object} field - the Concerto meta property
+     * @return {Object} the Proto3 field rule
+     * @public
+     */
+    concertoToProto3FieldRule(field) {
+        if (field.isArray()) {
+            return 'repeated';
```
DianaLease, line 47: Might be good to add a comment here that arrays are implicitly optional in Proto3 so that it's clear why these two conditions are mutually exclusive.

DianaLease (commented):
Looks great! Left a small suggestion.

packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
stefanblaginov: Addressed ✅

DianaLease (approved):

packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    /**
+     * Visitor design pattern
+     * @param {Field} field - the object being visited
+     * @param {Object} parameters - the parameter
+     * @return {Object} the result of visiting or null
+     * @private
+     */
+    visitField(field, parameters) {
```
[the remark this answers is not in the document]
dselman: This limitation is troubling: 
```
Inheritance of Concerto classes is lost. Instead we have explicit listing of properties inherited from the parent class, into our child class.
```

What does this mean for models that have references to abstract types, similar to this?

```
abstract concept Animal {
   o String name
}

concept Tiger extends Animal {
   o Integer humansEaten optional
}

concept Zebra extends Animal {
   o Integer numStripes
}

concept Zoo {
   o Animal[] animals
}
```

packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
```diff
@@ -0,0 +1,369 @@
...
+                        (property) => {
+                            if (
+                                // If property starts with a $ and is not a $identifier.
+                                (
+                                    property.getName().charAt(0) === '$' &&
+                                    property.getName() !== '$identifier'
+                                ) ||
+                                // Or if property is an $identifier, but there already is another identifier defined.
+                                (
+                                    property.getName() === '$identifier' &&
```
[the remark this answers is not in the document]
stefanblaginov: Thanks @dselman ! Addressed ✅

packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
```diff
@@ -0,0 +1,368 @@
...
+                            }
+                            // All the other properties should be added to the list of properties to be written in the file.
+                            return property.accept(this, parameters);
+                        }
+                    )
+                    // Make sure that we don't have any unidentified fields.
+                    .filter(proto3FieldObject => proto3FieldObject !== undefined)
+                    // Write the fields, adding a Proto3 index to them.
+                    .forEach(
+                        ({ preposition, type, fieldName }, i) => {
```
dselman, line 276: I think the use of `I` here may be a problem, as it means that if the user reorders their list of properties in the CTO file (a non-breaking change) [or in a super-type CTO file] then the indexes used for the protobuf will all change (a breaking change for protobuf). Instead, I suggest that we sort the properties by name and use the sorted list to compute the protobuf index.

packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    /**
+     * Get the names of the children of a class.
+     * @param {string} className - the name of the class
+     * @param {Object[]} declarations - the declarations in scope
+     * @return {string[]} an array of the names of the children of the classes
+     * @public
+     */
+    getChildrenOfClass(className, declarations) {
```
dselman, line 112: I think you should maybe use:
https://github.com/accordproject/concerto/blob/93659e3b59370729ac62c466009f88a56fc32758/packages/concerto-core/lib/introspect/classdeclaration.js#L456

This will recursively return all specialisations of a given class (including the class itself).

Have you tested the code with multiple levels of inheritance?

packages/concerto-tools/lib/codegen/fromcto/protobuf/protobufvisitor.js
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    /**
+     * Check if a class has children.
+     * @param {string} className - the name of the class
+     * @param {Object[]} declarations - the declarations in scope
+     * @return {boolean} whether or not the class has children
+     * @public
+     */
+    doesClassHaveChildren(className, declarations) {
```
dselman, line 127: Subclasses is a better term than "children" for this.

dselman (changes requested):

dselman (changes requested):
I think we have an issue where we use fieldIndexes in the generated protobuf. E.g. when you reorder properties in your CTO file the field index in your protobuf will change. I **think** that is a breaking change for protobuf, but a non-breaking change for CTO.

```
enum Color {
  o RED
  o GREEN
  o BLUE
}
```

Should generate the same protobuf as:

```
enum Color {
  o GREEN
  o RED
  o BLUE
}
```

I think we can accomplish that by sorting the names of the properties before we iterate on them, so that the protobuf index is consistent and based on lexicographical order of the property names, rather than the order of declaration of the properties in the CTO file.

Make sense?

stefanblaginov:
> 

Got it! Addressed ✅

dselman (approved):
Bravo!

sstone1:
> I think we have an issue where we use fieldIndexes in the generated protobuf. E.g. when you reorder properties in your CTO file the field index in your protobuf will change. I **think** that is a breaking change for protobuf, but a non-breaking change for CTO.
> 
> 
> 
> ```
> 
> enum Color {
> 
>   o RED
> 
>   o GREEN
> 
>   o BLUE
> 
> }
> 
> ```
> 
> 
> 
> Should generate the same protobuf as:
> 
> 
> 
> ```
> 
> enum Color {
> 
>   o GREEN
> 
>   o RED
> 
>   o BLUE
> 
> }
> 
> ```
> 
> 
> 
> I think we can accomplish that by sorting the names of the properties before we iterate on them, so that the protobuf index is consistent and based on lexicographical order of the property names, rather than the order of declaration of the properties in the CTO file.
> 
> 
> 
> Make sense?

Won't you have a new issue now where adding a new field (a compatible Concerto change if it's optional) will change the field index for anything that comes after the new field alphabetically, which will be a breaking Protobuf change?

mttrbrts:
> Won't you have a new issue now where adding a new field (a compatible Concerto change if it's optional) will change the field index for anything that comes after the new field alphabetically, which will be a breaking Protobuf change?

You're right, but I don't believe that the aim is to maintain equivalence between the schema languages. The source of truth should be CTO.

mttrbrts:
@stefanblaginov Can you add a test case to cover the branch where there are no imports, please? I'd like to keep the coverage checker happy.
https://coveralls.io/builds/54094574/source?filename=packages%2Fconcerto-tools%2Flib%2Fcodegen%2Ffromcto%2Fprotobuf%2Fprotobufvisitor.js#L209

stefanblaginov:
>

sstone1:
> > Won't you have a new issue now where adding a new field (a compatible Concerto change if it's optional) will change the field index for anything that comes after the new field alphabetically, which will be a breaking Protobuf change?
> 
> You're right, but I don't believe that the aim is to maintain equivalence between the schema languages. The source of truth should be CTO.

I have no argument with the source of truth being CTO, but I think this is going to make it much harder to evolve a model that you want to use to build gRPC servers/client. The only way to keep the generated Protobuf compatible is to name your new properties so that they are at the end of the list - perhaps resulting in things like `ZZZnewProperty`. 

Perhaps we should support a decorator (either required or optional) for specifying the Protobuf field indexes?

```
concept Foo {
  @ProtobufIndex(1)
  o String propA
  @ProtobufIndex(2)
  o String propC
  @ProtobufIndex(44)
  o String propB
}
```

stefanblaginov:
> @sstone1 Here's my proposal to implement the decorator:
Have an `@Proto3ndexed` boolean decorator for the Concerto file as well. This would enable us to have an all-or-nothing approach towards indexes: either they're specified for all of the fields or for none. It would act as a guarantee that the resulting Proto3 file would be able to be extended safely and unambiguously. This is the problem that we will be trying to avoid:
Before:
```
concept Foo {
  @ProtobufIndex(1)
  o String propA
  @ProtobufIndex(2)
  o String propB
  o String propC
}
```
```protobuf
message Zebra {
  string propA = 1;
  string propB = 2;
  string propC = 3;
}
```
After:
```
concept Foo {
  @ProtobufIndex(1)
  o String propA
  @ProtobufIndex(2)
  o String propB
  o String propC
  @ProtobufIndex(3)
  o String propFoo
}
```
```protobuf
message Zebra {
  string propA = 1;
  string propB = 2;
  string propFoo = 3;
  string propC = 4; // This is remapped
}
```

stefanblaginov:
> @stefanblaginov Can you add a test case to cover the branch where there are no imports, please? I'd like to keep the coverage checker happy. https://coveralls.io/builds/54094574/source?filename=packages%2Fconcerto-tools%2Flib%2Fcodegen%2Ffromcto%2Fprotobuf%2Fprotobufvisitor.js#L209

@mttrbrts Addressed ✅

stefanblaginov:
@mttrbrts @sstone1 @dselman , should I go ahead and implement the field index decorator as proposed?

mttrbrts:
> @mttrbrts @sstone1 @dselman , should I go ahead and implement the field index decorator as proposed?

Let's get this code in so that we can start playing with it. We can add the index decorator in a subsequent PR.

Issue #552, opened 2022-11-29
Concerto assumes UTC for unqualified "Local Time" DateTime values


mttrbrts:
# Bug Report 🐛

When validating data against a model with the Concerto CLI, dates with the format `YYYY-MM-DD` are assumed to be in UTC. In general, this same issues is present for all DateTime values that are unqualified, i.e. they don't specify a zone designation. 

## Expected Behavior
When an unqualified date, or date-time value is provided (i.e. one without an explicit `Z` or offset component), users expect the date to be interpreted as in their local timezone.

This aligns with the ISO 8601 specification
https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)
> *Local time (unqualified)*
> If no UTC relation information is given with a time representation, the time is assumed to be in local time. 

## Current Behavior
For a user with a UTC-8 offset, a `DateTime` value of `2020-01-01` will be normalized to `2019-12-31T16:00:00.000-08:00`.  Note how the year, month, day and time have all changed. This is surprising to users.

This is caused by an explicit assumption that the default offset for values should be `0`.
https://github.com/accordproject/concerto/blob/main/packages/concerto-core/lib/serializer/jsonpopulator.js#L94

## Possible Solution

1. Explicitly detect usage of date-times without a zone designation, and to assume an offset that matches the local offset of the user's machine/server. 
2. Use of the `--utcOffset` flag in the CLI should always allow the offset to be set explicitly.

## Steps to Reproduce

1. Set your time zone to PST (GMT-8). 
3. Create two files:
test.cto:
```cs
namespace test@1.0.0

concept C {
  o DateTime d
}
```
test.json:
```json
{ "$class": "test@1.0.0.C", "d":"2020-01-01" }
```
4. Run 
```bash
$ concerto validate --model test.cto --input obj.json
8:47:23 AM - INFO: Input is valid
{"$class":"test@1.0.0.C","d":"2019-12-31T16:00:00.000-08:00"}
```

## Discussion
Fixing this bug could be sensitive. Existing users may rely on this behaviour, even though I believe that it should be considered a bug.

DS-AdamMilazzo:
> When an unqualified date, or date-time value is provided (i.e. one without an explicit Z or offset component), users expect the date to be interpreted as in their local timezone. This aligns with the ISO 8601 specification
> https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)
> > Local time (unqualified)
[...]

PR #553, opened 2022-11-29
fix(core): deal with unqualified date times in local offsets

mttrbrts:
This PR adds a new option to the `JSONGenerator` and `JSONPopulator` classes (and the `Serializer`) called `strictQualifiedDateTimes`. This allows client applications to drop support for unqualified, and ambiguous dates/datetimes.

This solves #552 in a backward-compatible way by not-allowing dates and date-times that don't have an offset. This also has the side-effect of making the `utcOffset` parameter only relevant to serialization; i.e. it becomes a rendering option. We always trust the offset provided when `strictQualifiedDateTimes` is `true`

When `strictQualifiedDateTimes` is `true`, only the following date-time formats are accepted.
- `YYYY-MM-DDTHH:mm:ssZ`
- `YYYY-MM-DDTHH:mm:ss.SZ`
- `YYYY-MM-DDTHH:mm:ss.SSZ`
- `YYYY-MM-DDTHH:mm:ss.SSSZ`
- `YYYY-MM-DDTHH:mm:ss+HH:mm`
- `YYYY-MM-DDTHH:mm:ss.S+HH:mm`
- `YYYY-MM-DDTHH:mm:ss.SS+HH:mm`
- `YYYY-MM-DDTHH:mm:ss.SSS+HH:mm`
- `YYYY-MM-DDTHH:mm:ss-HH:mm`
- `YYYY-MM-DDTHH:mm:ss.S-HH:mm`
- `YYYY-MM-DDTHH:mm:ss.SS-HH:mm`
- `YYYY-MM-DDTHH:mm:ss.SSS-HH:mm`

There are lots of formats that are no-longer accepted, for example (see the test for the full set): 
- `YYYY-MM-DD`
- `YYYY-MM-DDTHH:mm:ss`
- `YYYY-MM-DDTHH:mm:ss.S`
- `YYYY-MM-DDTHH:mm:ss.SS`
- `YYYY-MM-DDTHH:mm:ss.SSS`

### Changes

- Adds lots of format tests for the serialization/deserialization logic.
- Adds a new constructor argument called `strictQualifiedDateTimes` to `JSONGenerator` and `JSONPopulator`.

### Flags

- This is a candidate for becoming the default behaviour in future.
- Currently, the existing behaviour is retained for backwards compatibility. This does permit some unexpected behaviour as described in #552.
- When using the `strictQualifiedDateTimes` flag, this has the effect of removing support dates (without explicit times). 
- Client applications such as Cicero, should take responsibility for (de)normalizing to this format.

### Author Checklist

### DIFF
--- a/packages/concerto-cli/lib/commands.js
+++ b/packages/concerto-cli/lib/commands.js
@@ -122,7 +122,7 @@ class Commands {
             const factory = new Factory(modelManager);
             const serializer = new Serializer(factory, modelManager);
 
-            const object = serializer.fromJSON(json);
+            const object = serializer.fromJSON(json, options);
             return JSON.stringify(serializer.toJSON(object, options));
         }
     }
--- a/packages/concerto-core/lib/serializer.js
+++ b/packages/concerto-core/lib/serializer.js
@@ -15,18 +15,15 @@
 'use strict';
 
 const { TypedStack } = require('@accordproject/concerto-util');
-const DateTimeUtil = require('./datetimeutil');
 const Globalize = require('./globalize');
 const JSONGenerator = require('./serializer/jsongenerator');
 const JSONPopulator = require('./serializer/jsonpopulator');
 const Typed = require('./model/typed');
 const ResourceValidator = require('./serializer/resourcevalidator');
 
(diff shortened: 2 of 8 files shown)


dselman (approved):
Great!

PR #546, opened 2022-11-10
feat(concerto): Implement Faceted Primitive Types in Concerto

stefanblaginov:
# Closes #433

Faceted primitive types (aka `scalar`s) permit reuse of primitive fields. The desired effect it to allow developers to DRY out Concerto model definitions and improve development ergonomics.

A `scalar` can be defined using such a syntax:
```
scalar <name> extends <primitive Concerto type> <optional primitive Concerto type meta data>
```

Example:
```
scalar SSN extends String default="000-00-0000" regex=/\d{3}-\d{2}-\{4}+/
```

The `scalar` can be referenced in the same way as any other class:
```
concept Person identified by ssn {
    o SSN ssn
}
```

The above is equivalent to:
```
concept Person identified by ssn {
    o String ssn default="000-00-0000" regex=/\d{3}-\d{2}-\{4}+/
}
```

## TODOs
- Passing unit tests.
- Functioning `concerto-cli` `concerto print`.
- Functioning `concerto-cli` `concerto compare`.
- Updated Concerto VS Code extension.

## CLI-requests and responses:

This is the CTO that we’re using for the tests:

```
namespace org.acme

scalar BooleanScalar extends Boolean
scalar BooleanScalarWithDefault extends Boolean default=true

scalar IntegerScalar extends Integer
scalar IntegerScalarWithDefault extends Integer default=42
scalar IntegerScalarWithDefaultAndRange extends Integer default=42 range=[1,]

scalar LongScalar extends Long
scalar LongScalarWithDefault extends Long default=42
scalar LongScalarWithDefaultAndRange extends Long default=42 range=[1,]

scalar DoubleScalar extends Double
scalar DoubleScalarWithDefault extends Double default=42.5
scalar DoubleScalarWithDefaultAndRange extends Double default=42.5 range=[1.0,]

scalar StringScalar extends String
scalar StringScalarWithDefault extends String default="something"
scalar StringScalarWithDefaultAndRegex extends String default="something" regex=/.*/

scalar DateTimeScalar extends DateTime
scalar DateTimeScalarWithDefault extends DateTime default="2022-11-18"

concept Scalars {
    o BooleanScalar booleanScalar
    o BooleanScalarWithDefault booleanScalarWithDefault
    o IntegerScalar integerScalar
    o IntegerScalarWithDefault integerScalarWithDefault
    o IntegerScalarWithDefaultAndRange integerScalarWithDefaultAndRange
    o LongScalar longScalar
[...]

PR #557, opened 2022-12-02
fix(*): update metamodel cto to reflect metamodel json

sstone1:
The changes in #546 did not update the CTO representation of the metamodel we ship in the code, and so this change updates the CTO to match the new metamodel JSON. It also updates the metamodel JSON with the missing `@DotNetNamespace` decorator that has been removed.

This metamodel management process is broken and needs a revisit, but this will fix it for now.


mttrbrts (approved):

dselman (commented):
I think this breaks web packing

PR #577, opened 2022-12-16
feat(core): infer $class when obvious

### LINKED ISSUE #542: Concerto should allow $class to be omitted or abbreviated when it's obvious from context
The need to provide a full $class declaration on every subobject in an object tree adds a lot of redundant information to a serialized Concerto object. In nearly all cases it could be removed or abbreviated without loss of information. This has a few benefits:
- Less redundant and more aesthetically pleasing data
- Closer to the "just the data, please" ideal
- Easier to programmatically manage changes to the version of the data, and less likely to end up in Frankenstein scenarios where different parts of an object have different model versions, if the version is only declared in one place

# Feature Request #
The $class property should not be required where it can be inferred from context, and an abbreviated form of the $class property should be allowed when it can be mostly inferred from context. The only times a full $class property is really needed are 1) on the top-level object, and 2) when referencing a type from another namespace or version. Otherwise, the $class property could usually be omitted entirely, even in some polymorphic cases. In the polymorphic cases where a $class property is needed, it's rare to need a full $class; usually the type name only would suffice, without the namespace and version, which can be inferred from the containing object.

If there is only one $class attribute, at the top level, it can easily be altered to change the version. If there's a $class attribute on every subobject, the user would need to recursively update all of them. Done programmatically, this requires significantly more complicated logic, and if they aren't so careful to update all the $class attributes, then they can easily get objects where some field values have version X and other field values have version Y. The object could no longer be said to come from a single model version, but would be a mix of multiple model versions. Perhaps in rare cases that might be desirable, but it should be something that's hard to do by accident.

I'll note that this request is about deserializing and validating objects. It would also be nice if Concerto would produce a more compact serialized format as well, but that is outside the scope of this request. (#482 is somewhat related, describing a "final" feature to allow Concerto to omit $class in one specific case. I notice that the title of that issue was changed earlier today to something broader, but the body is still about the "final" feature.)

## Possible Solution
1. A full $class is required at the top level. Based on this, you know the concept of the top-level object and the declared type of every field in that concept.
2. In each data field (or each element if the field is an array), if the $class is omitted, it's assumed to be the same as the declared field type. (If the declared field type is abstract, that's an error.) If $class is just a type name, the namespace and version are assumed to be the same as the declared field type.
3. The logic applies recursively.

I'll note that this is backwards-compatible with the existing, more verbose specification that requires a full $class property on every sub-object.

mttrbrts:
# Closes #542 
This change allows us to validate and deserialize JSON objects where the root type is known, but where no $class properties for nested objects are provided. This will work for basic cases where there is only one possible concrete subtype. 

For example, an object such as this will validate provided that we tell the validator what the root type is.

```json
{
    "$class" : "test.Vehicle",
    "color": "red",
    "wheels" : [{
        "brand" : "Michelin"
    }],
    "owner": {
        "email": "foo@bar.com"
    }
}
```

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/serializer/objectvalidator.js
+++ b/packages/concerto-core/lib/serializer/objectvalidator.js
@@ -120,6 +120,15 @@ class ObjectValidator {
 
         const obj = parameters.stack.pop();
 
+        if (!obj.$class){
+            // Choose the only concrete subclass if it exists
+            const subclasses = classDeclaration.getAssignableClassDeclarations()
+                .filter(clazz => !clazz.isAbstract());
+            if (subclasses.length === 1){
+                obj.$class = subclasses[0].getFullyQualifiedName();
+            }
+        }
+
         if(this.concerto.isIdentifiable(obj)) {
             parameters.rootResourceIdentifier = this.concerto.getFullyQualifiedIdentifier(obj);
         }
@@ -326,6 +335,13 @@ class ObjectValidator {
         else {
             // a field that points to a transaction, asset, participant...
             let classDeclaration = this.concerto.getModelManager().getType(field.getFullyQualifiedTypeName());
+            if (!obj.$class){
+                const subclasses = classDeclaration.getAssignableClassDeclarations()
+                    .filter(clazz => !clazz.isAbstract());
+                if (subclasses.length === 1){
+                    obj.$class = subclasses[0].getFullyQualifiedName();
+                }
+            }
             try {
                 classDeclaration = this.concerto.getModelManager().getType(obj.$class);
             } catch (err) {
--- a/packages/concerto-core/lib/serializer/objectvalidator.js
+++ b/packages/concerto-core/lib/serializer/objectvalidator.js
@@ -121,7 +121,7 @@ class ObjectValidator {
         const obj = parameters.stack.pop();
 
         if (!obj.$class){
-            // Choose the only concrete subclass if it exists
+            // Choose the only concrete class/subclass if it exists
             const subclasses = classDeclaration.getAssignableClassDeclarations()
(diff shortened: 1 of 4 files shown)


packages/concerto-core/lib/serializer/objectvalidator.js
```diff
@@ -120,6 +120,15 @@ class ObjectValidator {
 
         const obj = parameters.stack.pop();
 
+        if (!obj.$class){
+            // Choose the only concrete subclass if it exists
+            const subclasses = classDeclaration.getAssignableClassDeclarations()
+                .filter(clazz => !clazz.isAbstract());
```
DS-AdamMilazzo, line 126: Does the getAssignableClassDeclarations method also return the base type? If so, the comment should probably be corrected to say "class" instead of "subclass". If not, I think the code should also handle the more common case where the actual type is equal to the declared type, and not a subclass of it.

mttrbrts: Thanks Adam. `getAssignableClassDeclarations` does return itself too, so I've updated the comments to reflect this.

dselman (changes requested):

mttrbrts:
Closing in light of @dselman's comments.
https://github.com/accordproject/concerto/issues/542#issuecomment-1362057631

DS-AdamMilazzo:
I agree with Dan's comment about deserialization. The approach I recommend is to simply treat an object as an instance of the declared field type if it's not annotated. This should be even simpler to implement than this PR (because you don't have to care about how many concrete assignable types there may be), more general, and would not break in the case Dan describes.

PR #579, opened 2022-12-30
feat(core): scalar inline defaults

### LINKED ISSUE #580: Concerto scalar inline defaults
# Feature Request 🛍️

> 💡 Proposed by @DS-AdamMilazzo here: https://github.com/accordproject/concerto/pull/546#issuecomment-1334137210.

We would like to be able to extend a scalar’s `default` property inside a class after its definition.

Example:

```
scalar SSN extends String regex=/\d{3}-\d{2}-\d{4}+/

concept Person identified by ssn {
    o SSN ssn default="999-99-9999"
}
```

This would be effectively the same as:

```
scalar SSN extends String default="999-99-9999" regex=/\d{3}-\d{2}-\d{4}+/

concept Person identified by ssn {
    o SSN ssn
}
```
or
```
concept Person identified by ssn {
    o String ssn default="999-99-9999" regex=/\d{3}-\d{2}-\d{4}+/
}
```

# Design questions

### `default` is not part of the scalar type

If the `default` meta property **is not** part of the scalar type, but simply an add-on (similar to `optional`), then the inline overriding of the `default` **would not be** considered a violation of the type and the following would be possible:

```
scalar SSN extends String default="000-00-0000" regex=/\d{3}-\d{2}-\d{4}+/

concept Person identified by ssn {
    o SSN ssn default="999-99-9999"
}
```

Drawback of this approach: `Boolean` `scalar`s become useless, as their only meta property currently is the `default`.

Which would effectively be the same as:

```
concept Person identified by ssn {
    o String ssn default="999-99-9999" regex=/\d{3}-\d{2}-\d{4}+/
}
```

### `default` is part of the scalar type

If the `default` meta property **is** part of the scalar type, then the inline overriding of the `default` **would be** considered a violation of the type and the following would throw an error:

```
scalar SSN extends String default="000-00-0000" regex=/\d{3}-\d{2}-\d{4}+/

concept Person identified by ssn {
    o SSN ssn default="999-99-9999"
}
```

This would force the developer to define a new scalar with the desired `default`:

```
scalar SSN extends String default="000-00-0000" regex=/\d{3}-\d{2}-\d{4}+/

concept Person identified by ssn {
    o SSN ssn
}

scalar OtherSSN extends String default="999-99-9999" regex=/\d{3}-\d{2}-\d{4}+/

concept OtherPerson identified by ssn {
    o OtherSSN ssn
}
```

Drawback of this approach: The developer may need to define an entirely new `scalar` if he wants to have a different `default`.

One think to note when deciding on a path forward is that we can always implement the "`default` is part of the scalar type" approach and then later transition to the "`default` is not part of the scalar type" as a non-braking change, while the reverse is not possible.

## What about `range`?

Although not in the scope of this issue, it may be useful to look forward to the question of extending the `range` meta property.

My opinion is that `range` should be part of the `scalar` type (otherwise `scalar`s lose their purpose). What could be possible would be for the child range to always be more restrictive than the parent one. This, however, would mean that we would need to check this greater restrictiveness.

Example of a valid `range` extension:

```
scalar Age extends Integer range=[0,]

concept Person {
    o Age age
}

concept Adult {
    o Age age range=[18,]
}
```

Example of an invalid `range` extension:

```
scalar Age extends Integer range=[0,]

concept Person {
    o Age age
}

concept Unborn {
    o Age age range=[-1,]
}
```

stefanblaginov:
# Closes #580 

> 💡 Proposed by Adam Milazzo.

We would like to be able to extend a scalar’s `default` property inside a class after its definition.

Example:

```
scalar SSN extends String regex=/\d{3}-\d{2}-\{4}+/

concept Person identified by ssn {
    o SSN ssn default="999-99-9999"
}
```

This would be effectively the same as:

```
scalar SSN extends String default="999-99-9999" regex=/\d{3}-\d{2}-\{4}+/

concept Person identified by ssn {
    o SSN ssn
}
```

Additionally, if a `default` is already defined, we would like to be able to override it inline.

Example:

```
scalar SSN extends String default="000-00-0000" regex=/\d{3}-\d{2}-\{4}+/

concept Person identified by ssn {
    o SSN ssn default="999-99-9999"
}
```

This would be effectively the same as:

```
scalar SSN extends String default="999-99-9999" regex=/\d{3}-\d{2}-\{4}+/

concept Person identified by ssn {
    o SSN ssn
}
```

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/field.js
+++ b/packages/concerto-core/lib/introspect/field.js
@@ -195,6 +195,11 @@ class Field extends Property {
         }
 
         fieldAst.name = this.ast.name;
+
+        if (this.getDefaultValue()) {
+            fieldAst.defaultValue = this.getDefaultValue();
+        }
+
         this.scalarField = new Field(this.getParent(), fieldAst);
         return this.scalarField;
     }
--- a/packages/concerto-cto/test/cto/scalar.cto
+++ b/packages/concerto-cto/test/cto/scalar.cto
@@ -66,3 +66,15 @@ concept Person identified by ssn {
   o SSN ssn
   o String givenName
 }
+
+scalar ExtendedSSN extends String regex=/\d{3}-\d{2}-\{4}+/
+
+concept PersonWithExtendedSsn identified by ssn {
+  o ExtendedSSN ssn default="999-99-9999"
+  o String givenName
+}
+
+concept PersonWithOverriddenExtendedSsn identified by ssn {
+  o SSN ssn default="999-99-9999"
+  o String givenName
+}
--- a/packages/concerto-cto/test/cto/scalar.json
+++ b/packages/concerto-cto/test/cto/scalar.json
@@ -351,6 +351,71 @@
                 "$class": "concerto.metamodel@1.0.0.IdentifiedBy",
                 "name": "ssn"
             }
+        },
+        {
+            "$class": "concerto.metamodel@1.0.0.StringScalar",
+            "validator": {
+                "$class": "concerto.metamodel@1.0.0.StringRegexValidator",
+                "pattern": "\\d{3}-\\d{2}-\\{4}+",
+                "flags": ""
+            },
+            "name": "ExtendedSSN"
+        },
+        {
+            "$class": "concerto.metamodel@1.0.0.ConceptDeclaration",
+            "name": "PersonWithExtendedSsn",
+            "isAbstract": false,
+            "properties": [
+                {
+                    "$class": "concerto.metamodel@1.0.0.ObjectProperty",
+                    "name": "ssn",
+                    "type": {
+                        "$class": "concerto.metamodel@1.0.0.TypeIdentifier",
+                        "name": "ExtendedSSN"
+                    },
+                    "isArray": false,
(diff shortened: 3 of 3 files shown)


mttrbrts (changes requested):
Can you add a test to demonstrate that the default value from the lowest type in the tree is chosen when validating/deserialising an object, please?

Also, it's worth stating that this change has the effect of making `default` an untyped constraint. For example, when a user specifies a `default` value, we don't check that it is compatible with the `default` value of its parent type.

This argument makes more sense for modifers such as `regex` and `range` where it's feasible that a subtype could apply a more restrictive version of the modifer. For example, a type `T` has constraint `range=[0,100]`, and subtype `S`, such that `S <: T` has the constraint `range=[20,30]`.  In this case, all values of `S` are guaranteed to also have type `T`, and satisfy the subtype relationship.

This argument does not work for the `default` modifier. I'm not saying that this is wrong, just that this distinction has never been explicit previously.

mttrbrts:
@dselman this work is on hold. It's still not clear to me that constraint specialization is a desirable feature.

Issue #572, opened 2022-12-09
Identifier Regex is too restrictive.


mttrbrts:
# Bug Report 🐛
As originally reported by @DS-AdamMilazzo.

Regarding the regex for a valid Concerto field name in the JSON representation of a model:

```regex
^(?!null|true|false)(\p{Lu}|\p{Ll}|\p{Lt}|\p{Lm}|\p{Lo}|\p{Nl}|\$||\\u[0-9A-Fa-f]{4})(?:\p{Lu}|\p{Ll}|\p{Lt}|\p{Lm}|\p{Lo}|\p{Nl}|\$||\\u[0-9A-Fa-f]{4}|\p{Mn}|\p{Mc}|\p{Nd}|\p{Pc}|\u200C|\u200D)*$
```
Is it really necessary to avoid the names "null", "true", and "false"?  There's no ambiguity between "null" and null in JSON, anyway. The only trouble may be in the Concerto language, but this regex applies to the JSON format.

DS-AdamMilazzo:
More broadly, I think that Concerto should enforce different formats in different contexts. First of all, I'll note that the Concerto metamodel (where these regexes are defined) is supposed to be for the JSON format and not the Concerto (.cto) format, which may well have different rules.

1. **Field names and enum values** - There should be no restrictions whatsoever in my opinion. Any legal JSON property name or string should be a legal field name or enum value in Concerto, except maybe empty strings if they're too hard to support.
2. **Type names (concepts, etc)** - I think these should be restricted. In fact, I would restrict them even more than you do above. I would not allow arbitrary characters via escaping. The usual `^[_A-Za-z][_0-9A-Za-z]*$` format should be fine. As for trying to exclude certain reserved words (e.g. "concept"), I think that's okay (if done correctly) but I also think it's unnecessary.
3. **Namespace names** - I think these should also be restricted more than you currently do. You currently allow namespace names like `foo\u0020\bar`. I'd limit them to dot-separated strings where each piece has to match `^[_0-9A-Za-z]+$` (or maybe `^[_A-Za-z][_0-9A-Za-z]*$`) or something.

As for the Concerto (.cto) language, I think the \uXXXX escaping mechanism that you have today is okay (except where it doesn't fully work), but it'd be nicer if you could simply quote things, e.g. `"my value"` versus `my\u0020value`.

mttrbrts:
## Regular Expressions

First, some definitions.

### Unicode Categories
[...]

PR #581, opened 2023-01-05
feat(core): relax reserved property names

mttrbrts:
Partially addresses Issue #572 

This PR relaxes the parsing and core API restrictions on identifier names. The following names are now permitted:
- `false`
- `true`
- `null`
- `1`
- `1abc`

### Changes

- Adds explicit test cases which give good coverage of valid & invalid identifiers
- Changes Parser rules to allow NullLiteral and BooleanLiterals as identifiers
- Changes Parser rules to allow identifiers to begin with UnicodeNumber values
- Removes the explicit reserved keyword check in the core API. This was only introduced recently to address a ReDOS limitation when generating sample JSON files regexes (negative lookahead was not supported). #484 

### Flags

- I believe that this should be considered a non-breaking change, as it loosens the constraints on identifiers.

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/property.js
+++ b/packages/concerto-core/lib/introspect/property.js
@@ -29,8 +29,6 @@ if (global === undefined) {
 /* eslint-enable no-unused-vars */
 
 
-const RESERVED_PROPERTY_NAMES = ['null', 'true', 'false'];
-
 /**
  * Property representing an attribute of a class declaration,
  * either a Field or a Relationship.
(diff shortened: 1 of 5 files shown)


mttrbrts:
@Jonathan-Casey

dselman (approved):
I assume imposing the identifier constraints on the AST will be in a follow-up PR?

mttrbrts:
> I assume imposing the identifier constraints on the AST will be in a follow-up PR?

Yes, that's right

PR #590, opened 2023-01-17
test(core): fix system time offset for tests

mttrbrts:
Unit tests would previously fail when run in an environment with a non-zero UTC offset.
This change explicitly sets the offset to UTC+0 so that the unit tests are more stable.


dselman (approved):

dselman (commented):
Do we also want to update package.json to set TZ?

PR #591, opened 2023-01-17
fix(core): enforce identifier validation on AST JSON

mttrbrts:
Related to #581 

### Changes
- Modifies the Core API to validate identifiers for Class Declarations, Properties, Enum values, and Namespace parts.
- Also, loosens a restriction in the code that prevented all user-defined properties beginning with a `$`. This now aligns with the metamodel.

### Flags
- This is a bug fix, to address a gap in previously released code, however, it has the potential to break clients who relied upon the previous behaviour.
- The ID_REGEX is duplicated between the metamodel and the API, ideally, we should reuse the same definition. One solution is to validate the JSON AST against the metamodel using Concerto. However, for simplicity I've followed the existing pattern of implementing semantic checks in the introspection API.

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/classdeclaration.js
+++ b/packages/concerto-core/lib/introspect/classdeclaration.js
@@ -79,6 +79,12 @@ class ClassDeclaration extends Decorated {
     process() {
         super.process();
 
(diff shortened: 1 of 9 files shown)


dselman (approved):
LGTM!

PR #600, opened 2023-02-06
feat(core): add support for Records types

### LINKED ISSUE #447: Record type to express key/value pairs
# Feature Request 🛍️

This proposes the addition of a Record type into the Concerto syntax to be able to express key/value pairs in an object. For example a `Record<String, Foo>` would express a key/value pair where the keys of the objects are `String`'s and the values of this object are `Foo` objects.

You can look more at TypeScript's Record type documentation here: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type, which I think is good inspiration for what should be possible here.

> EDIT: The [Design Document for this feature is in the Wiki](https://github.com/accordproject/concerto/wiki/Aggregate-Types-Design)

## Use Case

Here are a variety of different use cases that I think we should support:

Example 1:
```
concept Foo {
  o String propName
}

concept Example {
  o Record<String, Foo> fooDictionary
}

// An Example object could look like:
// {
//   fooDictionary: {
//     key: {
//       propName: 'Hello World'
//     }
//   }
// }
```

Example 2:
```
enum Keys {
  Some,
  Enum,
  Values
}

concept Foo {
  o String propName
}

concept Example {
  o Record<Keys, Foo> fooDictionary
}

// An Example object could look like:
// {
//   fooDictionary: {
//     Some: {
//       propName: 'Hello World'
//     },
//     Enum: {
//       propName: 'Other Hello World'
//     }
//   }
// }
```

Example 3:
```
concept Example {
  o Record<String, Boolean> fooDictionary
}

// An Example object could look like:
// {
//   fooDictionary: {
//     key: true
//   }
// }
```

## Possible Solution

Not sure about implementation here, I would assume that we would at least need to be able to support the concept of generics within the language as a pre-requisite to this.

I also think it would be good to constrain the Record type to only support `String` or `String` derivates (e.g., Enum) as keys for a Record type. This is similar to what TypeScript does, and translates well to JSON. This does limit Record's functionality in terms of what is possible in other languages like Java or C# which could use classes as key types, but I think this constraint is desirable.

## Context

Currently describing key/values pairs where keys are user generated content is currently not possible in Concerto. A Record type would help solve that problem. We see this when trying to describe variables for example, where variable names are defined by the user.

In these scenarios it's still very useful to developers at development time to have a representation of these Key/Value pairs, even if they aren't fully type safe, because a developer needs to understand what is possibly on this property of a concept.

mttrbrts:
# Closes #447 

### Changes

- <ONE>
- <TWO>

### Flags

- <ONE>
- <TWO>

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist


packages/concerto-core/lib/introspect/recordfield.js
```diff
@@ -0,0 +1,138 @@
...
+            classDecl.getModelFile().resolveType( 'property ' + this.getFullyQualifiedName(), this.valueType);
+        }
+    }
+
+    /**
+     * Returns the fully qualified type name of a property
+     * @return {string} the fully qualified type of this property
+     */
+    getFullyQualifiedTypeName() {
+        return `Record<${this.keyType}, ${this.valueType}>`;
```
dselman, line 85: Moderately sure that we have code that assumes FQNs are namespace@version.TypeName...

packages/concerto-core/lib/introspect/property.js
```diff
@@ -233,6 +233,14 @@ class Property extends Decorated {
     isPrimitive() {
         return ModelUtil.isPrimitiveType(this.getType());
     }
+
+    /**
+     * Returns true if this property is an aggregate type.
+     * @return {boolean} true if the property is an aggregate type.
+     */
+    isAggregate() {
```
dselman, line 241: isRecord?

packages/concerto-core/lib/introspect/modelfile.js
```diff
@@ -290,7 +290,7 @@ class ModelFile extends Decorated {
      */
     resolveType(context, type, fileLocation) {
         // is the type a primitive?
-        if(!ModelUtil.isPrimitiveType(type)) {
+        if(!ModelUtil.isPrimitiveType(type) && !ModelUtil.isAggregateType(type)) {
```
dselman, line 293: This check seems surprising. Why would a Record<string,*> be a primitive?

packages/concerto-core/lib/introspect/recordfield.js
```diff
@@ -0,0 +1,138 @@
...
+        this.map = new Map();
+    }
+
+    /**
+     * Process the AST and build the model
+     * @throws {IllegalModelException}
+     * @private
+     */
+    process() {
+        super.process();
```
dselman, line 56: Properties have a `type`. What it the type for a Record? We seem to be doing something tricky here...?

packages/concerto-core/lib/introspect/recordfield.js
```diff
@@ -0,0 +1,138 @@
...
+     */
+    isTypeEnum() {
+        return false;
+    }
+
+    /**
+     * Returns true if this property is an aggregate type.
+     * @return {boolean} true if the property is an aggregate type.
+     */
+    isAggregate() {
```
dselman, line 133: I think this is confusing now for arrays.

dselman: Also, can we not have an array of Sets or Records?

packages/concerto-core/test/serializer.js
```diff
@@ -78,6 +78,11 @@ describe('Serializer', () => {
         o String newValue
         }
 
+        concept Dictionaries {
+            o Record<String, String> simpleDictionary optional
+            // o Record<String, SampleParticipant> complexDictionary optional
```
dselman, line 83: Note that Set<ComplexType> requires identity for ComplexType, so should only be supported for Identifiables.

packages/concerto-metamodel/lib/metamodel.cto
```diff
@@ -30,6 +30,7 @@ concept Range {
 concept TypeIdentifier {
   o String name
   o String namespace optional
+  o TypeIdentifier[] typeParameters optional
```
dselman, line 33: How do typeParameters (optionally) contribute to type identity?

dselman (commented):
An updated design/spec with examples would be very useful.

packages/concerto-core/lib/introspect/property.js
mttrbrts: I intended this to cover the case for `Set`, `Union`, `Array` in future too

packages/concerto-core/lib/introspect/recordfield.js
mttrbrts: We can, but arrays are something separate in this prototype. Array is not an aggregate, but a modifier on any type (including `Record`).

We could also add `Array` as an Aggregate type, as an alternative to `[]` for backwards-compatibility?

packages/concerto-metamodel/lib/metamodel.cto
mttrbrts: Ah, this is unused. An artifact from an earlier iteration
```suggestion
```

packages/concerto-cto/test/parserMain.js
```diff
@@ -36,7 +36,7 @@ function getCTOFiles() {
     const files = fs.readdirSync('./test/cto');
 
     files.forEach(function(file) {
-        if(file.endsWith('.cto')) {
+        if(file.endsWith('records.cto')) {
```
Jonathan-Casey, line 39: FYI back this out

PR #607, opened 2023-02-15
feat(tools): enum, regex, referential type feature for csharp codegen

ragi-dayananda:
### Changes
 - Add regex annotation for csharp codegen when regex configured for a field
 - UUID nullable type bug fix
 - Make optional string properties as nullable string. 
 - Custom enum value support with AcceptedValue decorator
 - Handle referential type (-->) in csharp codegen

### Author Checklist


packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
```diff
@@ -283,8 +284,17 @@ class CSharpVisitor {
 
         let fieldType = externalFieldType ? externalFieldType : this.getFieldType(field);
 
+        if (fieldType === 'String') {
```
dselman, line 287: I worry that we might be using an `externalFieldType` here...

packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
```diff
@@ -283,8 +284,17 @@ class CSharpVisitor {
...
+            let validator = field.getValidator();
```
dselman, line 288: can be `const`?

packages/concerto-tools/lib/codegen/fromcto/csharp/csharpvisitor.js
ragi-dayananda: yes, we might. But external type can be UUID scalar type. We don't want to add regex decorator for UUID.

dselman (approved):
Looks great!

PR #613, opened 2023-02-27
feat(map): add initial implementation for concerto map type

Jonathan-Casey:
Not representative of finished article - Initial Implementation. Requesting feedback / comments.

Please ignore formatting anomalies due to editor, will fix.

### DIFF
--- a/packages/concerto-core/lib/introspect/field.js
+++ b/packages/concerto-core/lib/introspect/field.js
@@ -62,24 +62,32 @@ class Field extends Property {
         this.validator = null;
 
         switch (this.getType()) {
-        case 'Integer':
-        case 'Double':
-        case 'Long':
-            if (this.ast.validator) {
-                this.validator = new NumberValidator(
-                    this,
-                    this.ast.validator
-                );
-            }
-            break;
-        case 'String':
-            if (this.ast.validator) {
-                this.validator = new StringValidator(
-                    this,
-                    this.ast.validator
-                );
-            }
-            break;
+            case 'Integer':
+            case 'Double':
+            case 'Long':
+                if (this.ast.validator) {
+                    this.validator = new NumberValidator(
+                        this,
+                        this.ast.validator
+                    );
+                }
+                break;
+            case 'String':
+                if (this.ast.validator) {
+                    this.validator = new StringValidator(
+                        this,
+                        this.ast.validator
+                    );
+                }
+                break;
+            case 'Map':
+                if (this.ast.validator) {
+                    this.validator = new StringValidator(
+                        this,
+                        this.ast.validator
+                    );
+                }
+                break;
         }
 
         if (this.ast.defaultValue) {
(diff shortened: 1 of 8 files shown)


packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1397,6 +1398,43 @@ LongFieldDeclaration
...
+        ...buildRange(location())
+      };
+      if (decorators.length > 0) {
+        result.decorators = decorators;
+      }
+      return result;
+    }
+
+MapDeclarationBody
+  = decls:MapPropertyDeclaration* {
```
dselman, line 1418: This looks like a "zero or more" whereas I'd expect the Map declaration to have two child elements: key and value.

packages/concerto-cto/lib/printer.js
```diff
@@ -104,6 +104,9 @@ function typeFromMetaModel(mm){
     case `${MetaModelNamespace}.ObjectProperty`:
         result += ` ${mm.type.name}`;
         break;
+    case `${MetaModelNamespace}.MapProperty`:
```
dselman, line 107: These are not properties, in the sense we use it for the properties of a class (a name and a type, and meta-properties). These are the definitions of the type for the key and the type for the value.

packages/concerto-metamodel/lib/metamodel.cto
```diff
@@ -135,6 +135,20 @@ concept StringRegexValidator {
   o String flags
 }
 
+concept MapDeclaration extends Declaration {
+  o Boolean isAbstract default=false
```
dselman, line 139: I don't think the first three make sense for a map

packages/concerto-metamodel/lib/metamodel.cto
```diff
@@ -135,6 +135,20 @@ concept StringRegexValidator {
...
+  o Identified identified optional
+  o TypeIdentifier superType optional
+	o MapProperty key 
+	o MapProperty value 
+}
+
+concept MapProperty {
```
dselman, line 146: Let's not call this a "property" if it doesn't extend Property. This is a different animal...

packages/concerto-core/lib/introspect/field.js
```diff
@@ -62,24 +62,32 @@ class Field extends Property {
...
+                break;
+            case 'String':
+                if (this.ast.validator) {
+                    this.validator = new StringValidator(
+                        this,
+                        this.ast.validator
+                    );
+                }
+                break;
+            case 'Map':
```
dselman, line 83: Careful .. this looks like 'Map' is a new type of primitive - probably not want we want? There is lots of code that switches over the primitive types.

dselman (changes requested):
I think there's some conceptual confusion -- e.g are Maps primitives? What do we call the key/value definitions for a Map? What are the meta-properties on key and value?

packages/concerto-metamodel/lib/metamodel.cto
Jonathan-Casey: This is now refactored so that `MapKey` and `MapValue` inherit from `Declaration`

```
concept MapDeclaration extends Declaration {
  o MapKey key
  o MapValue value
}

concept MapKey extends Declaration {
  o Property property
  o TypeIdentifier type
}

concept MapValue extends Declaration {
  o Property property
  o TypeIdentifier type
 }
```

Which might be further refactored to  down to a type of

```
concept MapProperty extends Declaration  {
  o Property property
  o TypeIdentifier type
}

concept MapKey extends MapProperty { }

concept MapValue extends MapProperty { }
```

or perhaps just

```
concept MapProperty extends Declaration  {
  o Property property
  o TypeIdentifier type
}

concept MapDeclaration extends Declaration {
  o MapProperty key
  o MapProperty value
}
```


Conceptually do you feel like this is headed in the right direction, or am I headed into the weeds here? Based on your above comment unsure if you have something else in mind entirely eg: "Lets not call this a property" - but to my mind a Key and Value are both properties of a Map Declaration. Then again, `MapProperty extends Declaration` seems incorrect (Property extending Declaration?) - this could be better named `MapPropertyDeclaration`, though this sounds a little overloaded.

PR #614, opened 2023-02-28
feat(tools): add graph and graph search

mttrbrts:
Adds utilities to allow navigation of the dependency tree of a model as a graph. This allows for more advanced operations like "tree-shaking" (see `findConnectedGraph`), and finding unused imports (not implemented in this PR).

```js
const visitor = new ConcertoGraphVisitor();
const graph = new DirectedGraph();
modelManager.accept(visitor, { graph });

// Find the maximal subgraph starting from 'ChangeOfAddress'
const connectedGraph = graph.findConnectedGraph('org.acme.hr@1.0.0.ChangeOfAddress');

const filteredModelManager = modelManager
  .filter(declaration => 
    connectedGraph.hasVertex(declaration.getFullyQualifiedName())
  );
```

### Changes

- Adds `filter` functions to `BaseModelManager` and `ModelFile` to allow filtering declarations based on the graph.
- Adds `ConcertoGraphVisitor` and `DirectedGraph` to `concerto-tools`, and tests
- Adds new `Declaration` type to `concerto-core`. This is a new Class between `ClassDeclaration`/`ScalarDeclaration` and `Decorated`. This allows us to refactor common functions in the base class. This better matches the metamodel, will allow for better docs, better usage in TypeScript clients, and better extensibility to aggregate types in the future. We use this new type in the definition of the filter function.
- Also fixes and extends TypeScript types.

### Flags
- While refactoring `ClassDeclaration`/`ScalarDeclaration`, I noticed that `ScalarDeclaration` has several functions that should only appear on `ClassDeclaration`. I've marked them as `@deprecated`.

### Author Checklist


packages/concerto-core/test/introspect/declaration.js
```diff
@@ -0,0 +1,73 @@
...
+'use strict';
+
+const Declaration = require('../../lib/introspect/declaration');
+const ModelFile = require('../../lib/introspect/modelfile');
+
+require('chai').should();
+const should = require('chai').should();
+const sinon = require('sinon');
+
+describe('Decorated', () => {
```
dselman, line 24: Declaration?

packages/concerto-core/test/introspect/scalardeclaration.js
```diff
@@ -102,11 +102,9 @@ describe('ScalarDeclaration', () => {
             const modelFiles = introspectUtils.loadModelFiles([modelFileName], modelManager);
             modelManager.addModelFiles(modelFiles);
         });
-        it('should throw an error', () => {
+        it('should return null', () => {
             const testClass = modelManager.getType('com.testing.SSN');
-            (() => {
-                testClass.getSuperType();
-            }).should.throw('Scalars do not have super types.');
+            should.equal(testClass.getSuperType(), null);
```
dselman, line 107: Why return null here?

dselman (changes requested):
Awesome.

packages/concerto-core/test/introspect/scalardeclaration.js
mttrbrts: Good question. I'm adding a test to increase coverage which reflects the current implementation.

 This is one of the now deprecated functions. It should go away in the next major release, IMO

dselman (approved):
Nice work!

> Note that markdown-transform uses the https://www.npmjs.com/package/dijkstrajs library. Perhaps we can refactor this (or that)?

https://github.com/accordproject/markdown-transform/blob/a6c33ba41969f7a6241a68c1aaf6c066d116a900/packages/markdown-transform/package.json#L98

PR #632, opened 2023-03-24
fix(core): disallow reserved fields in JSON serialization

mttrbrts:
Better aligns the introspection API with the parser for fields that begin with a `$`.

Also, note that the existing behaviour is to add an additional field `$identifier` which shadows the value of any user-provided identifier field. This is by design because it allows client applications to resolve the identifier value without introspecting the model (which may not be available client-side). 

For example,
```cs
namespace test@1.0.0

concept test identified by name
{      o String name    }
```

The following payload is considered valid
```json
{    "$class": "test@1.0.0.test",     "name": "test",    "$identifier" : "test" }
```

### Changes

- When a concept is identified by an explicitly named field (e.g. `concept p identified by x {...}`), we give the value of that field primacy.
- We explicitly list reserved fields, rather than assuming that anything that starts with `$` is a system field.

### Flags

- This changes the behaviour of the operation `.getIdentifier()` when the user has explicitly set `$identifier` but hasn't changed the value of the explicitly named identifier field. See the test changes for `ValidatedResource`
- The list of private reserved keywords now includes `$superTypes`, and `$imports`. These are reserved for future use, where we may want to resolve or cache these values in serializations to improve performance.

### Author Checklist

### DIFF
--- a/packages/concerto-core/test/models/model.js
+++ b/packages/concerto-core/test/models/model.js
@@ -83,75 +83,75 @@ describe('Model Tests', function(){
             // o String stringProperty
             resource.setPropertyValue('stringProperty', 'string');
             resource.stringProperty.should.equal('string');
-            assert.throws( function() {resource.setPropertyValue('stringProperty', 1);}, /ValidationException: Model violation in the "org.acme.base.BaseAsset#123" instance. The field "stringProperty" has a value of "1" \(type of value: "number"\). Expected type of value: "String"./);
+            assert.throws( function() {resource.setPropertyValue('stringProperty', 1);}, /ValidationException: Model violation in the "org.acme.base.BaseAsset#string" instance. The field "stringProperty" has a value of "1" \(type of value: "number"\). Expected type of value: "String"./);
 
             // o Integer integerProperty
             resource.setPropertyValue('integerProperty', 999);
             resource.integerProperty.should.equal(999);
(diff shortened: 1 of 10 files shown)


packages/concerto-core/test/models/model.js
```diff
@@ -83,75 +83,75 @@ describe('Model Tests', function(){
             // o String stringProperty
             resource.setPropertyValue('stringProperty', 'string');
             resource.stringProperty.should.equal('string');
-            assert.throws( function() {resource.setPropertyValue('stringProperty', 1);}, /ValidationException: Model violation in the "org.acme.base.BaseAsset#123" instance. The field "stringProperty" has a value of "1" \(type of value: "number"\). Expected type of value: "String"./);
+            assert.throws( function() {resource.setPropertyValue('stringProperty', 1);}, /ValidationException: Model violation in the "org.acme.base.BaseAsset#string" instance. The field "stringProperty" has a value of "1" \(type of value: "number"\). Expected type of value: "String"./);
```
DianaLease, line 86: I don't really understand why the id in these tests changed from the "123" that was passed in when creating the instance to instead the string "string". Is this expected?

DianaLease (commented):
One question, but otherwise looks good!

packages/concerto-core/test/models/model.js
mttrbrts: It's because (rather confusingly) the definition of BaseAsset is
```
asset BaseAsset identified by stringProperty {
  o String stringProperty
  //...
``` 

So now, changing the value of that property, also has the effect of changing the identifier for the asset. This should have previously been considered a bug IMO!

DianaLease (approved):

PR #635, opened 2023-04-17
UUID format fix for jsonschema visitor

ragi-dayananda:
### Changes

JSON schema for CTO model generates the referential type to be "string" with appropriate referent type. However, it should also support the "identified by" type format for "uuid"

### Author Checklist


dselman (commented):
Do we also handle adding the UUID format to non relationship UUID scalars?

ragi-dayananda:
> Do we also handle adding the UUID format to non relationship UUID scalars?

@dselman no, we haven't handled that in this PR. Good catch, let me add it.

PR #637, opened 2023-04-21
(chore) remove Ergo References from Concerto (#610)

### LINKED ISSUE #610: Remove references to Ergo from Concerto
With the [Ergo language no longer maintained](https://github.com/accordproject/technical-steering-committee/issues/35#issuecomment-1262787490) we should clean up all of the code paths that are Ergo-specific in Concerto.

This is expected to be a breaking-change because it will require removing parameters from some public API operations. However, the impact is expected to be very limited, because the only usage of those parameters is within the Ergo engine.

We could consider retaining the ordering of parameters in API operations so as not to force a new major version ahead of time. 

Examples of redundant code can be found with the following query:
https://github.com/search?q=repo%3Aaccordproject%2Fconcerto+ergo+language%3ATypeScript+OR+language%3AJavaScript&type=code&l=TypeScript

e.g.
https://github.com/accordproject/concerto/blob/2f22485b57601dd82df38a0c987379d1ad993361/packages/concerto-core/lib/serializer/jsongenerator.js#L123-L133

The clean-up should also remove redundant tests, and CLI operations.

jamieshorten:
# Closes #610

### Changes
- remove Ergo References from Concerto 

### Author Checklist


jamieshorten:
Changed commit message format.

packages/concerto-core/lib/serializer/jsonpopulator.js
```diff
@@ -94,13 +94,11 @@ class JSONPopulator {
      * Constructor.
      * @param {boolean} [acceptResourcesForRelationships] Permit resources in the
      * place of relationships, false by default.
-     * @param {boolean} [ergo] target ergo.
      * @param {number} [utcOffset] - UTC Offset for DateTime values.
      * @param {number} [strictQualifiedDateTimes] - Only allow fully-qualified date-times with offsets.
      */
-    constructor(acceptResourcesForRelationships, ergo, utcOffset, strictQualifiedDateTimes) {
+    constructor(acceptResourcesForRelationships, utcOffset, strictQualifiedDateTimes) {
```
dselman, line 100: This is a breaking change for all upstream packages.

dselman (approved):
Looks good - though the breaking change to the JSONPopulator constructor signature makes me nervous for upstream users. Perhaps we should consider keeping the existing boolean in the signature and ignoring it?

packages/concerto-core/lib/serializer/jsonpopulator.js
mttrbrts: I'm not sure about that, it's marked as a private package in JSDoc and not exported AFAIK

mttrbrts: If we're nervous, we can add a dummy param until the next major version

dselman: Maybe I'm being too paranoid. I don't see any obvious uses of that in a code search?

jamieshorten: I'll go ahead and add a dummy parameter then. And note in the doc comment that it's a dummy parameter.

packages/concerto-core/lib/serializer/jsongenerator.js
```diff
@@ -42,15 +42,13 @@ class JSONGenerator {
      * written for subsequent instances, false by default.
      * @param {boolean} [convertResourcesToId] Convert resources that
      * are specified for relationship fields into their id, false by default.
-     * @param {boolean} [ergo] target ergo.
      * @param {number} [utcOffset] UTC Offset for DateTime values.
      */
-    constructor(convertResourcesToRelationships, permitResourcesForRelationships, deduplicateResources, convertResourcesToId, ergo, utcOffset) {
+    constructor(convertResourcesToRelationships, permitResourcesForRelationships, deduplicateResources, convertResourcesToId, utcOffset) {
```
mttrbrts, line 47: Looks like we also need a dummy param here?

jamieshorten: Oops! missed that one.

mttrbrts (approved):
Congrats on your first commit! 🎉

PR #646, opened 2023-05-16
feat(map): Add Introspection for Map Type

Jonathan-Casey:

This change adds the specified classes for runtime introspection of a MapType - `MapDeclaration`,`MapKeyType` and a general `AggregateValueType`. Each class is responsible for processing of an AST representation of a MapDeclaration before instantiation, exposing various functionality, as well as correct validation for each of its parts.

For more information on the design specification see [here](https://github.com/accordproject/concerto/wiki/Aggregate-Types-Design)

### Changes

- <ONE> Added classes for `MapDeclaration`,`MapKeyType` and `AggregateValueType`
- <TWO> Added test coverage
- <THREE> Added type definitions

### Related Issues
- Issue #447 

### Author Checklist
- [X] Ensure you provide a [DCO sign-off](https://github.com/probot/dco#how-it-works) for your commits using the `--signoff` option of git commit.
- [X] Vital features and changes captured in unit and/or integration tests
- [X] Commits messages follow [AP format](https://github.com/accordproject/techdocs/blob/master/DEVELOPERS.md#commit-message-format)

### DIFF
--- a/packages/concerto-core/lib/introspect/mapdeclaration.js
+++ b/packages/concerto-core/lib/introspect/mapdeclaration.js
@@ -1,0 +1,249 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const Decorated = require('./decorated');
+const IllegalModelException = require('./illegalmodelexception');
+const MapPropertyType = require('./mapvaluetype');
+const MapKeyType = require('./mapkeytype');
+const ModelUtil = require('../modelutil');
+
+// Types needed for TypeScript generation.
+/* eslint-disable no-unused-vars */
+/* istanbul ignore next */
+if (global === undefined) {
+    const ModelFile = require('./modelfile');
+}
+
+/**
+ * MapDeclaration defines an MapType of Key & Value pair.
+ *
+ * @extends Decorated
+ * @see See {@link Decorated}
+ * @class
+ * @memberof module:concerto-core
+ */
+class MapDeclaration extends Decorated {
+    /**
+     * Create an MapDeclaration.
+     * @param {ModelFile} modelFile the ModelFile for this class
+     * @param {Object} ast - The AST created by the parser
+     * @throws {IllegalModelException}
+     */
+    constructor(modelFile, ast) {
+        super(ast);
+        this.modelFile = modelFile;
+        this.process();
+    }
+
+    /**
+     * Process the AST and build the model
+     *
+     * @throws {IllegalModelException}
+     * @private
+     */
+    process() {
+        super.process();
+
+        if (this.ast.properties.length !== 2) {
+            throw new IllegalModelException(`MapDeclaration must contain exactly two properties -  MapKeyType & MapyValueType ${this.ast.name}`, this.modelFile, this.ast.location);
+        }
+
+        const k = this.ast.properties.find(p => p.$class === 'concerto.metamodel@1.0.0.MapKeyType');
+        const v = this.ast.properties.find(p => p.$class === 'concerto.metamodel@1.0.0.AggregateValueType' || p.$class === 'concerto.metamodel@1.0.0.AggregateRelationshipValueType');
+
+        if (!k) {
+            throw new IllegalModelException(`MapDeclaration must contain MapKeyType  ${this.ast.name}`, this.modelFile, this.ast.location);
+        }
+
+        if (!v) {
+            throw new IllegalModelException(`MapDeclaration must contain AggregateValueType  ${this.ast.name}`, this.modelFile, this.ast.location);
+        }
+
+        this.name = this.ast.name;
+        this.key = new MapKeyType(this, k);
+        this.value = new MapPropertyType(this, v);
+        this.fqn = ModelUtil.getFullyQualifiedName(this.modelFile.getNamespace(), this.ast.name);
+    }
+
+    /**
+     * Semantic validation of the structure of this class.
+     *
+     * @throws {IllegalModelException}
+     * @protected
+     */
+    validate() {
+        super.validate();
+        this.key.validate();
+        this.value.validate();
+    }
+
+    /**
+     * Returns the fully qualified name of this class.
+     * The name will include the namespace if present.
+     *
+     * @return {string} the fully-qualified name of this class
+     */
+    getFullyQualifiedName() {
+        return this.fqn;
+    }
+
+    /**
+     * Returns the ModelFile that defines this class.
+     *
+     * @public
+     * @return {ModelFile} the owning ModelFile
+     */
+    getModelFile() {
+        return this.modelFile;
+    }
+
+    /**
+     * Returns the short name of a class. This name does not include the
+     * namespace from the owning ModelFile.
+     *
+     * @return {string} the short name of this class
+     */
+    getName() {
+        return this.name;
+    }
+
+    /**
+     * Returns the type of the Map key property.
+     *
+     * @return {string} the short name of this class
+     */
+    getKey() {
+        return this.key;
+    }
+
+    /**
+     * Returns the type of the Map Value property.
+     *
+     * @return {string} the short name of this class
+     */
+    getValue() {
+        return this.value;
+    }
+
+    /**
+     * Returns the MapDeclaration properties
+     *
+     * @return {array} the short name of this class
+     */
+    getProperties() {
+        return this.ast.properties;
+    }
+
+    /**
+     * Returns the string representation of this class
+     * @return {String} the string representation of the class
+     */
+    toString() {
+        return 'MapDeclaration {id=' + this.getFullyQualifiedName() + '}';
+    }
+
+    /**
+     * Returns the kind of declaration
+     *
+     * @return {string} what kind of declaration this is
+     */
+    declarationKind() {
+        return 'MapDeclaration';
+    }
+
+    /**
+     * Returns true if this class is abstract.
+     *
+     * @return {boolean} true if the class is abstract
+     */
+    isAbstract() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of an asset.
+     *
+     * @return {boolean} true if the class is an asset
+     */
+    isAsset() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of a participant.
+     *
+     * @return {boolean} true if the class is a participant
+     */
+    isParticipant() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of a transaction.
+     *
+     * @return {boolean} true if the class is a transaction
+     */
+    isTransaction() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of an event.
+     *
+     * @return {boolean} true if the class is an event
+     */
+    isEvent() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of a concept.
+     *
+     * @return {boolean} true if the class is a concept
+     */
+    isConcept() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of an enum.
+     *
+     * @return {boolean} true if the class is an enum
+     */
+    isEnum() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of a scalar declaration.
+     *
+     * @return {boolean} true if the class is a scalar
+     */
+    isScalarDeclaration() {
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of a class declaration.
+     *
+     * @return {boolean} true if the class is a class
+     */
+    isMapDeclaration() {
+        return true;
+    }
+}
+
+module.exports = MapDeclaration;
--- a/packages/concerto-core/lib/introspect/mapkeytype.js
+++ b/packages/concerto-core/lib/introspect/mapkeytype.js
@@ -1,0 +1,152 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const { MetaModelNamespace } = require('@accordproject/concerto-metamodel');
+
+const Decorated = require('./decorated');
+const IllegalModelException = require('./illegalmodelexception');
+const ModelUtil = require('../modelutil');
+
+// Types needed for TypeScript generation.
+/* eslint-disable no-unused-vars */
+/* istanbul ignore next */
+if (global === undefined) {
+    const ModelFile = require('./modelfile');
+}
+
+/**
+ * MapKeyType defines a Key type of an MapDeclaration.
+ *
+ * @extends Decorated
+ * @see See {@link Decorated}
+ * @class
+ * @memberof module:concerto-core
+ */
+class MapKeyType extends Decorated {
+    /**
+     * Create an MapKeyType.
+     * @param {MapDeclaration} parent - The owner of this property
+     * @param {Object} ast - The AST created by the parser
+     * @throws {IllegalModelException}
+     */
+    constructor(parent, ast) {
+        super(ast);
+        this.parent = parent;
+        this.name = null;
+        this.type = null;
+        this.fqn = null;
+        this.process();
+    }
+
+    /**
+     * Process the AST and build the model
+     *
+     * @throws {IllegalModelException}
+     * @private
+     */
+    process() {
+        super.process();
+        this.name = this.ast.name;
+        this.type = this.ast.name;
+        this.fqn = ModelUtil.getFullyQualifiedName(this.parent.getModelFile().getNamespace(), this.ast.name);
+    }
+
+    /**
+     * Semantic validation of the structure of this class.
+     *
+     * @throws {IllegalModelException}
+     * @protected
+     */
+    validate() {
+        const declaration = this.getModelFile().getAllDeclarations();
+        const key = declaration.find(decl => decl.name === this.name);
+
+        if (!key?.isConcept?.()           &&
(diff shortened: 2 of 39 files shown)


packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -0,0 +1,249 @@
...
+ */
+class MapDeclaration extends Decorated {
+    /**
+     * Create an MapDeclaration.
+     * @param {ModelFile} modelFile the ModelFile for this class
+     * @param {Object} ast - The AST created by the parser
+     * @throws {IllegalModelException}
+     */
+    constructor(modelFile, ast) {
+        super(ast);
```
mttrbrts, line 46: As discussed, let's hide the behind a feature flag and throw a warning/exception until we're stable and have sufficient coverage.

packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -0,0 +1,249 @@
...
+        const k = this.ast.properties.find(p => p.$class === 'concerto.metamodel@1.0.0.MapKeyType');
+        const v = this.ast.properties.find(p => p.$class === 'concerto.metamodel@1.0.0.AggregateValueType' || p.$class === 'concerto.metamodel@1.0.0.AggregateRelationshipValueType');
+
+        if (!k) {
+            throw new IllegalModelException(`MapDeclaration must contain MapKeyType  ${this.ast.name}`, this.modelFile, this.ast.location);
+        }
+
+        if (!v) {
+            throw new IllegalModelException(`MapDeclaration must contain AggregateValueType  ${this.ast.name}`, this.modelFile, this.ast.location);
+        }
```
mttrbrts, line 73: Nit, please use descriptive names for variables. `key`, `value`, `property` etc.

packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * MapDeclaration defines an MapType of Key & Value pair.
+ *
+ * @extends Decorated
+ * @see See {@link Decorated}
+ * @class
+ * @memberof module:concerto-core
+ */
+class MapDeclaration extends Decorated {
```
mttrbrts, line 38: Should this extend `Declaration` instead? If so, it means that we can remove many of the boiler-plate function definitions below which are duplicated in that Class.

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -0,0 +1,152 @@
...
+     * Process the AST and build the model
+     *
+     * @throws {IllegalModelException}
+     * @private
+     */
+    process() {
+        super.process();
+        this.name = this.ast.name;
+        this.type = this.ast.name;
+        this.fqn = ModelUtil.getFullyQualifiedName(this.parent.getModelFile().getNamespace(), this.ast.name);
```
mttrbrts, line 64: I'm not convinced that this is right.

Do we ever want to reference a key or value outside of the context of a MapDeclaration?

If so, the fqn should have the prefix of the parent MapDeclaration's fqn, not just it's namespace.
If not, a fqn doesn't make much sense for keys/values

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -0,0 +1,152 @@
...
+            throw new IllegalModelException(`MapKeyType has invalid Type: ${this.type}`);
+        }
+
+        if (key?.isConcept?.()) {
+            if(!key.isIdentified()) {
+                throw new IllegalModelException(
+                    `ConceptDeclaration must be identified in context of MapKeyType: ${this.type}`
+                );
+            }
+        }
```
mttrbrts, line 90: Nit, unnecessary nested conditions?

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -0,0 +1,152 @@
...
+        if (key?.isScalarDeclaration?.()) {
+            if (!(key?.ast.$class === `${MetaModelNamespace}.StringScalar`)  &&
+                !(key?.ast.$class === `${MetaModelNamespace}.BooleanScalar`) &&
+                !(key?.ast.$class === `${MetaModelNamespace}.DateTimeScalar` )) {
+                throw new IllegalModelException(
+                    `Scalar must be one of StringScalar, BooleanScalar, DateTimeScalar in context of MapKeyType. Invalid Scalar: ${this.name}`
+                );
+            }
+        }
+    }
```
mttrbrts, line 101: Nit, unnecessary nested conditions?

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,138 @@
...
+     * Process the AST and build the model
+     *
+     * @throws {IllegalModelException}
+     * @private
+     */
+    process() {
+        super.process();
+        this.name = this.ast.name;
+        this.type = this.ast.name;
+        this.fqn = ModelUtil.getFullyQualifiedName(this.parent.getModelFile().getNamespace(), this.ast.name);
```
mttrbrts, line 62: As per my comment on KeyTypes

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,138 @@
...
+     * @protected
+     */
+    validate() {
+        const declarations = this.getModelFile().getAllDeclarations();
+
+        const value = declarations.find(decl => decl.name === this.name);
+
+        if (!value?.isConcept?.()           &&
+            !value?.isEnum?.()              &&
+            !value?.isEvent?.()             &&
```
mttrbrts, line 78: Why do we need to call out Event specifically?

If we do, we should check for Transactions, Assets and Participants too

packages/concerto-core/test/data/parser/mapdeclaration/mapdeclaration.goodkey.declaration.event.cto
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+namespace com.acme@1.0.0
+
+event Activity {}
+
+map Dictionary {
+  o Activity
+  o String
+}
```
mttrbrts, line 22: This is an interesting idea, not necessarily wrong. But not in the design.

Presumably, we would key-off of the timestamp of the event?

mttrbrts (changes requested):
Love all of the test cases.

Mostly minor comments, happy to chat about the best inheritance hierarchy for MapDeclarations and if FQNs make sense for Keys/Values.

packages/concerto-core/lib/introspect/mapvaluetype.js
Jonathan-Casey: `Event` is a legal map value type , `Transactions`, `Assets` and `Participants` are not specified as value types.

packages/concerto-core/test/data/parser/mapdeclaration/mapdeclaration.goodkey.declaration.event.cto
Jonathan-Casey: Yes this should not have been committed, and it is not referenced from any test case.

packages/concerto-core/lib/introspect/mapkeytype.js
Jonathan-Casey: guilty..

packages/concerto-core/lib/introspect/mapvaluetype.js
mttrbrts: Ok, I think that they should. However, Event, Transactions, Assets and Participants are all Concepts IIRC. so this line is redundant?

Jonathan-Casey: The evaluation expression is based on the `$class` value, which is different for both ConceptDeclaration and EventDeclaration:

`return this.type === ${MetaModelNamespace}.ConceptDeclaration
`

`return this.type === ${MetaModelNamespace}.EventDeclaration
`

dselman:
Please ensure that the latest metamodel is publish to the public model repo: https://models.accordproject.org/concerto/metamodel@1.0.0.html

dselman:
Please ensure that the latest metamodel is publish to the public model repo: https://models.accordproject.org/concerto/metamodel@1.0.0.html

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Returns the name of the MapKey. This name does not include the
+     * namespace from the owning ModelFile.
+     *
+     * @return {string} the short name of this class
+     */
+    getName() {
+        return this.ast.name;
+    }
```
mttrbrts, line 115: The AST doesn't have a property called `name` AFAICT?

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Returns the Type of the MapKey. This name does not include the
+     * namespace from the owning ModelFile.
+     *
+     * @return {string} the short name of this class
+     */
+    getType() {
+        return this.ast.name;
```
mttrbrts, line 124: should this be `this.ast.type`?

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Returns the name of the MapValue. This name does not include the
+     * namespace from the owning ModelFile.
+     *
+     * @return {string} the short name of this class
+     */
+    getName() {
+        return this.ast.name;
```
mttrbrts, line 104: Unused?

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Returns the Type of the MapValue. This name does not include the
+     * namespace from the owning ModelFile.
+     *
+     * @return {string} the short name of this class
+     */
+    getType() {
+        return this.ast.name;
```
mttrbrts, line 114: Should be `this.ast.type`?

packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -0,0 +1,182 @@
...
+    getName() {
+        return this.name;
+    }
+
+    /**
+     * Returns the type of the Map key property.
+     *
+     * @return {string} the short name of this class
+     */
+    getKey() {
```
dselman, line 133: getKeyType would be more consistent with Property

packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -0,0 +1,182 @@
...
+        return this.key;
+    }
+
+    /**
+     * Returns the type of the Map Value property.
+     *
+     * @return {string} the short name of this class
+     */
+    getValue() {
```
dselman, line 142: getValueType

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * Semantic validation of the structure of this class.
+     *
+     * @throws {IllegalModelException}
+     * @protected
+     */
+    validate() {
+        const declaration = this.getModelFile().getAllDeclarations();
```
dselman, line 70: I was confused by this. The goal is to check the type of the key... 

I think I'd try this as a switch/case to make it clearer. Something like:

switch(this.getKeyType()) {
   case 'String':
   case 'Boolean':
   case 'DateTime':
    // valid!
   break;
  default:
 {
    // resolveType using model manager (throws exception if type is missing)
   // if type is scalar and not in [string, boolean, date time] throw error
   // if type is concept and not identified throw error
  }
}

This looks like it doesn't handle the case of key being null. You should use `BaseModelManager.resolveType` to lookup complex types so that if the lookup fails we get a consistent error message.

packages/concerto-core/lib/introspect/mapkeytype.js
Jonathan-Casey: You are correct. Its a little ambiguous, the AST in this case is a subset representing the MapDeclaration only. There is no need to process the entire AST, the constructor only uses what it needs from the parse tree.

```
{
  '$class': 'concerto.metamodel@1.0.0.MapDeclaration',
  name: 'Dictionary',
  properties: [
    {
      '$class': 'concerto.metamodel@1.0.0.MapKeyType',
      name: 'Person',
      location: [Object]
    },
    {
      '$class': 'concerto.metamodel@1.0.0.AggregateValueType',
      name: 'String',
      location: [Object]
    }
  ],
  location: {
    '$class': 'concerto.metamodel@1.0.0.Range',
    start: {
      offset: 676,
      line: 19,
      column: 1,
      '$class': 'concerto.metamodel@1.0.0.Position'
    },
    end: {
      offset: 716,
      line: 22,
      column: 2,
      '$class': 'concerto.metamodel@1.0.0.Position'
    }
  }
} 
```

packages/concerto-core/lib/introspect/mapkeytype.js
Jonathan-Casey: see above

packages/concerto-core/lib/introspect/mapvaluetype.js
Jonathan-Casey: see above. This is set to return a global variable now, but is assigned from the same property

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * @throws {IllegalModelException}
+     * @protected
+     */
+    validate() {
+        const declarations = this.getModelFile().getAllDeclarations();
+
+        const value = declarations.find(decl => decl.name === this.type);
+
+        if (!value?.isConcept?.()           &&
```
dselman: Same comment as the other validate call. I'd rewrite this with a positive switch/case rather than all these negative tests.

packages/concerto-core/test/data/parser/mapdeclaration/mapdeclaration.goodkey.scalar.boolean.cto
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace com.acme@1.0.0
+
+scalar BOOL extends Boolean
+
+map Dictionary {
+  o BOOL
```
dselman, line 20: This one is interesting to me... It implies we are stringifying the boolean values "true" and "false" to use them as map keys - which sort of begs the question why not support stringifying numbers? We seem a bit inconsistent here.

packages/concerto-core/types/lib/introspect/classdeclaration.d.ts
```diff
@@ -149,6 +149,12 @@ declare class ClassDeclaration extends Declaration {
      * @return {boolean} true if the class is an asset
      */
     isConcept(): boolean;
+    /**
+     * Returns true if this class is the definition of a enum.
+     *
+     * @return {boolean} true if the class is an asset
```
dselman, line 155: Typo

packages/concerto-core/types/lib/introspect/mapdeclaration.d.ts
```diff
@@ -0,0 +1,96 @@
+export = MapDeclaration;
+/**
+ * MapDeclaration defines an MapType of Key & Value pair.
```
dselman, line 3: Improve description. E.g. 

A MapDeclaration defines a map/dictionary of values indexed by a key.

dselman (changes requested):
Looking good - a few nits to pick!

packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Process the AST and build the model
+     *
+     * @throws {IllegalModelException}
+     * @private
+     */
+    process() {
+        super.process();
+
+        if (this.ast.properties.length !== 2) {
```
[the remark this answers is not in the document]
Jonathan-Casey: Can you provide some more detail on what you mean here? They _are_ two AST properties.

packages/concerto-core/lib/introspect/mapdeclaration.js
Jonathan-Casey: We would then be in the territory of defining an api consisting of calls like `clz.getValueType().getType()` which is unclear imo.

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,127 @@
...
+     * Create an MapValueType.
+     * @param {MapDeclaration} parent - The owner of this property
+     * @param {Object} ast - The AST created by the parser
+     * @param {ModelFile} modelFile - the ModelFile for the Map class
+     * @throws {IllegalModelException}
+     */
+    constructor(parent, ast) {
+        super(ast);
+        this.parent = parent;
+        this.type = null;
```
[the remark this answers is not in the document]
Jonathan-Casey: this is populated below in the process() call. I don't think it matters too much, but for readability i've bumped the assignment to the constructor.

packages/concerto-core/lib/introspect/mapkeytype.js
Jonathan-Casey: In the case of a key or value, they can never be null, otherwise the MapType cannot be constructed.

packages/concerto-core/lib/introspect/mapvaluetype.js
Jonathan-Casey: Because all Concerto concepts and primitives are legal MapValueTypes I've since changed this to return true without any checking, but on consideration of your comment  I'm not so sure this is actually correct.

For example, whats to stop a bad AST being fed to the the constructor:

```
{
    '$class': 'concerto.metamodel@1.0.0.MapKeyType',
    name: 'DateTime'
},
{
    '$class': 'concerto.metamodel@1.0.0.AggregateValueType',
    name: 'String3'
}
```

`String3` is not a valid concerto type, but would still pass the current validation step. So, validation is mandatory and should be explicit so as to catch issues early on in the case where a malformed AST is in use. Will add additional validation logic based on your suggestion.

packages/concerto-core/test/data/parser/mapdeclaration/mapdeclaration.goodkey.scalar.boolean.cto
Jonathan-Casey: Based on our discussion, I've removed `Boolean` type from the legal key values.

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -0,0 +1,125 @@
...
+            !value?.isEvent?.()             &&
+            !value?.isParticipant?.()       &&
+            !value?.isTransaction?.()       &&
+            !value?.isMapDeclaration?.()    &&
+            !value?.isScalarDeclaration?.() &&
+            !['String', 'Long', 'Integer', 'Double', 'Boolean', 'DateTime'].includes(this.type)) {
+
+            throw new IllegalModelException(
+                `MapPropertyType has invalid Type: ${this.type}`
+            );
```
mttrbrts, line 82: This line still needs a test case.

mttrbrts (approved):

Issue #643, opened 2023-05-08
Support min and max length for a string property in CTO


ragi-dayananda:
# Feature Request 🛍️

Currently, we have the option to use regex to enforce string length, but this can be an expensive operation. Considering that major data modeling tools support string length validators, it would be beneficial to have this functionality in Concerto as well.

### Solution 1: Introduce a new decorator to support this feature.
- This could an easy option in terms of implementation.
- Example -
```
concept Contact {
  @length(0,100)
   o String firstName
} 
```
### Solution 2: Enhance CTO property definition to incorporate this feature.

- I noticed the existence of the meta property `range[1,10]` to declare a valid range for numeric properties. We can leverage this meta property for strings as well, where it would represent the length of a string. 
- Example -
```
concept Contact {
   o String firstName range[0,100]
} 
```
- If using the range meta property is not suitable for string properties, we can introduce a new meta property, perhaps called `length` to support this feature.

## Context

We are in the process of replacing our data models with Concerto models. Some of the string properties in our data model have string length configurations/validators, and it would be great to have a similar setting in Concerto.

dselman:
We should only use decorators for things that are external/opaque to Concerto. In this case we'd be using these facets to validate string properties so I think the meta model should be extended to support this. 'length' seems like a good name, and we could implement it similar to range for numbers, with an optional lower and upper bound, or we could add 'maxLength' and 'minLength' as explicit meta properties.

mttrbrts:
The [`range` property modifier syntax](https://concerto.accordproject.org/docs/design/specification/model-properties#meta-properties) includes an `=` so we should mirror that. e.g.

```
concept Contact {
   o String firstName length=[0,100]
}
```  

We could also make lower and upper bounds optional (like with `range`), e.g.

```
concept Contact {
   o String firstName length=[1,] // only non-empty strings are allowed
   o String lastName length=[,100] // syntactic sugar for [0,100]
}
```  

Also note that with `range`, the upper-bound is inclusive, i.e. `range=[3,10]` means `3 >= x <= 10`
https://github.com/accordproject/concerto/blob/9009e82817d833fcc9172bf5313b5529c6765250/packages/concerto-core/lib/introspect/numbervalidator.js#L92-L102

mttrbrts:
As discussed on today's working group call, we should give this modifier precedence over the regex modifier if both are used.

Validating against `length` is the simpler, faster, and less-general operation.

f5th-dimensional:
[...]

PR #647, opened 2023-05-17
feat(*): Support string length validation

ragi-dayananda:
# Closes 

WIP

### Related Issues
- Issue #[643](https://github.com/accordproject/concerto/issues/643)
- Pull Request #[concerto-metamodel/pull/3](https://github.com/accordproject/concerto-metamodel/pull/3)

### Author Checklist


dselman:
Looks great so far @ragi-dayananda! Once the introspection class is in place for validation etc. it would be great to update the JSON Schema code generator to use this as an example / test of it being applied.

dselman:
Docs page to be updated is here: https://concerto.accordproject.org/docs/design/specification/model-properties

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -37,17 +38,45 @@ class StringValidator extends Validator{
...
+                this.reportError(null, 'Invalid string length, minLength and-or maxLength must be specified.');
+            } else if (this.minLength < 0 || this.maxLength < 0) {
+                this.reportError(null, 'minLength and-or maxLength must be positive integers.');
+            } else if (this.minLength === null || this.maxLength === null) {
+                // this is fine and means that we don't need to check whether minLength > maxLength
+            } else {
+                if(this.minLength > this.maxLength) {
+                    this.reportError(null, 'minLength must be less than or equal to maxLength.');
+                }
+            }
```
mttrbrts, line 69:
```suggestion
            } else if(this.minLength > this.maxLength) {
                this.reportError(null, 'minLength must be less than or equal to maxLength.');
            }
```

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -37,17 +38,45 @@ class StringValidator extends Validator{
...
+        if (lengthValidator) {
+            if(Object.prototype.hasOwnProperty.call(lengthValidator, 'minLength')) {
+                this.minLength = lengthValidator.minLength;
+            }
+            if(Object.prototype.hasOwnProperty.call(lengthValidator, 'maxLength')) {
+                this.maxLength = lengthValidator.maxLength;
+            }
+            if(this.minLength === null && this.maxLength === null) {
+                // can't specify no upper and lower value
+                this.reportError(null, 'Invalid string length, minLength and-or maxLength must be specified.');
```
mttrbrts, line 60: Should these be `this.reportError(field.getName(), ...`?

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -60,14 +89,38 @@ class StringValidator extends Validator{
...
     validate(identifier, value) {
         if(value !== null) {
-            if(!this.regex.test(value)) {
-                this.reportError(identifier, 'Value \'' + value + '\' failed to match validation regex: ' + this.regex);
+            if (this.regex) {
+                if(!this.regex.test(value)) {
+                    this.reportError(identifier, 'Value \'' + value + '\' failed to match validation regex: ' + this.regex);
+                }
+            }
+            //Enforce string length rule after regex check
```
mttrbrts, line 97: This comment suggests that length checking should come first.
https://github.com/accordproject/concerto/issues/643#issuecomment-1542134421

packages/concerto-core/test/introspect/stringvalidator.js
```diff
@@ -36,22 +48,51 @@ describe('StringValidator', () => {
...
+                new StringValidator(mockField, null, NEGETIVE_LENGTH);
+            }).should.throw(/minLength and-or maxLength must be positive integers/);
+        });
     });
 
     describe('#validate', () => {
 
         it('should ignore a null string', () => {
-            let v = new StringValidator(mockField, { pattern: '^[A-z][A-z][0-9]{7}' });
+            let v = new StringValidator(mockField, { pattern: '^[A-z][A-z][0-9]{7}' }, VALID_MIN_LENGTH_AND_MAX_LENGTH_AST);
```
mttrbrts, line 89: Why do we need to change the existing tests? It's important that if we skip this parameter, then we should get the same behaviour.

packages/concerto-core/test/introspect/stringvalidator.js
```diff
@@ -36,22 +48,51 @@ describe('StringValidator', () => {
...
             v.getRegex().toString().should.equal('/^[A-z][A-z][0-9]{7}/');
             v.validate('id', null);
         });
 
         it('should validate a string', () => {
-            let v = new StringValidator(mockField, { pattern: '^[A-z][A-z][0-9]{7}' });
+            let v = new StringValidator(mockField, { pattern: '^[A-z][A-z][0-9]{7}' }, VALID_MIN_LENGTH_AND_MAX_LENGTH_AST);
```
mttrbrts, line 95: Why do we need to change the existing tests? It's important that if we skip this parameter, then we should get the same behaviour.

packages/concerto-core/test/introspect/stringvalidator.js
```diff
@@ -87,48 +128,120 @@ describe('StringValidator', () => {
...
+
     });
 
     describe('#validate with custom RegEx engine', () => {
         const options = {
             regExp: XRegExp
         };
         it('should ignore a null string', () => {
-            let v = new StringValidator(mockField, { pattern: '^[A-z][A-z][0-9]{7}' }, options);
+            let v = new StringValidator(mockField, { pattern: '^[A-z][A-z][0-9]{7}' }, VALID_MIN_LENGTH_AND_MAX_LENGTH_AST, options);
```
mttrbrts, line 170: Why do we need to change the existing tests? It's important that if we skip this parameter, then we should get the same behaviour.

packages/concerto-core/test/introspect/stringvalidator.js
```diff
@@ -87,48 +128,120 @@ describe('StringValidator', () => {
...
             v.getRegex().toString().should.equal('/^[A-z][A-z][0-9]{7}/');
             v.validate('id', null);
         });
 
         it('should validate a string', () => {
-            let v = new StringValidator(mockField, { pattern: '^[\\p{Letter}\\p{Number}]{7}', flags: 'u' }, options);
+            let v = new StringValidator(mockField, { pattern: '^[\\p{Letter}\\p{Number}]{7}', flags: 'u' }, VALID_MIN_LENGTH_AND_MAX_LENGTH_AST, options);
```
mttrbrts, line 176: Why do we need to change the existing tests? It's important that if we skip this parameter, then we should get the same behaviour.

mttrbrts (changes requested):
Good progress, Ragi. I've added some minor comments in-line.

Can you update the issue to track the remaining work, please?

For example:

  - [x] validate           validate JSON against model files
  - [ ]  compile            generate code for a target platform
    - [ ] TypeScript
    - [ ] JSON Schema
    - etc...  
  - [x] parse              parse a cto string to a JSON syntax tree
  - [x] print              print a JSON syntax tree to a cto string
  - [ ] compare            compare two Concerto model files
  - [ ] infer              generate a concerto model from a source schema
  - [ ] generate <mode>    generate a sample JSON object for a concept
  
 We don't necessarily need complete coverage before we release this work, however, we do need to make sure that we fail gracefully / warn, or hide this work behind an environment variable until it's ready.

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -60,14 +89,38 @@ class StringValidator extends Validator{
      */
     validate(identifier, value) {
         if(value !== null) {
-            if(!this.regex.test(value)) {
-                this.reportError(identifier, 'Value \'' + value + '\' failed to match validation regex: ' + this.regex);
+            if (this.regex) {
```
Jonathan-Casey, line 92: I think its better to avoid unnecessary nesting of "if" statements here. You could always merge this to one conditional check which would improve code readability

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -60,14 +89,38 @@ class StringValidator extends Validator{
...
+                if(!this.regex.test(value)) {
+                    this.reportError(identifier, 'Value \'' + value + '\' failed to match validation regex: ' + this.regex);
```
Jonathan-Casey, line 94: why not use string interpolation for this reportError?

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -37,17 +38,45 @@ class StringValidator extends Validator{
...
+            if(Object.prototype.hasOwnProperty.call(lengthValidator, 'maxLength')) {
+                this.maxLength = lengthValidator.maxLength;
+            }
+            if(this.minLength === null && this.maxLength === null) {
+                // can't specify no upper and lower value
+                this.reportError(null, 'Invalid string length, minLength and-or maxLength must be specified.');
+            } else if (this.minLength < 0 || this.maxLength < 0) {
+                this.reportError(null, 'minLength and-or maxLength must be positive integers.');
+            } else if (this.minLength === null || this.maxLength === null) {
+                // this is fine and means that we don't need to check whether minLength > maxLength
```
Jonathan-Casey, line 64: Is this necessary if there is no code executed on the condition evaluating to true?

packages/concerto-core/lib/introspect/stringvalidator.js
ragi-dayananda: that's an existing code.

packages/concerto-core/lib/introspect/stringvalidator.js
ragi-dayananda: I copied most of the logic from number validator. I overlooked it may be. Let me remove it

packages/concerto-core/lib/introspect/stringvalidator.js
ragi-dayananda: I can update the logic, but I don't think order of execution is gonna solve the problem if regex has length constraint in it

mttrbrts: You're right, it won't. We just have the design principle to apply the least complex validation first.

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -37,17 +38,45 @@ class StringValidator extends Validator{
...
+        this.minLength = null;
+        this.maxLength = null;
+        this.regex = null;
+
+        if (lengthValidator) {
+            if(Object.prototype.hasOwnProperty.call(lengthValidator, 'minLength')) {
+                this.minLength = lengthValidator.minLength;
+            }
+            if(Object.prototype.hasOwnProperty.call(lengthValidator, 'maxLength')) {
+                this.maxLength = lengthValidator.maxLength;
```
dselman, line 56: this.minLength = lengthValidator?.minLength;
this.maxLength = lengthValidator?.maxLength;

dselman (commented):
Good work.

packages/concerto-core/test/introspect/stringvalidator.js
ragi-dayananda: I can't skip that param, for now passing null instead of valid length ast. 
Since here we are passing `options` as one more param, validator will try to bind it to length validator if we don't explicitly pass null/valid length validator

mttrbrts:
@ragi-dayananda There are a few edge cases missing from your tests. Can you bump the coverage, please?
https://coveralls.io/jobs/121606352/source_files/8583783901

mttrbrts (approved):

PR #654, opened 2023-05-25
feat(map): add serialisation for map<string, string>

Jonathan-Casey:
This change adds basic serialisation of a map in the form of <string, string>. Further type variations will follow in subsequent PR.

For more information on the design specification see [here](https://github.com/accordproject/concerto/wiki/Aggregate-Types-Design)

### Related Issues
- Issue #447 

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/modelutil.js
+++ b/packages/concerto-core/lib/modelutil.js
@@ -191,7 +191,7 @@ class ModelUtil {
     }
 
     /**
-     * Returns the true if the given field is an enumerated type
+     * Returns true if the given field is an enumerated type
      * @param {Field} field - the string
      * @return {boolean} true if the field is declared as an enumeration
      * @private
@@ -203,7 +203,7 @@ class ModelUtil {
     }
 
     /**
-     * Returns the true if the given field is an map type
+     * Returns true if the given field is an map type
      * @param {Field} field - the string
      * @return {boolean} true if the field is declared as an map
      * @private
@@ -215,7 +215,7 @@ class ModelUtil {
     }
 
     /**
-     * Returns the true if the given field is a Scalar type
+     * Returns true if the given field is a Scalar type
      * @param {Field} field - the Field to test
      * @return {boolean} true if the field is declared as an scalar
      * @private
--- a/packages/concerto-core/lib/modelutil.js
+++ b/packages/concerto-core/lib/modelutil.js
@@ -199,7 +199,7 @@ class ModelUtil {
     static isEnum(field) {
         const modelFile = field.getParent().getModelFile();
         const typeDeclaration = modelFile.getType(field.getType());
-        return (typeDeclaration !== null && typeDeclaration.isEnum());
+        return typeDeclaration?.isEnum();
     }
 
     /**
@@ -211,7 +211,7 @@ class ModelUtil {
     static isMap(field) {
         const modelFile = field.getParent().getModelFile();
         const typeDeclaration = modelFile.getType(field.getType());
-        return (typeDeclaration !== null && typeDeclaration?.isMapDeclaration?.());
+        return typeDeclaration?.isMapDeclaration?.();
     }
 
     /**
@@ -223,7 +223,7 @@ class ModelUtil {
     static isScalar(field) {
         const modelFile = field.getParent().getModelFile();
         const declaration = modelFile.getType(field.getType());
-        return (declaration !== null && declaration.isScalarDeclaration?.());
+        return declaration?.isScalarDeclaration?.();
     }
 
     /**
--- a/packages/concerto-core/lib/modelutil.js
+++ b/packages/concerto-core/lib/modelutil.js
@@ -211,7 +211,7 @@ class ModelUtil {
     static isMap(field) {
         const modelFile = field.getParent().getModelFile();
         const typeDeclaration = modelFile.getType(field.getType());
-        return (typeDeclaration !== null && typeDeclaration?.isMapDeclaration());
+        return (typeDeclaration !== null && typeDeclaration?.isMapDeclaration?.());
     }
 
     /**
--- a/packages/concerto-core/lib/modelutil.js
+++ b/packages/concerto-core/lib/modelutil.js
@@ -203,6 +203,18 @@ class ModelUtil {
     }
 
     /**
+     * Returns the true if the given field is an map type
+     * @param {Field} field - the string
+     * @return {boolean} true if the field is declared as an map
+     * @private
+     */
+    static isMap(field) {
+        const modelFile = field.getParent().getModelFile();
+        const typeDeclaration = modelFile.getType(field.getType());
+        return (typeDeclaration !== null && typeDeclaration?.isMapDeclaration());
+    }
+
+    /**
      * Returns the true if the given field is a Scalar type
      * @param {Field} field - the Field to test
      * @return {boolean} true if the field is declared as an scalar
--- a/packages/concerto-core/lib/serializer.js
+++ b/packages/concerto-core/lib/serializer.js
@@ -174,7 +174,7 @@ class Serializer {
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
         } else if (classDeclaration.isMapDeclaration?.()) {
-            throw new Error('Attempting to create an Map declaration is not supported.');
+            throw new Error('Attempting to create a Map declaration is not supported.');
         } else if (classDeclaration.isEnum()) {
             throw new Error('Attempting to create an ENUM declaration is not supported.');
         } else {
--- a/packages/concerto-core/lib/serializer.js
+++ b/packages/concerto-core/lib/serializer.js
@@ -173,6 +173,8 @@ class Serializer {
             resource = this.factory.newConcept(classDeclaration.getNamespace(),
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
+        } else if (classDeclaration.isMapDeclaration()) {
+            throw new Error('Attempting to create an Map declaration is not supported.');
         } else if (classDeclaration.isEnum()) {
             throw new Error('Attempting to create an ENUM declaration is not supported.');
         } else {
--- a/packages/concerto-core/lib/serializer.js
+++ b/packages/concerto-core/lib/serializer.js
@@ -161,19 +161,19 @@ class Serializer {
 
         // create a new instance, using the identifier field name as the ID.
         let resource;
-        if (classDeclaration.isTransaction()) {
+        if (classDeclaration.isTransaction?.()) {
             resource = this.factory.newTransaction(classDeclaration.getNamespace(),
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
-        } else if (classDeclaration.isEvent()) {
+        } else if (classDeclaration.isEvent?.()) {
             resource = this.factory.newEvent(classDeclaration.getNamespace(),
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
-        } else if (classDeclaration.isConcept()) {
+        } else if (classDeclaration.isConcept?.()) {
             resource = this.factory.newConcept(classDeclaration.getNamespace(),
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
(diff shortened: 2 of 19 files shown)


packages/concerto-core/lib/modelutil.js
```diff
@@ -202,6 +202,18 @@ class ModelUtil {
...
+    /**
+     * Returns the true if the given field is an map type
+     * @param {Field} field - the string
+     * @return {boolean} true if the field is declared as an map
+     * @private
+     */
+    static isMap(field) {
+        const modelFile = field.getParent().getModelFile();
+        const typeDeclaration = modelFile.getType(field.getType());
+        return (typeDeclaration !== null && typeDeclaration?.isMapDeclaration?.());
```
mttrbrts, line 214: `typeDeclaration !== null` is redundant given the nullish check

packages/concerto-core/lib/serializer.js
```diff
@@ -161,18 +161,20 @@ class Serializer {
...
             resource = this.factory.newEvent(classDeclaration.getNamespace(),
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
-        } else if (classDeclaration.isConcept()) {
+        } else if (classDeclaration.isConcept?.()) {
             resource = this.factory.newConcept(classDeclaration.getNamespace(),
                 classDeclaration.getName(),
                 jsonObject[classDeclaration.getIdentifierFieldName()] );
+        } else if (classDeclaration.isMapDeclaration?.()) {
+            throw new Error('Attempting to create an Map declaration is not supported.');
```
jamieshorten, line 177: nitpicking: `an Map` should be `a Map`

jamieshorten (approved):

packages/concerto-core/lib/serializer.js
jamieshorten: Apart from Matts comments above I can't see any issues.

packages/concerto-core/test/model/concept.js
```diff
@@ -128,6 +128,17 @@ describe('Concept', function () {
...
 
+        it('should generate an error trying to create an Map from JSON', function () {
+            let conceptModel = fs.readFileSync('./test/data/model/concept.cto', 'utf8');
+            modelManager.addCTOModel(conceptModel, 'concept.cto');
+            const factory = new Factory(modelManager);
+            const serializer = new Serializer(factory, modelManager);
+            const jsObject = JSON.parse('{"$class":"org.acme.biznet.Dictionary"}');
+            (function () {
+                serializer.fromJSON(jsObject);
+            }).should.throw(/Attempting to create an Map declaration is not supported./);
```
mttrbrts, line 139: Will this be supported in future, or is there some more fundamental issue?

packages/concerto-core/test/serializer.js
```diff
@@ -232,6 +243,60 @@ describe('Serializer', () => {
...
 
+        it('should generate concept with a map value', () => {
+            let address = factory.newConcept('org.acme.sample', 'Address');
+            address.city = 'Winchester';
+            address.country = 'UK';
+            address.elevation = 3.14;
+            address.postcode = 'SO21 2JN';
+            address.dict = {
+                $class: 'org.acme.sample.Dictionary',
+                value: new Map(Object.entries({'Lorem':'Ipsum'}))
```
dselman, line 254: Yeah, this looks pretty ugly to me, compared to supporting JS objects here... rather than forcing people to use the Map type.

dselman (changes requested):
Some questions:
1. Don't we want to support Map values other than String?
2. Runtime type for maps, support JS object?

packages/concerto-core/test/model/concept.js
Jonathan-Casey: I'll answer with a question—does it make sense to serialize a MapDeclaration independently of a concept?

Jonathan-Casey:
> Some questions:
> 
> 1. Don't we want to support Map values other than String?
> 2. Runtime type for maps, support JS object?

1. Yes, but in a future phase. The alignment on the first release was <String, String> support only. (The rest is coming, I promise!)
2. As per [spec](https://github.com/accordproject/concerto/wiki/Aggregate-Types-Design): `Maps should be implemented using the JavaScript [Map object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map). We inherit the semantics of that implementation here.`

packages/concerto-core/messages/en.json
```diff
@@ -69,6 +69,7 @@
         "resourcevalidator-undeclaredfield": "Instance \"{resourceId}\" has a property named \"{propertyName}\", which is not declared in \"{fullyQualifiedTypeName}\".",
         "resourcevalidator-invalidfieldassignment": "Instance \"{resourceId}\" has a property \"{propertyName}\" with type \"{objectType}\" that is not derived from \"{fieldType}\".",
         "resourcevalidator-emptyidentifier": "Instance \"{resourceId}\" has an empty identifier.",
+        "resourcevalidator-invalidmap": "Model violation in the \"{resourceId}\" instance. Invalid Type for Map Key or Value - expected String type.",
```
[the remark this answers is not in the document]
Jonathan-Casey: As per comment below: The alignment on the first release was <String, String> support only.

packages/concerto-core/lib/serializer/valuegenerator.js
```diff
@@ -328,6 +336,14 @@ class SampleValueGenerator extends EmptyValueGenerator {
         return enumValues[Math.floor(Math.random() * enumValues.length)];
     }
 
+    /**
+     * Get a map instance with randomly generated values for key & value.
+     * @return {*} an enum value.
+     */
+    getMap() {
+        return new Map([[this.getString(1,10), this.getString(1,10)]]);
```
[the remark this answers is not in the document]
Jonathan-Casey: Yes, but for now we are supporting <String, String> only. In future the method will generate random values of all types supported.

packages/concerto-core/lib/serializer/instancegenerator.js
```diff
@@ -190,7 +194,7 @@ class InstanceGenerator {
      * @throws {Error} if no concrete subclasses exist.
      */
     findConcreteSubclass(declaration) {
-        if (!declaration.isAbstract()) {
+        if (!declaration.isAbstract?.()) {
```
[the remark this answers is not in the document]
Jonathan-Casey: Done. I'd prefer a terse check over an explicit check per declaration type, and given we have Union and Set declarations on the roadmap, we should probably add `isAbstract()` to those classes and MapDeclaration in order to avoid bloated conditional here.

packages/concerto-core/lib/serializer.js
```diff
@@ -161,18 +161,20 @@ class Serializer {
 
         // create a new instance, using the identifier field name as the ID.
         let resource;
-        if (classDeclaration.isTransaction()) {
+        if (classDeclaration.isTransaction?.()) {
```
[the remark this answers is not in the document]
Jonathan-Casey: MapDeclaration doesn't implement `isTransaction()`

packages/concerto-core/lib/modelutil.js
Jonathan-Casey: done+cleaned up the other expressions while there.

packages/concerto-core/test/serializer.js
mttrbrts: I think that there's a crossed wire here...

We absolutely must support deserialization from vanilla JSON.

However, this test is creating a Resource instance with an internal representation of the map. We could argue that that is not a good design, but that's separate from de(serialization) IMO.

packages/concerto-core/test/model/concept.js
mttrbrts: No, but we should be able to for a MapDeclaration within a Concept.

packages/concerto-core/lib/serializer.js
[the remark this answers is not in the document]
mttrbrts: Perhaps it should? we could move the function to the `Declaration` class?

dselman (approved):

DS-AdamMilazzo:
@Jonathan-Casey @mttrbrts @dselman I think it's very unfortunate (and ugly) that the map is pushed down into a "value" property. It seems that the primary reason is so that you can put Concerto metadata like $class at the top level of the map without colliding with user keys. However, I'd argue that there is virtually no reason to ever put such metadata there.

If you don't intend to allow a map to be extended, e.g. `map Foo { ... }  map Bar extends Foo { ... }`, then there is no need for a $class, because the type will always equal the declared type of the field. (Furthermore, there is no need for $timestamp, $identifier, etc. because those are for concepts and have no meaning for a map.)

Let's say you _were_ considering map extension. I'd argue it's a bad idea. There has to be a relation between the keys and values of the base and derived maps. In particular, the derived keys and values could either be a subset or a superset of the base keys and values. In this case...
```
concept B { }
concept D extends B { }
concept D2 extends B { }
map M {
  o String
  o D
}
map M2 extends M {
  o String
  o B
}
```
This violates the basic expectations of polymorphism, because you can store a B or D2 in M2, thereby violating the requirements of the base map M. This should just never be allowed, from the standpoint of a type system.

In the other case...
```
concept B { }
concept D extends B { }
concept D2 extends B { }
map M {
  o String
  o B
}
map M2 extends M {
  o String
  o D
}
```
First of all, you can already store a D (or D2) as a value in map M via basic polymorphism on B, with no need to extend map M to make that happen. Second, it's backwards from how `extends` would work for concepts and enums, where the derived type is a superset of the base type. Third, it violates a basic expectation of polymorphism, that a derived type can be treated as a base type. If you have an instance of map M2, you can't store a B or D2 in it. But if you treat M2 as an instance of M, which it _is_, then you _can_ store a B or D2 in it. That seems kind of broken, conceptually. Do you really want to allow this?

It's not _completely_ crazy, though. It can make some sense if we analogize to scalars. Let's say there was a way to extend a scalar type:
```
scalar PhoneNumber extends String regex=/^[0-9](-[0-9]+)*$/
scalar USPhoneNumber extends PhoneNumber regex=/^([1-9][0-9]{2}-){2}[0-9]{4}$/
```
A derived scalar would be a subset of the base scalar, just as a scalar now is a subset of the base type (e.g. String), and just as, in this case, a derived map is a subset of a base map. The important question, though, is how do you know when to apply the PhoneNumber rules versus the UsPhoneNumber rules? Strings have no $class to tell you whether a string is a PhoneNumber or a USPhoneNumber. So how could you do it? It would have to be done based on the declared field type. That is to say, _you have extension without polymorphism_.

And if maps are extended analogously to scalars (i.e. they are a _restriction_ of the base type, not an expansion), and if map polymorphism violates some expectations of polymorphism, then if we allow map extension at all (which I'd argue is a pretty pointless feature, unlike scalar extension with its layered constraints), it should _also_ be extension without polymorphism. And without polymorphism, you need no $class property.

Let's say you _did_ want to allow map extension, and you _did_ want polymorphism for maps. Does this mean a $class property would be required? Let's say you have `concept C { o M map }` and this instance data:
```
{
  "map": {
    "key": { "$class": "ns.D" }
  }
}
```

* Would you need a $class property to read the data? No. All derived maps are instances of the base map type, M, which we know from the field definition.
* Would you need a $class property to write the data? If you're writing valid data, then no, but if you're writing invalid data, then yes.
* Do you need a $class property to validate the data? Yes, in order to apply the added type constraint from the derived map.

So to summarize:
* Maps don't need $class unless they support both extension and polymorphism.
* Map extension only makes sense as a restriction of the base map, analogous to extension of scalars.
* Unlike scalar extension, map extension would be a pretty useless feature (and would also raise questions like "how do we combine the decorators?").
* Map extension would have as much need for polymorphism as scalar extensions, i.e. none.
* Map extension also violates some expectations of polymorphism and so shouldn't have it anyway.

Overall, the case for needing $class in a map is extremely thin. So can we please move the map data out of the "value" property? This:

```
{
  "foo": "bar"
}
```
looks so much nicer than:
```
{
  "$class": "ns@version.maptype",
  "value": {
    "foo": "bar"
  }
}
```

Jonathan-Casey:
@DS-AdamMilazzo thanks for your feedback. I agree with you on the `$class` value remarks, as do some others, and ultimately the communities voice takes precedent and the change has followed in a subsequent PR. 

In future it would be most helpful to share your opinion before a sign off. The spec was public and shared for several months prior to implementation. In any case, your feedback is always welcome and appreciated.

PR #677, opened 2023-08-01
feat(map): Extend Type support

Jonathan-Casey:
### Changes

This PR extends the current Type support for MapDeclaration Serialisation & Deserialisation to include:

### Key Type Support
- Primitives: String, DateTime
- Scalar String, DateTime

### Value Type Support
- Primitives: String, Integer, Long, Double, Boolean
- Scalar : String, Integer, Long, Double, Boolean
- Concepts: Identified & Non Identified

## Sample

Given the ModelFile: 

```
namespace org.acme.sample

concept Concepts {
    o Appointment appointment optional
    o Rolodex rolodex optional
    o Points points optional
}

scalar Time extends DateTime

concept Person identified by name {
    o String name
}

map Appointment {
    o Time
    o String
}

map Rolodex {
    o String
    o Person
}

map Points {
    o String
    o Long
}
```

The Serialiser will generate & validate against the following format

**<Scalar, String>**
```
{
  $class: 'org.acme.sample.Concepts',
  appointment: {
    '$class': 'org.acme.sample.Appointment',
    '2023-11-28T01:02:03Z': 'BobBirthday',
    '2024-10-28T01:02:03Z': 'AliceAnniversary'
  }
}
```

**<String, Concept>**
```
{
  $class: 'org.acme.sample.Concepts',
  rolodex: {
    $class: 'org.acme.sample.Rolodex',
    'Dublin': '{"$class":"org.acme.sample.Person","name":"Bob"}',
    'London': '{"$class":"org.acme.sample.Person","name":"Alice"}'
  }
}


```

**<String, Long>**
```
{
  $class: 'org.acme.sample.Concepts',
  points: {
    $class: 'org.acme.sample.Points',
    Bob: -398741129664271,
    Alice: 8999999125356546
  }
}
```

## AST

Below is a sample of a MapDeclaration entry in the AST:

```
{
  $class: 'concerto.metamodel@1.0.0.MapDeclaration',
  name: 'MapPermutation1',
  key: {
    $class: 'concerto.metamodel@1.0.0.StringMapKeyType'
  },
  value: {
   $class: 'concerto.metamodel@1.0.0.StringMapValueType'
  }
}
```

In the case of a Key or Value which is an Object, a TypeIdentifier is now present
```
{
  $class: 'concerto.metamodel@1.0.0.MapDeclaration',
  name: 'MapPermutation1',
  key: {
    $class: 'concerto.metamodel@1.0.0.StringMapKeyType',
  },
  value: {
    $class: 'concerto.metamodel@1.0.0.ObjectMapValueType'
    type: {
        $class: 'concerto.metamodel@1.0.0.TypeIdentifier',
        name: 'Person'
    }
  }
}
```

### Key

```
concerto.metamodel@1.0.0.StringMapKeyType
concerto.metamodel@1.0.0.DateTimeMapKeyType
concerto.metamodel@1.0.0.ObjectMapKeyType
```

### Value

```
concerto.metamodel@1.0.0.BooleanMapValueType
concerto.metamodel@1.0.0.DateTimeMapValueType
concerto.metamodel@1.0.0.StringMapValueType
concerto.metamodel@1.0.0.IntegerMapValueType
concerto.metamodel@1.0.0.LongMapValueType
concerto.metamodel@1.0.0.DoubleMapValueType
concerto.metamodel@1.0.0.ObjectMapValueType
```


packages/concerto-core/lib/serializer/jsonpopulator.js
```diff
@@ -170,13 +170,64 @@ class JSONPopulator {
...
+
+        jsonObj.forEach((value, key) => {
+
+            if (key === '$class') {
+                map.set(key, value);
+                return;
+            }
+
+            // If its a Non-Primitive, its likely a ClassDeclaration which needs visiting.
+            if (!ModelUtil.isSystemProperty(key) && !ModelUtil.isPrimitiveType(mapDeclaration.getKey().getType())) {
```
mttrbrts, line 191: Why do we check for system properties here? We already checked on line 177

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -129,6 +129,24 @@ class MapKeyType extends Decorated {
...
+        return true;
+    }
+
+    /**
+     * Returns true if this class is the definition of a Map Value.
+     *
+     * @return {boolean} true if the class is a Map Value
+     */
+    isValue() {
+        return false;
```
mttrbrts, line 148: Add a test for this line.

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -121,6 +121,24 @@ class MapValueType extends Decorated {
...
+        return false;
+    }
+
+    /**
+     * Returns true if this class is the definition of a Map Value.
+     *
+     * @return {boolean} true if the class is a Map Value
+     */
+    isValue() {
+        return true;
```
mttrbrts, line 139: Add a test for this line.

packages/concerto-core/test/serializer/maptype/serializer.js
```diff
@@ -0,0 +1,926 @@
...
+            concepts.marriages.set('$class', 'org.acme.sample.MarriageRegister');
+            concepts.marriages.set(bob, alice);
+
+            const json = serializer.toJSON(concepts);
+
+            json.should.deep.equal({
+                $class: 'org.acme.sample.Concepts',
+                marriages: {
+                    $class: 'org.acme.sample.MarriageRegister',
+                    '{"$class":"org.acme.sample.Person","name":"Bob"}' : '{"$class":"org.acme.sample.Person","name":"Alice"}'
```
mttrbrts, line 587: This doesn't feel write to me. Certainly for the value. 

Should the key be the identifier for the Concept instead?

packages/concerto-core/lib/serializer/resourcevalidator.js
```diff
@@ -106,39 +110,99 @@ class ResourceValidator {
...
+                throw new Error(`Model violation in ${mapDeclaration.getFullyQualifiedName()}. Expected Type of DateTime but found '${value}' instead.`);
+            }
+            break;
+        case 'Boolean':
+            if (typeof value !== 'boolean') {
+                const type = typeof value;
+                throw new Error(`Model violation in ${mapDeclaration.getFullyQualifiedName()}. Expected Type of Boolean but found ${type} instead, for value '${value}'.`);
+            }
+            break;
+        }
```
mttrbrts, line 170: Can we reuse the implementation of checkItem here?

packages/concerto-core/test/serializer/maptype/serializer.js
Jonathan-Casey: As discussed, dropping support for this until we agree on a concrete design. @dselman FYI

packages/concerto-core/lib/serializer/jsonpopulator.js
```diff
@@ -170,13 +170,64 @@ class JSONPopulator {
...
+                    let subResource = parameters.factory.newConcept(decl.getNamespace(),
+                        decl.getName(), decl.getIdentifierFieldName() );
+
+                    parameters.jsonStack.push(JSON.parse(value));
+                    parameters.resourceStack.push(subResource);
+
+                    value = decl.accept(this, parameters);
+                }
+            }
+
```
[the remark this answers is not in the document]
Jonathan-Casey: Yep, this is done

packages/concerto-core/lib/serializer/jsonpopulator.js
Jonathan-Casey: You're right, fixed.

packages/concerto-core/lib/introspect/mapdeclaration.js
```diff
@@ -180,6 +168,38 @@ class MapDeclaration extends Declaration {
...
+        return [
+            `${MetaModelNamespace}.BooleanMapValueType`,
+            `${MetaModelNamespace}.DateTimeMapValueType`,
+            `${MetaModelNamespace}.StringMapValueType`,
+            `${MetaModelNamespace}.IntegerMapValueType`,
+            `${MetaModelNamespace}.LongMapValueType`,
+            `${MetaModelNamespace}.DoubleMapValueType`,
+            `${MetaModelNamespace}.ObjectMapValueType`
+        ].includes(value.$class);
+    }
```
mttrbrts, line 202: These would fit better in the `ModelUtil` class, IMO

packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -64,23 +67,64 @@ class MapValueType extends Decorated {
...
+
+            // ObjectMapValueType must have TypeIdentifier.
+            if (!('type' in ast)) {
+                throw new IllegalModelException(`ObjectMapValueType must contain property 'type', for MapDeclaration named ${this.parent.name}`);
+            }
+
+            // ObjectMapValueType TypeIdentifier must be properly formed.
+            if (!('$class' in ast.type) || !('name' in ast.type)) {
+                throw new IllegalModelException(`ObjectMapValueType type must contain property '$class' and property 'name', for MapDeclaration named ${this.parent.name}`);
+            }
```
mttrbrts, line 101: Great defensive programming. This should also be checked now by the Metamodel validation in #684, so we may find that we have dead code once both PRs are merged.

packages/concerto-core/lib/serializer/jsonpopulator.js
```diff
@@ -170,13 +170,61 @@ class JSONPopulator {
...
+    }
+
+    /**
+     * Visitor design pattern
+     * @param {MapDeclaration} mapDeclaration - the object being visited
+     * @param {Object} parameters  - the parameter
+     * @param {Object} value - the key or value belonging to the Map Entry.
+     * @param {Object} type - the Type associated with the Key or Value Map Entry.
+     * @return {Object} value - the key or value belonging to the Map Entry.
+     */
```
mttrbrts, line 211:
```suggestion
     * @private
     */
```

packages/concerto-core/types/lib/introspect/mapkeytype.d.ts
```diff
@@ -17,14 +17,20 @@ declare class MapKeyType extends Decorated {
...
      * @throws {IllegalModelException}
      * @protected
      */
     protected validate(): void;
+    /**
+     * Sets the Type name for the Map Key
+     *
+     * @param {Object} ast - The AST created by the parser
+     */
+    processType(ast: any): void;
```
mttrbrts, line 32: This should be an internal function only

packages/concerto-core/types/lib/introspect/mapvaluetype.d.ts
```diff
@@ -16,14 +16,20 @@ declare class MapValueType extends Decorated {
...
      * @throws {IllegalModelException}
      * @protected
      */
     protected validate(): void;
+    /**
+     * Sets the Type name for the Map Value
+     *
+     * @param {Object} ast - The AST created by the parser
+     */
+    processType(ast: any): void;
```
mttrbrts, line 31: This should be an internal function only

packages/concerto-core/lib/introspect/mapvaluetype.js
Jonathan-Casey: Noted for cleanup

packages/concerto-core/types/lib/introspect/mapkeytype.d.ts
Jonathan-Casey: I did intend to use `#private` class feature but it produces the error `Private identifiers are only available when targeting ECMAScript 2015 and higher`.

Is there a convention in place for handling internal functions, or do you have a suggestion? I see `@private` used, but my understanding is this is for doc-gen purposes.

DS-AdamMilazzo:
I'd argue that we shouldn't have a $class attribute within the map values. In your string/string map example, say...
```
{
  $class: 'org.acme.sample.Concepts',
  appointment: {
    '$class': 'org.acme.sample.Appointment',
    '2023-11-28T01:02:03Z': 'BobBirthday',
    '2024-10-28T01:02:03Z': 'AliceAnniversary'
  }
}
```
... what values could the "$class" key have? It seems like "org.acme.sample.Appointment" (the declared field type) is the only legal value.

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -67,30 +68,46 @@ class MapKeyType extends Decorated {
...
+                !(decl?.ast.$class === `${MetaModelNamespace}.StringScalar`)  &&
+                !(decl?.ast.$class === `${MetaModelNamespace}.DateTimeScalar`)) {
+                throw new IllegalModelException(
+                    `Scalar must be one of StringScalar, DateTimeScalar in context of MapKeyType. Invalid Scalar: ${this.type}, for MapDeclaration ${this.parent.name}`
+                );
+            }
+
+            if (decl?.isConcept?.() || decl?.isClassDeclaration?.()) {
+                throw new IllegalModelException(
+                    `MapKeyType supports String and DateTime only,for MapDeclaration ${this.parent.name}`
```
mttrbrts, line 85:
```suggestion
                    `Invalid Map key type in MapDeclaration ${this.parent.name}. Only String and DateTime types are supported for Map key types`
```

packages/concerto-core/lib/introspect/mapkeytype.js
```diff
@@ -67,30 +68,46 @@ class MapKeyType extends Decorated {
...
-        if (key?.isConcept?.() && !key.isIdentified()) {
-            throw new IllegalModelException(
-                `ConceptDeclaration must be identified in context of MapKeyType: ${this.type}`
-            );
+        if (!ModelUtil.isPrimitiveType(this.type)) {
+            let decl = this.parent.getModelFile().getAllDeclarations().find(d => d.name === this.ast.type?.name);
+
+            if  (decl?.isScalarDeclaration?.() &&
+                !(decl?.ast.$class === `${MetaModelNamespace}.StringScalar`)  &&
+                !(decl?.ast.$class === `${MetaModelNamespace}.DateTimeScalar`)) {
```
mttrbrts, line 77: I'm surprised that we don't have a util function to encapsulate this?

packages/concerto-core/lib/serializer/jsongenerator.js
```diff
@@ -85,7 +85,35 @@ class JSONGenerator {
      */
     visitMapDeclaration(mapDeclaration, parameters) {
         const obj = parameters.stack.pop();
-        return Object.fromEntries(obj);
+
+        // initialise Map with $class property
+        let map = new Map([['$class',obj.get('$class')]]);
+
+        obj.forEach((value, key) => {
+
```
mttrbrts, line 93:
```suggestion
```

mttrbrts (approved):

DS-AdamMilazzo:
> Without a discriminator in the instance, the client application could assign a Phonebook to a Dictionary (and visa-versa) and the parent concept would validate. Type conversion like this should be explicit, IMO.

I think the same is true for all scalar types, both implicit and explicit. For example, you can assign any string field value to any other string field regardless of the validators applied to it. I see maps as values, like scalars are. As argued previously, their semantics make the most sense when treated analogously to scalar values. Now perhaps you'd say that's a problem with scalars that we shouldn't propagate to maps, but I'm not sure it's a problem. In both cases it could be handled via stronger typing, as you mention, but that might be a medicine worse than the disease if there were wrappers for every scalar and map type.

> The small optimisation omitting `$class` gives us doesn't justify the loss of precision that we'd get in some scenarios.

I don't see it as an optimization but as a matter of consistency and convenience. "$class" would have to be treated specially in all maps because it's the one key that is metadata rather than data. People looping over map values would have to specifically ignore it, and deserializing types like the Appointment map above into Dictionary\<DateTime, string\> wouldn't work easily because "$class" is not a valid DateTime, so some kind of custom serializer and deserializer would need to be used. I think it makes a bunch of things more onerous than they need to be. There's also the admittedly small matter that people wouldn't be able to use a key called "$class" because it's interpreted differently from the other keys.

mttrbrts:
> I see maps as values, like scalars are.

I see your perspective here. The (perhaps implicit) design here is that Maps are not primitive values, but rather are compound values. The design mirrors what we do for Concepts and other class declarations already. 

> some kind of custom serializer and deserializer would need to be used

We already implement our own (de)serialization logic for other reasons.

I support merging this PR as-is, but I would welcome a debate about the assignment of scalars, or reserved fields in class declarations. Please open issues as necessary.

DS-AdamMilazzo:
> I see your perspective here. The (perhaps implicit) design here is that Maps are not primitive values, but rather are compound values. The design mirrors what we do for Concepts and other class declarations already.

Yes, they're compound, but so are arrays and I think they're closer to an array in that sense than a concept. They can contain concepts, but they don't really support polymorphism in and of themselves. (The concepts inside support polymorphism as normal.)

> We already implement our own (de)serialization logic for other reasons.

Yes, but I don't think you have custom deserializers for more than a tiny subset of the languages that you can generate types for (just Javascript and a prototype for C#?). I think most people at this point will have to (or want to) use the built-in or standard JSON deserializers, which may choke on the $class properties in maps in a way that they won't for concepts. When deserializing classes, unknown properties are just ignored. When deserializing dictionaries, no properties are ignored.

PR #684, opened 2023-08-11
feat(core): add validation of model files against metamodel

### LINKED ISSUE #678: Parsed models are not validated against the metamodel
# Bug Report 🐛

The parser doesn't validate the JSON AST against the declared metamodel. This means that we could be creating instances that are not valid with respect to the metamodel. It also makes adding new features more error prone (e.g. https://github.com/accordproject/concerto-metamodel/pull/6) 

## Possible Solution

- We could create a modelManager that we use internally for validation. This would have a small performance hit, but would be limited only to parsing.

mttrbrts:
# Closes #678

Adds an extra validation step in the ModelManager to validate `ModelFile` object against the Metamodel definition.

When opting into this feature: 
- In strict mode, invalid models will now throw an error.
- In the default non-strict mode, a console warning is printed.

Turn on the feature with a parameter to the ModelManager constructor, e.g. 
```
new ModelManager({ metamodelValidation: true });
```

### Changes

- Adds a new `MetamodelException` class to report errors found while validating the ModelFile
- Adds a new private function on the `BaseModelManager` that is used when adding a model file.

### Flags

- Ideally, I would have liked to build a separate ModelManager instance purely for this purpose. However, this introduced issues with circular dependencies (a ModelManager would then contain a ModelManager!). This approach mutates the client's ModelManager temporarily; this should be safe if we clean up properly after ourselves.
- We also have to disable validation when adding the RootModel to a ModelManager because it is now a pre-requisite for metamodel validation

### Related Issues
- Issue #678 

### Author Checklist


packages/concerto-core/test/modelmanager.js
```diff
@@ -262,6 +265,43 @@ describe('ModelManager', () => {
...
+        });
+
+        it('should warn for a bad metamodel AST', () => {
+            const basemodelmanager = new BaseModelManager();
+            const ast = {
+                $class: `${MetaModelNamespace}.Model`,
+                namespace: 'org.acme',
+                undeclared: []
+            };
+            // Displays console warning
```
ekarademir, line 288: Maybe better to check if an error is not thrown?

ekarademir (approved):

Jonathan-Casey (approved):

PR #722, opened 2023-09-29
feat(dcs): add map type support for decorator command targets

Jonathan-Casey:
### Description
This change extends the Decorator Manager to support Map Types. A new optional property `mapElement` is added to the `CommandTarget` which allows users to specify the granularity of the intended target on a Map. The legal options of which are `KEY`, `VALUE`, or `KEY_VALUE` (both Key and Value). 

Example 1:

```
{
    "$class" : "org.accordproject.decoratorcommands@0.2.0.DecoratorCommandSet",
    "name" : "web",
    "version": "1.0.0",
    "commands" : [
 
        {
            "$class" : "org.accordproject.decoratorcommands@0.2.0.Command",
            "type" : "UPSERT",
            "target" : {
                "$class" : "org.accordproject.decoratorcommands@0.2.0.CommandTarget",
                "namespace" : "test@1.0.0",
                "declaration" : "Dictionary",
                "mapElement": "KEY_VALUE"
            },
            "decorator" : {
                "$class" : "concerto.metamodel@1.0.0.Decorator",
                "name" : "Foo",
                "arguments" : []
            }
        }
    ]
}
```

In the above sample, the `@Foo` decorator will be added to both elements for the Map Declaration `Dictionary`

Example 2:

```
{
    "$class" : "org.accordproject.decoratorcommands@0.2.0.DecoratorCommandSet",
    "name" : "web",
    "version": "1.0.0",
    "commands" : [
 
        {
            "$class" : "org.accordproject.decoratorcommands@0.2.0.Command",
            "type" : "UPSERT",
            "target" : {
                "$class" : "org.accordproject.decoratorcommands@0.2.0.CommandTarget",
                "namespace" : "test@1.0.0",
                "declaration" : "Dictionary",
                "type" : "concerto.metamodel@1.0.0.StringMapKeyType",
            },
            "decorator" : {
                "$class" : "concerto.metamodel@1.0.0.Decorator",
                "name" : "Foo",
                "arguments" : []
            }
        }
    ]
}
```

In the above sample, the `@Foo` decorator will be added to the Dictionary Map Declaration, if it contains a StringMapKeyType. 

Example 3:

```
{
    "$class" : "org.accordproject.decoratorcommands@0.2.0.DecoratorCommandSet",
    "name" : "web",
    "version": "1.0.0",
    "commands" : [
 
        {
            "$class" : "org.accordproject.decoratorcommands@0.2.0.Command",
            "type" : "UPSERT",
            "target" : {
                "$class" : "org.accordproject.decoratorcommands@0.2.0.CommandTarget",
                "namespace" : "test@1.0.0",
                "type" : "concerto.metamodel@1.0.0.StringMapKeyType",
            },
            "decorator" : {
                "$class" : "concerto.metamodel@1.0.0.Decorator",
                "name" : "Foo",
                "arguments" : []
            }
        }
    ]
}
```

In the above sample, the `@Foo` decorator will be added to all MapDeclarations found in the ModelFile, which contain a StringMapKeyType. 

### Author Checklist


packages/concerto-core/lib/decoratormanager.js
```diff
@@ -296,12 +297,57 @@ class DecoratorManager {
      */
     static executeCommand(namespace, declaration, command) {
         const { target, decorator, type } = command;
-        const { name } = ModelUtil.parseNamespace(namespace);
-        if (
-            this.falsyOrEqual(target.namespace, [namespace, name]) &&
-            this.falsyOrEqual(target.declaration, [declaration.name])
-        ) {
-            if (!target.property && !target.type) {
+        const { name } = ModelUtil.parseNamespace( namespace );
```
dselman, line 300: This seems like a lot compared to the existing code... 50 lines to add decorators to map key/value. No way to simplify?

dselman (commented):
Looks good. `element` sounds very generic ... Perhaps something more explicitly (map specific) would help in the model? `mapElement` ?

Alternatively we could co-opt `property` and `properties` for maps, but use special property names so we don't need a model change.

```
property: 'MAP_KEY'
property: 'MAP_VALUE'
properties: ['MAP_KEY', 'MAP_VALUE']
```

Thinking about this - we will also need to update the `Vocabulary` and `VocabularyManager` to check that terms can be retrieved for map keys and values.

Jonathan-Casey:
> Looks good. `element` sounds very generic ... Perhaps something more explicitly (map specific) would help in the model? `mapElement` ?
> 

Yes, I'm good with mapElement, although I liked the elegance of `element`, I see your point. Given we don't typically reference anything else as an element I felt it sits nicely in the command.

> Alternatively we could co-opt `property` and `properties` for maps, but use special property names so we don't need a model change.
> 
> ```
> property: 'MAP_KEY'
> property: 'MAP_VALUE'
> properties: ['MAP_KEY', 'MAP_VALUE']
> ```
> 

I am less in favour of this suggestion . Strictly speaking, Maps do not contain properties. I think we should align the language of the command to that of the models AST, so as to provide consistency and avoid confusion.

> Thinking about this - we will also need to update the `Vocabulary` and `VocabularyManager` to check that terms can be retrieved for map key and value.

Yes this came up in a team meeting, will be implemented as part of the ongoing Vocabulary efforts.

packages/concerto-core/lib/decoratormanager.js
Jonathan-Casey: I've isolated logic to `applyDecoratorForMapElement()`

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -296,12 +318,36 @@ class DecoratorManager {
...
+        if (this.falsyOrEqual(target.namespace, [namespace,name]) &&
+            this.falsyOrEqual(target.declaration, [declaration.name])) {
+
+            if (declaration.$class === 'concerto.metamodel@1.0.0.MapDeclaration') {
```
dselman, line 325: const { MetaModelNamespace } = require('@accordproject/concerto-metamodel');

Can we please use `${MetaModelNamespace}.MapDeclaration`?

packages/concerto-core/test/data/decoratorcommands/invalid-target-element.json
```diff
@@ -0,0 +1,21 @@
...
+    "version": "1.0.0",
+    "commands" : [
+        {
+            "$class" : "org.accordproject.decoratorcommands@0.3.0.Command",
+            "type" : "UPSERT",
+            "target" : {
+                "$class" : "org.accordproject.decoratorcommands@0.3.0.CommandTarget",
+                "namespace" : "test@1.0.0",
+                "declaration" : "Dictionary",
+                "mapElement" : "INVALID_ELEMENT"
```
dselman, line 13: Using an enum in your model would make this test redundant.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -50,6 +50,7 @@ concept CommandTarget {
     o String property optional
     o String[] properties optional // property and properties are mutually exclusive
     o String type optional 
+    o String mapElement optional
```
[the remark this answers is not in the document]
dselman: We should also bump the model version to 3.1.0 out of an abundance of caution...?

dselman (changes requested):
Very close! I think making the mapElement an enum will make the model self-describing and improve validation.

packages/concerto-core/lib/decoratormanager.js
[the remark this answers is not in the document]
Jonathan-Casey: Agree, using an Enum set makes sense now that we have DecoratorSet validation.

packages/concerto-core/lib/decoratormanager.js
Jonathan-Casey: I was looking for that, thanks!

dselman:
> > Very close! I think making the mapElement an enum will make the model self-describing and improve validation.
> 
> Agree using an Enum set makes sense now that we have DecoratorSet validation changes merged earlier today.
> 
> One thing to clarify - do you mean to suggest bumping the concerto version, or the namespace version to `org.accordproject.decoratorcommands@0.4.0`? It makes sense to bump only the namespace as the changes are scoped to the `decoratorcommands` version.

Validate has been present for a while - the change yesterday exposed it as a standalone method, rather than just doing it implicitly as part of decorateModels.

I meant to bump to `org.accordproject.decoratorcommands@0.4.0`.

packages/concerto-core/test/data/decoratorcommands/map-declaration.json
```diff
@@ -0,0 +1,189 @@
+{
+    "$class" : "org.accordproject.decoratorcommands@0.3.0.DecoratorCommandSet",
+    "name" : "web",
+    "version": "1.0.0",
+    "commands" : [
+        {
+            "$class" : "org.accordproject.decoratorcommands@0.3.0.Command",
```
dselman, line 7: When you bump `org.accordproject.decoratorcommands` version you will have to find/replace in this file...

Jonathan-Casey:
@dselman given we haven't yet cut a release with the `org.accordproject.decoratorcommands@0.3.0` changes , I suggest we stay with version `0.2.0` for now. This will solve the emergent migration story (in the short term) at DocuSign which surfaces should we move from `v0.2.0`. Longer term, we need to consider how we support versioning across DCS models, and the metamodel,  in a scalable manner.

dselman:
The problem remaining at v0.2.0 for DCS model will cause is that someone can define a DCS with mapElement and use it with concerto-core v3.13.0 and it will fail validation.

I think downstream consumers need to migrate (on the fly or in batch) DCS docs to support what the version of concerto-core supports.

dselman (changes requested):
Looking good. A couple of minor changes requested.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -160,8 +196,29 @@ class DecoratorManager {
      * @returns {ModelManager} a new model manager with the decorations applied
      */
     static decorateModels(modelManager, decoratorCommandSet, options) {
+
+        // get the version of the input decoratorCommandSet from its $class property
+        const inputVersion = ModelUtil.parseNamespace(decoratorCommandSet.$class.replace(/\.([^.]*)$/, '')).version;
+
+        // if its < the currect DCS_Version, rewrite the $class version to match the supported DCS_VERSION
+        if (semver.lt(inputVersion, DCS_VERSION)) {
```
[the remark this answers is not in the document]
Jonathan-Casey: Good idea

dselman (approved):
Bravo!

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -157,11 +194,35 @@ class DecoratorManager {
...
      * @returns {ModelManager} a new model manager with the decorations applied
      */
     static decorateModels(modelManager, decoratorCommandSet, options) {
+
+        if (options?.migrate) {
+            // get the version of the decoratorCommandSet from its $class property
+            const inputVersion = ModelUtil.parseNamespace(ModelUtil.getNamespace(decoratorCommandSet.$class)).version;
+
+            // if its < the currect DCS_Version, rewrite the $class version to match the supported DCS_VERSION
+            if (semver.lt(inputVersion, DCS_VERSION)) {
```
mttrbrts, line 207: This could do strange things if `inputVersion` is not compatible with `DCS_VERSION`, right? e.g. v1.0.0 to v30.0.0

Jonathan-Casey: This function is designed to support changes across minor versions only (hence the original naming). When we get in to the territory or migrating the model across major versions, a rewrite (or a conceptual reconsideration) on how we manage DCS will be necessary.

Jonathan-Casey: Will add an exception to safeguard against the strangeness in the above scenario.

PR #727, opened 2023-10-09
feat(map): allows ENABLE_MAP_TYPE env var to be set via ModelManager constructor

Jonathan-Casey:
### Description

This change allows clients to set the `ENABLE_MAP_TYPE` environment variable via the `ModelManagers` options.  Clients can add any environment variable they wish to the options, but only `ENABLE_MAP_TYPE` will ever be set, if it is configured. This eliminates the possibility that Concerto is used as an attack vector under certain scenarios.

**This change is temporary and will be removed on full release of Map Type.**

### Usage

```
const options = { flags : {'ENABLE_MAP_TYPE': 'true'} }; // add the env var to options
modelManager = new ModelManager(options);
```

### DIFF
--- a/packages/concerto-core/lib/basemodelmanager.js
+++ b/packages/concerto-core/lib/basemodelmanager.js
@@ -93,6 +93,16 @@ class BaseModelManager {
         this.options = options;
         this.addRootModel();
 
+        // TODO remove on full release of Map Type.
+        // Temporarily allow for ENABLE_MAP_TYPE environment variable to be set from options.
+        if (options?.flags) {
+            for (const [key, value] of Object.entries(options.flags)) {
+                if(key === 'ENABLE_MAP_TYPE') {
+                    process.env[key] = value;
+                }
+            }
+        }
+
         // Cache a copy of the Metamodel ModelFile for use when validating the structure of ModelFiles later.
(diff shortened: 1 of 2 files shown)


packages/concerto-core/lib/basemodelmanager.js
```diff
@@ -93,6 +93,16 @@ class BaseModelManager {
         this.options = options;
         this.addRootModel();
 
+        // TODO remove on full release of Map Type.
+        // Temporarily allow for ENABLE_MAP_TYPE environment variable to be set from options.
+        if (options?.flags) {
+            for (const [key, value] of Object.entries(options.flags)) {
+                if(key === 'ENABLE_MAP_TYPE') {
+                    process.env[key] = value;
```
mttrbrts, line 101: Why can't a client application do this?

Jonathan-Casey: Some issues with Webpack from internal customer at DocuSign...

dselman: I don't think this will work once the code is web packed by Babel. I think the code that is reading process.env.ENABLE_MAP_TYPE will have to be updated to reference an options object in the ModelManager to allow this value to be set dynamically by web clients.

Jonathan-Casey: @dselman Matts concern is similar. I'll add a property based flag on the model manager to work around the issue.

PR #728, opened 2023-10-09
feat(map): add property based flag

Jonathan-Casey:
### Description

Add property based flag to `ModelManager` options as an alternative to setting process.env `ENABLE_MAP_TYPE='true`. 

**This change is temporary and will be removed on full release of Map Type.**

### Usage
`modelManager = new ModelManager({enableMapType: true});
`

### DIFF
--- a/packages/concerto-core/test/introspect/mapdeclaration.js
+++ b/packages/concerto-core/test/introspect/mapdeclaration.js
@@ -100,7 +100,7 @@ describe('MapDeclaration', () => {
             }
         });
 
-        it('should not throw if Map Type not enabled in ModelManager options', () => {
+        it('should not throw if Map Type is enabled in ModelManager options', () => {
             const mm = new ModelManager({enableMapType: true});
             Util.addComposerModel(mm);
             const introspectUtils = new IntrospectUtils(mm);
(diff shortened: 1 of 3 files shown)


packages/concerto-core/test/introspect/mapdeclaration.js
```diff
@@ -84,11 +84,29 @@ describe('MapDeclaration', () => {
...
+            Util.addComposerModel(mm);
+            const introspectUtils = new IntrospectUtils(mm);
+            try {
+                introspectUtils.loadLastDeclaration('test/data/parser/mapdeclaration/mapdeclaration.goodkey.primitive.datetime.cto', MapDeclaration);
+            } catch (error) {
+                expect(error.message).to.equal('MapType feature is not enabled. Please set the environment variable "ENABLE_MAP_TYPE=true", or add {enableMapType: true} to the ModelManger options, to access this functionality.');
+            }
+        });
+
+        it('should not throw if Map Type not enabled in ModelManager options', () => {
```
DianaLease, line 103: 'should not throw if Map Type ~not~ _is_ enabled in ModelManager options'

DianaLease (changes requested):
typo in test description. otherwise looks good!

packages/concerto-core/test/introspect/mapdeclaration.js
Jonathan-Casey: Thanks for spotting @DianaLease, typo is corrected.

DianaLease (approved):

PR #730, opened 2023-10-10
feat(analyser): add config builder to extend the analyser config

### LINKED ISSUE #720: Provide an API to extend the default rules of the analyser
# Feature Request 🛍️

When creating a `Compare` object, we can provide an `extend` method where we can just extend the `defaultCompareConfig`. Currently `defaultCompareConfig` is not a module level export, so in order to extend it we would need to import it from `@accordproject/concerto-analysis/dist/compare-config`.
## Use Case

`Compare` already allows to use custom configuration, so this would provide better ergonomics for when the user wants to use the default config as is but also wants to add few more rules.

## Context

I had to this exact thing in the work we are doing at DocuSign where we wanted to make the config more strict than the default, but also use the default config.

ekarademir:
# Closes #720 

Adds a builder to create a `CompareConfig` that can be used for creating `Compare` class.
### Changes

- Add a `CompareConfigBuilder` class that starts from an empty configuration and able to build up configuration as we go.
- As a final step user can call the `build` method to get the resulting configuration.

### Author Checklist

### DIFF
--- a/packages/concerto-analysis/src/compare-config.ts
+++ b/packages/concerto-analysis/src/compare-config.ts
@@ -16,55 +16,133 @@ import { ComparerFactory } from './comparer';
 import { comparerFactories } from './comparers';
 
 export enum CompareResult {
-    NONE,
-    PATCH,
-    MINOR,
-    MAJOR,
-    ERROR,
+  NONE,
+  PATCH,
+  MINOR,
+  MAJOR,
+  ERROR,
 }
 
 export function compareResultToString(result: CompareResult) {
-    switch (result) {
+  switch (result) {
     case CompareResult.NONE:
-        return 'none';
+      return 'none';
     case CompareResult.PATCH:
-        return 'patch';
+      return 'patch';
     case CompareResult.MINOR:
-        return 'minor';
+      return 'minor';
     case CompareResult.MAJOR:
-        return 'major';
+      return 'major';
     case CompareResult.ERROR:
-        return 'error';
-    }
+      return 'error';
+  }
 }
 
 export type CompareConfig = {
-    comparerFactories: ComparerFactory[];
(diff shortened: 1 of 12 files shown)


packages/concerto-analysis/src/compare-config.ts
```diff
@@ -63,8 +63,95 @@ export const defaultCompareConfig: CompareConfig = {
...
+    rules: {},
+};
+
+export class CompareConfigBuilder {
+    /**
+     * A utility to build CompareConfig to be used in Compare class.
+     * A new compare config can be edited with provided functions and finally
+     * resulting config can be used by calling `build`.
+     *
+     * By default, it starts with an emoty configuration.
```
mttrbrts, line 83:
```suggestion
     * By default, it starts with an empty configuration.
```

packages/concerto-analysis/src/compare-config.ts
```diff
@@ -63,8 +63,95 @@ export const defaultCompareConfig: CompareConfig = {
...
+    public build(): CompareConfig {
+        return {
+            comparerFactories: [...this._config.comparerFactories],
+            rules: { ...this._config.rules },
+        };
+    }
+
+    /**
+      * Adds default comparer configuration onto the configuration
+      * baing built.
```
mttrbrts, line 102:
```suggestion
      * being built.
```

ekarademir:
> Do you understand why we have the linting/whitespace changes? Is there a build step that we should be automating, or are these changes made by your editor's configuration?

What generated those differences was to run the `lint` command within the package with `--fix` option with auto formatting turned off. We can definitely add an automated step. I can look into that.

Jonathan-Casey (approved):

PR #739, opened 2023-10-19
feat(map): Support Map Key Type Imports

Jonathan-Casey:
### Description

Adds logic to support declaration of Map Key Type as an Imported Type. Additional test cases also added.

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/modelutil.js
+++ b/packages/concerto-core/lib/modelutil.js
@@ -19,6 +19,13 @@ const { MetaModelUtil } = require('@accordproject/concerto-metamodel');
 const semver = require('semver');
 const Globalize = require('./globalize');
 
+// Types needed for TypeScript generation.
+/* eslint-disable no-unused-vars */
+/* istanbul ignore next */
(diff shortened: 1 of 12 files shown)


mttrbrts (approved):

packages/concerto-core/lib/modelutil.js
```diff
@@ -331,6 +338,21 @@ class ModelUtil {
             `${MetaModelNamespace}.ObjectMapValueType`
         ].includes(value.$class);
     }
+
+    /**
+     * Returns the corresponding ClassDeclaration representation of the Map Type
```
dselman, line 343: Typo in the jsdoc - this is not map specific. 

Not sure we need this function. ModelFile.getType takes a short type name and returns the FQN, taking care of primitives and imported types.

So, this could be:

```
// decl will be null for primitive types ...

const decl = !ModelUtil.isPrimitiveType((shortName) ? modelFile.getModelManager().getType(modelFile.getType(shortName)) : null;
```

PR #742, opened 2023-10-20
feat(*): use modelfile get type

Jonathan-Casey:
### Descrition

Gets the type from the ModelFile, removes the utility function.

### Author Checklist


packages/concerto-core/lib/introspect/mapvaluetype.js
```diff
@@ -69,7 +69,8 @@ class MapValueType extends Decorated {
      */
     validate() {
         if (!ModelUtil.isPrimitiveType(this.type)) {
-            const decl = ModelUtil.getTypeDeclaration(this.ast.type.name, this.modelFile);
+
+            const decl = this.modelFile.getType(this.ast.type.name);
 
             // All declarations, with the exception of MapDeclarations, are valid Values.
```
dselman, line 75: Have we got a test for enums and scalars as values?

dselman (commented):
Also spotted, this:

```
    static isValidMapKeyScalar(decl) {
        return (decl?.isScalarDeclaration?.() &&
        (decl?.ast.$class !== `${MetaModelNamespace}.StringScalar`  ||
        decl?.ast.$class !== `${MetaModelNamespace}.DateTimeScalar`));
    }
```

It's a stylistic / defensive programming thing, but this function will return true if you pass null/undefined or a random object to it, which seems odd given its name and role. 

Generally the check for "is this thing a scalar" should be performed by the caller of this function, and then this function could throw an exception if you give it anything that is not a scalar (exceptional behaviour, because the caller should have read the jsdoc and checked they were passing in the correct args). The function can then return false for valid scalars, but that cannot be used as keys.

The proliferation of the untyped `.?` is to be avoided when possible - they will all have to be removed if/when we move to Typescript, for example, which will enforce type safety.

packages/concerto-core/lib/introspect/mapvaluetype.js
Jonathan-Casey: For Map Key [yes](https://github.com/accordproject/concerto/blob/9ff9b0c48a5d7944103e2882d3a5022d4cdbf464/packages/concerto-core/test/introspect/mapdeclaration.js#L167). Coverage for imported Scalars as Values I've just added..

Jonathan-Casey: We don't support Enums as Map types (yet).

PR #664, opened 2023-07-05
feat(vocabulary) : support multiple terms per concept/property

dselman:
# Closes #645 

### Changes
- <ONE>
- <TWO>

### Flags

- <ONE>
- <TWO>

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist
- [ ] Ensure you provide a [DCO sign-off](https://github.com/probot/dco#how-it-works) for your commits using the `--signoff` option of git commit.
- [ ] Vital features and changes captured in unit and/or integration tests
- [ ] Commits messages follow [AP format](https://github.com/accordproject/techdocs/blob/master/DEVELOPERS.md#commit-message-format)
- [ ] Extend the documentation, if necessary
- [ ] Merging to `main` from `fork:branchname`

PR #754, opened 2023-11-06
feat(core):added extract decorator method

### LINKED ISSUE #753: feat: Extract the vocabulary set and decorator command set from a model manager
# Feature Request 🛍️
Extract the vocabulary set and decorator command set from a model manager

## Possible Solution
Extend the decorator manager to support the extractDecorator method

## Detailed Description
`decoratorManager`shall expose `extractDecorators`  that take two options i.e. whether to remove decorators from source model and locale for vocabs, along with a ModelManager.
It returns:

- modelManager with models without decorators(depeneding on the option passed)
- List of json dcms strings
- List of vocab strings

muskanbararia:
# Closes #753
Extend DecoratorManager to support extraction of vocab and decorators from a modelManager.

### Changes
- Exposes a new `extractDecorator` method from decoratorManager.

### Author Checklist


packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +252,186 @@ class DecoratorManager {
...
+    /**
+    * Adds a key-value pair to a dictionary (object) if the key exists,
+    * or creates a new key with the provided value.
+    *
+    * @param {Object} dictionary - The dictionary (object) to which to add the key-value pair.
+    * @param {string} key - The key to add or update.
+    * @param {any} value - The value to add or update.
+    * @param {string} declaration - The target decl.
+    * @param {string} property The target property.
+    */
```
mttrbrts, line 264:
```suggestion
    * @private
    */
```

I don't believe that this should be publically addressable?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +252,186 @@ class DecoratorManager {
...
+        });
+        return data;
+    }
+    /**
+     * Parses the dict data into an array of decorator jsons
+     * @param {Object} decoratorDict the input dict
+     * @param {String} locale locale for target vocab
+     * @returns {Array<Object>} the parsed decorator command set array
+     */
+    static parseVocabs(decoratorDict,locale){
```
mttrbrts, line 372:
```suggestion
    static parseVocabularies(decoratorDict,locale){
```

The pattern in this repo is to avoid abbreviations in function names

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -474,6 +654,38 @@ class DecoratorManager {
...
+    /**
+    * Process decorators for a given set of decorators.
+    * @param {string} namespace - The namespace to associate with the decorators.
+    * @param {Array<string>} decorators - The array of decorators to process.
+    * @param {Object} decoratorDict - The decorator dictionary to store the decorators.
+    * @param {boolean} shouldRemoveDecorators - A flag to indicate whether decorators hould be removed.
+    * @param {string} decl - declaration on which decorator has been applied
+    * @param {string} property - property on which decorator has been applied
+    */
+    static processDecorators(namespace, decorators, decoratorDict, shouldRemoveDecorators,decl,property) {
```
mttrbrts, line 681: Is there another name that's more descriptive for this function? `process` doesn't tell me anything about what this function does unfortunately.

I also think that this should have a `@private` annotation in the JSDoc (this will hide it from the api.txt file and TypeScript definitions file).

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -474,6 +654,38 @@ class DecoratorManager {
...
         }
     }
+    /**
+     * extracts a Command from a ClassDeclaration or its properties, as required.
+     * @param {String} namespace the namespace for the declaration
+     * @param {*} declaration the class declaration
+     * @param {Object} decoratorDict the dictionary with extracted decorators
+     * @param {boolean} shouldRemoveDecorators whether to remove decorators from base model
+     */
+    static extractDecoratorFromModel(namespace, declaration, decoratorDict, shouldRemoveDecorators) {
```
mttrbrts, line 664: This function appears to have side-effects:
1. To modify the declaration
2. To update the decoratorDict

Is there a way to make this more functional?

packages/concerto-core/test/decoratormanager.js
```diff
@@ -470,4 +470,32 @@ describe('DecoratorManager', () => {
...
+        it('should be able to extract decorators and vocabs from a model', async function() {
+            const testModelManager = new ModelManager({strict:true});
+            const modelText = fs.readFileSync('./test/data/decoratorcommands/test-decorated-model.cto', 'utf-8');
+            testModelManager.addCTOModel(modelText, 'test.cto');
+            const options = {
+                removeDecoratorsFromModel:true,
+                locale:'en'
+            };
+            let resp = DecoratorManager.extractDecorators( testModelManager, options);
+            resp.should.not.be.null;
```
mttrbrts, line 484: Can we be more precise about what the response should be, please?

packages/concerto-core/test/decoratormanager.js
```diff
@@ -470,4 +470,32 @@ describe('DecoratorManager', () => {
...
     });
-});
+
+    describe('#extractDecorators', function() {
+        it('should be able to extract decorators and vocabs from a model', async function() {
+            const testModelManager = new ModelManager({strict:true});
+            const modelText = fs.readFileSync('./test/data/decoratorcommands/test-decorated-model.cto', 'utf-8');
+            testModelManager.addCTOModel(modelText, 'test.cto');
+            const options = {
+                removeDecoratorsFromModel:true,
```
mttrbrts, line 480: Do we also need a variant with this parameter undefined / set to false?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +252,186 @@ class DecoratorManager {
...
+            jsonData.forEach((obj)=>{
+                // check if obj.decl already in dictVoc
+                if (!dictVoc[obj.declaration]){
+                    dictVoc[obj.declaration]={
+                        propertyVocabs:{}
+                    };
+                }
+                let decos=JSON.parse(obj.dcs);
+                decos.forEach((dcs)=>{
+                    if (dcs.name==='Term' || dcs.name==='Term_description'){
```
mttrbrts, line 390: If I understand correctly, this function should handle any decorator that has the prefix `Term_`, right? For example, I should also be able to extract `Term_tooltip` or `Term_myCustomLabel` etc.

packages/concerto-core/test/decoratormanager.js
mttrbrts: You could also confirm that the response is a valid decorator set? Furthermore, I assume that extracting/removing decorators and then reapplying them should be an identity operation?

mttrbrts (commented):
Thanks Muskan, in this pass I've mostly focussed on functional design and naming.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +252,186 @@ class DecoratorManager {
...
+            vocabularies:vocab
+        };
+    }
+
+    /**
+     * Parses the dict data into an array of decorator jsons
+     * @param {Object} decoratorDict the input dict
+     * @returns {Array<Object>} the parsed decorator command set array
+     */
+    static parseDecorators(decoratorDict){
```
ekarademir, line 312: Since this is excluding vocabulary decorators, maybe it's name can be less generic, like `parseNonVocabularyDecorators`?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +252,186 @@ class DecoratorManager {
...
+                            '$class': 'concerto.metamodel@1.0.0.Decorator',
+                            'name': dcs.name,
+                        };
+                        if (dcs.arguments){
+                            let args=dcs.arguments.map((arg)=>{
+                                return {
+                                    '$class':arg.$class,
+                                    'value':arg.value
+                                };
+                            });
```
ekarademir, line 343: nit
```suggestion
                            let args = dcs.arguments.map( arg => {
                                '$class':arg.$class,
                                'value':arg.value
                             });
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +252,186 @@ class DecoratorManager {
...
+        }
+    }
+
+    /**
+     * Extracts all the decorator commands from all the models in modelManager
+     * @param {ModelManager} modelManager the input model manager
+     * @param {object} [options] - decorator models options
+     * @returns {Object} a new model manager with the decorations removed and a list of extracted decorator jsons
+     */
+    static extractDecorators(modelManager,options) {
```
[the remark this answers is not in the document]
muskanbararia: Updated

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,233 @@ class DecoratorManager {
...
+        options = {
+            ...{
+                removeDecoratorsFromModel: false,
+                locale:'en'
+            },
+            ...options || {}
+        };
```
mttrbrts, line 271: Is this equivalent to:
```suggestion
        options = {
            removeDecoratorsFromModel: false,
            locale:'en',
            ...options
        };
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,233 @@ class DecoratorManager {
...
+                        }
+                        return property;
+                    });
+                    decl.properties=processedProperties;
+                }
+                return decl;
+            });
+            model.declarations=processedDecl;
+            return model;
+        });
```
mttrbrts, line 305: Please verify that this provides full coverage to all decorators. See the metamodel for reference, https://github.com/accordproject/concerto-metamodel/blob/main/lib/metamodel.cto

For example, does it also work for MapDeclarations and EnunDeclarations. 

A visitor pattern may be more appropriate here.

packages/concerto-core/test/decoratormanager.js
```diff
@@ -470,4 +470,37 @@ describe('DecoratorManager', () => {
...
+            };
+            let resp = DecoratorManager.extractDecorators( testModelManager, options);
+            let dcs = resp.decoratorCommandSet;
+            let modelManager = resp.modelManager;
+            dcs.forEach((decorator)=>{
+                modelManager=DecoratorManager.decorateModels( modelManager, decorator);
+            });
+            let updatedModels = modelManager.getNamespaces();
+            let testModels = testModelManager.getNamespaces();
+            updatedModels.should.deep.equal(testModels);
```
mttrbrts, line 502: Can we be more precise here? For example, I wouldn't expect the list of namespaces to be modified by your changes. Does testing deep equality between the ASTs work?

packages/concerto-core/test/decoratormanager.js
```diff
@@ -470,4 +470,37 @@ describe('DecoratorManager', () => {
...
+            const options = {
+                removeDecoratorsFromModel:true,
+                locale:'en'
+            };
+            let resp = DecoratorManager.extractDecorators( testModelManager, options);
+            let dcs = resp.decoratorCommandSet;
+            let modelManager = resp.modelManager;
+            dcs.forEach((decorator)=>{
+                modelManager=DecoratorManager.decorateModels( modelManager, decorator);
+            });
```
mttrbrts, line 499: Is there a way to apply a DCS atomically to a model manager?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,233 @@ class DecoratorManager {
...
+        let processedAST={
+            ...decoratedAst,
+            models:processedModels,
+        };
+        const newModelManager = new ModelManager();
+        newModelManager.fromAst(processedAST);
+        let dcms=(this.parseNonVocabularyDecorators(extractionDictionary));
+        let vocab = this.parseVocabularies(extractionDictionary,options.locale);
```
mttrbrts, line 313: Nit. I'm surprised that the linter didn't pick this up, but we have a preference for `const` declarations unless neccessary in this repo. Also, the whitespace is inconsistent around the assignment.

packages/concerto-core/lib/decoratormanager.js
ekarademir: If the options is not provided, then it would fail the unpacking.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,233 @@ class DecoratorManager {
         return newModelManager;
     }
 
+    /**
+     * Extracts all the decorator commands from all the models in modelManager
+     * @param {ModelManager} modelManager the input model manager
+     * @param {object} options - decorator models options
+     * @param {boolean} options.removeDecoratorsFromModel - flag to strip out decorators from models
+     * @param {string} options.locale - locale for extracted vocabulary set
+     * @returns {Object} a new model manager with the decorations removed and a list of extracted decorator jsons
```
ekarademir, line 262: If it's returning a new `ModelManager`, then the return type is `ModelManager` is it not?

packages/concerto-core/lib/decoratormanager.js
ekarademir: My bad, it doesn't fail.

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: @muskanbararia I agree with @mttrbrts 
For more reference: https://github.com/accordproject/concerto/blob/668a033d0f4df7aeddeaf42afc25b2588920148f/packages/concerto-vocabulary/lib/vocabularymanager.js#L326

https://github.com/accordproject/concerto/pull/664

packages/concerto-core/lib/decoratormanager.js
muskanbararia: It is returning an Object of type:
```
{
            modelManager:newModelManager,
            decoratorCommandSet,
            vocabularies
}
```

packages/concerto-core/lib/decoratormanager.js
muskanbararia: Updated the PR. I was incorrectly assuming that only `Term_description` is identifier for vocab

packages/concerto-core/lib/decoratormanager.js
muskanbararia: Visitor Pattern may not work best here because I am modifying the ast for simplicity while removing decorators from source. 
The same approach (iterating on ast) was taken in applying the decorators as well. ( I am guessing it's because the APIS are read-only and we need to modify the ast to modify the model )
I have ensured it supports all types of declarations and corresponding props, including maps, scalars and enums. Have added a relevant test case as well.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,252 @@ class DecoratorManager {
...
+     */
+    static extractDecorators(modelManager,options) {
+        options = {
+            removeDecoratorsFromModel: false,
+            locale:'en',
+            ...options
+        };
+        const ast = modelManager.getAst(true);
+        const decoratedAst = JSON.parse(JSON.stringify(ast));
```
Jonathan-Casey, line 271: Why is this necessary?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,252 @@ class DecoratorManager {
...
+        let extractionDictionary={};
+        const processedModels = decoratedAst.models.map((model)=>{
+            if (model.decorators && model.decorators.length>0){
```
Jonathan-Casey, line 274: nit: formatting
```suggestion
            if (model.decorators && model.decorators.length > 0){
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,252 @@ class DecoratorManager {
...
+     */
+    static extractDecorators(modelManager,options) {
+        options = {
+            removeDecoratorsFromModel: false,
+            locale:'en',
+            ...options
+        };
+        const ast = modelManager.getAst(true);
+        const decoratedAst = JSON.parse(JSON.stringify(ast));
+        let extractionDictionary={};
```
Jonathan-Casey, line 272:
```suggestion
        let extractionDictionary = {};
```

Jonathan-Casey: hey @muskanbararia can you please run a linter to fix the formatting niggles? Thanks!

Jonathan-Casey (changes requested):
hey @muskanbararia can you please run a linter to fix the formatting niggles? Thanks!

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,252 @@ class DecoratorManager {
...
+                    if (decl.key){
+                        if (decl.key.decorators){
+                            extractionDictionary=this.constructDCSDictionary(extractionDictionary,model.namespace,decl.key.decorators,decl.name,'','KEY');
+                            decl.key.decorators=null;
+                        }
+                    }
+                    if (decl.value){
+                        if (decl.value.decorators){
+                            extractionDictionary=this.constructDCSDictionary(extractionDictionary,model.namespace,decl.value.decorators,decl.name,'','VALUE');
+                            decl.value.decorators=null;
```
Jonathan-Casey, line 297: `undefined` is better aligned to JS best practices. Does it make sense to delete the decorators property entirely?

packages/concerto-core/lib/decoratormanager.js
muskanbararia: @Jonathan-Casey this reports no errors when I run `npm run lint`. do you want me to add new rules to eslintrc?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,251 @@ class DecoratorManager {
...
+    * @param {Object} dictionary - The dictionary (object) to which to add the key-value pair.
+    * @param {string} key - The key to add or update.
+    * @param {any} value - The value to add or update.
+    * @param {string} declaration - The target decl.
+    * @param {string} property - The target property.
+    * @param {string} mapElement - The target mapElement.
+    * @returns {Object} - constructed DCS Dict
+    * @private
+    */
+    static constructDCSDictionary(dictionary, key, value, declaration,property,mapElement='') {
```
ekarademir, line 486: I still think the last three can be part of an options object.

ekarademir: Maybe you can extract out all these new methods to a separate class, like `DecoratorExtractor`, and not export it, so that it becomes an internal tool and wouldn't add bulk to the `DecoratorManager`. Then you can inject managers as a dependencies.

muskanbararia: Shall I create Utils instead? I can even move some of the private methods used by `decorateModels` there.

ekarademir: Either way works I think. `DecoratorUtil` like we have for models, `ModelUtil` would also be a good extraction.

muskanbararia: Updated. Only moved the methods related to extraction for now.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,197 @@
...
+    * @param {any} value - The value to add or update.
+    * @param {Object} options - The target declaration, property and mapElement.
+    * @returns {Object} - constructed DCS Dict
+    * @private
+    */
+    static constructDCSDictionary(dictionary, key, value, options) {
+        const val = {
+            declaration:options.declaration || '',
+            property:options.property || '',
+            mapElement:options.mapElement || '',
```
ekarademir, line 186:
```suggestion
            declaration:options?.declaration || '',
            property:options?.property || '',
            mapElement:options?.mapElement || '',
```

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,197 @@
...
+        return data;
+    }
+    /**
+    * Adds a key-value pair to a dictionary (object) if the key exists,
+    * or creates a new key with the provided value.
+    *
+    * @param {Object} dictionary - The dictionary (object) to which to add the key-value pair.
+    * @param {string} key - The key to add or update.
+    * @param {any} value - The value to add or update.
+    * @param {Object} options - The target declaration, property and mapElement.
```
ekarademir, line 178:
```suggestion
    * @param {Object} options
    * @param {string} options.declaration - Target declaration
    * @param {string} options.property - Target property
    * @param {string} options.mapElement - Target map element
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +253,103 @@ class DecoratorManager {
...
+                        }
+                        return property;
+                    });
+                    decl.properties = processedProperties;
+                }
+                return decl;
+            });
+            model.declarations = processedDecl;
+            return model;
+        });
```
ekarademir, line 338: I think this can be separated out into another method for `DecoratorManager` and be called, `collectDecorators`.

Then `extractDecorators` would simply be collecting decorators and parsing them.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -252,6 +253,103 @@ class DecoratorManager {
...
+        const processedAST={
+            ...decoratedAst,
+            models:processedModels,
+        };
+        const newModelManager = new ModelManager();
+        newModelManager.fromAst(processedAST);
+        const decoratorCommandSet = DecoratorUtil.parseNonVocabularyDecorators(extractionDictionary, DCS_VERSION);
+        const vocabularies = DecoratorUtil.parseVocabularies(extractionDictionary, options.locale);
```
ekarademir, line 346: You can move the loop up from `parseNonVocabularyDecorators` and `parseVocabularies`, and refactor those functions to expect just the `namespace` and `jsonData`, then call appropriate one based on the conditions.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,197 @@
...
+     * @param {String} locale locale for target vocab
+     * @returns {Array<Object>} the parsed decorator command set array
+     * @private
+     */
+    static parseVocabularies(decoratorDict,locale){
+        const data = [];
+        const patternToDetermineVocab = /^Term_/i;
+        Object.keys(decoratorDict).forEach((namespace)=>{
+            if (JSON.stringify(decoratorDict[namespace]).includes('Term') ||
+            JSON.stringify(decoratorDict[namespace]).includes('Term_')){
```
ekarademir, line 103: Why call `JSON.stringify` two separate times? I'm pretty sure it's a relatively expensive operation. You can form the string before the if statement.

ekarademir: Also if this is needed because `Term` is deep in the data structure, you can consider flattening the structure and looping over those decorators?

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,197 @@
...
+                strVoc = strVoc+`namespace: ${namespace}\n`;
+                strVoc = strVoc+'declarations:\n';
+                const jsonData = decoratorDict[namespace];
+                const dictVoc = {};
+                jsonData.forEach(obj =>{
+                    if (!dictVoc[obj.declaration]){
+                        dictVoc[obj.declaration] = {
+                            propertyVocabs: {}
+                        };
+                    }
```
ekarademir, line 115:
```suggestion
                    dictVoc[obj.declaration] = dictVoc[obj.declaration] || { propertyVocabs: {} };
```

packages/concerto-core/lib/decoratorutil.js
muskanbararia: Figured a way to do without the stringify check

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,301 @@
...
+     * @param {Object} jsonData - the model ast for a particular namespace
+     * @param {String} namespace - the namespace for which parsing is to be done
+     * @param {String} DCS_VERSION - the version string
+     * @returns {Array<Object>} - the parsed decorator command set array
+     * @private
+     */
+    static parseNonVocabularyDecorators(jsonData, namespace, DCS_VERSION){
+        const patternToDetermineVocab = /^Term_/i;
+        const dcsObjects = [];
+        jsonData.forEach(obj =>{
```
ekarademir, line 37: By running this loop here, and also in `parseVocabularies`, you are running this loop twice, instead you can pull that into `parseDecosAndVocabs`, you can be done with one loop. You can call either `parseNonVocabularyDecorators ` or `parseVocabularies` based on the conditions you decide to process. At each loop, both functions use `jsonData` just to run the loop.

If you run the loop at `parseDecosAndVocabs`, you can also pull the loop you run on `decos` in `parseDecosAndVocabs` and run the decision making process there, i.e., you can run the `forEach` loop on `jsonData`, then for each `obj` you can get `decos` and run the `forEach` loop on that as well, then finally, based on `dcs.name === 'Term' || patternToDetermineVocab.test(dcs.name)` condition, you can just pass `obj` and `decos` and other things needed to `parseVocabularies` and `parseNonVocabularyDecorators` functions. This would simplify these functions a lot.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,301 @@
...
+                            property.decorators = undefined;
+                        }
+                        return property;
+                    });
+                    decl.properties = processedProperties;
+                }
+                return decl;
+            });
+            model.declarations = processedDecl;
+            return model;
```
ekarademir, line 285: What I meant before, was that this huge chunk is collection and can be separated out to a different function. Having everything in this function, simplified `extractDecorators` but now this huge chunk is under this function.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,371 @@
...
+                };
+                if (obj.declaration && obj.declaration!==''){
+                    target.declaration = obj.declaration;
+                }
+                if (obj.property && obj.property!==''){
+                    target.property = obj.property;
+                }
+                if (obj.mapElement && obj.mapElement!==''){
+                    target.mapElement = obj.mapElement;
+                }
```
ekarademir, line 152: Much cleaner now, but since `target` is only needed within `parseNonVocabularyDecorators`, and never used else where down the line, in `parseNonVocabularyDecorators `, you can build `target` object within `parseNonVocabularyDecorators` and you can pass `DCS_VERSION` and `namespace` as options to `parseNonVocabularyDecorators`. That would make this loop a lot cleaner and contain the logic related to `parseNonVocabularyDecorators` in one place.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,371 @@
...
+            });
+            if (dcsObjects?.length > 0){
+                const dcmsForNamespace = {
+                    '$class': `org.accordproject.decoratorcommands@${DCS_VERSION}.DecoratorCommandSet`,
+                    'name': nameOfDcs,
+                    'version': versionOfDcs,
+                    'commands': dcsObjects
+                };
+                decoratorData.push(dcmsForNamespace);
+            }
```
ekarademir, line 170: you can now create another function that "transform"s `dcsObjects`, maybe called `transformNonVoabularyDecorators`. I know this is a simple transformation, but it would help to contain transformations in one place. Then you can return the `dcmsForNamespace` and push that to `decoratorData`.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,371 @@
...
+                        }
+                        strVoc += '    properties:\n';
+                        Object.keys(vocabObject[decl].propertyVocabs).forEach(prop =>{
+                            strVoc += `      - ${prop}: ${vocabObject[decl].propertyVocabs[prop].term}\n`;
+                            const otherProps = Object.keys(vocabObject[decl].propertyVocabs[prop]).filter((str)=>str!=='term');
+                            otherProps.forEach(key =>{
+                                strVoc += `        ${key}: ${vocabObject[decl].propertyVocabs[prop][key]}\n`;
+                            });
+                        });
+                    }
```
ekarademir, line 196: The same goes for this transformation.

packages/concerto-core/lib/decoratorutil.js
muskanbararia: Since the `parseNonVocabularyDecorators` is called per declaration level, a new target will be getting created for each declaration. however, currently target declaration is outside the declaration loop. I think it is more performant currently.
Do you think I should still move it to `parseNonVocabularyDecorators` ?

ekarademir: Ah, in that case, it doesn't make sense to move inside the function, but maybe move it inside a separate function called, `buildDecoratorTarget` or something? Here it just crowds the rest of the function.

muskanbararia: updated.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,409 @@
...
+    /**
+     * Transforms the collected decorators into proper decorator command sets
+     * @param {Array<Object>} dcsObjects - the collection of collected decorators
+     * @param {string} DCS_VERSION - the version string
+     * @param {string} namespace - the current namespace
+     * @param {Array<Object>} decoratorData - the collection of existing decorator command sets
+     * @returns {Array<Object>} - the collection of decorator command sets
+     * @private
+     */
+    static transformNonVoabularyDecorators(dcsObjects, DCS_VERSION, namespace, decoratorData){
```
ekarademir, line 127: typo
```suggestion
    static transformNonVocabularyDecorators(dcsObjects, DCS_VERSION, namespace, decoratorData){
```

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,409 @@
...
+    /**
+     * Transforms the collected vocabularies into proper vocabulary command sets
+     * @param {Array<Object>} vocabObject - the collection of collected vocabularies
+     * @param {string} locale - the locale of vocabulary
+     * @param {string} namespace - the current namespace
+     * @param {Array<Object>} vocabData - the collection of existing vocabularies command sets
+     * @returns {Array<Object>} - the collection of vocabularies command sets
+     * @private
+     */
+    static transformVoabularyDecorators(vocabObject, locale, namespace, vocabData){
```
ekarademir, line 151: typo
```suggestion
    static transformVocabularyDecorators(vocabObject, locale, namespace, vocabData){
```

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,409 @@
...
+                        const otherProps = Object.keys(vocabObject[decl].propertyVocabs[prop]).filter((str)=>str!=='term');
+                        otherProps.forEach(key =>{
+                            strVoc += `        ${key}: ${vocabObject[decl].propertyVocabs[prop][key]}\n`;
+                        });
+                    });
+                }
+            });
+            vocabData.push(strVoc);
+        }
+        return vocabData;
```
ekarademir, line 181: If you are returning `vocabData`, then you don't need to accept it as an argument. If you are mutating as a side-effect, you don't need to return it.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,409 @@
...
+        if (dcsObjects?.length > 0){
+            const dcmsForNamespace = {
+                '$class': `org.accordproject.decoratorcommands@${DCS_VERSION}.DecoratorCommandSet`,
+                'name': nameOfDcs,
+                'version': versionOfDcs,
+                'commands': dcsObjects
+            };
+            decoratorData.push(dcmsForNamespace);
+        }
+        return decoratorData;
```
ekarademir, line 140: The comment for `vocabData` applies to this `decoratorData`.

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,409 @@
...
+    /**
+    * parses the extracted decorators and generates arrays of decorator command set and vocabularies
+    *
+    * @param {Object} extractionDictionary - extracted decorators and vocabularies
+    * @param {String} DCS_VERSION - dcs_version
+    * @param {String} locale - locale for vocabularies
+    * @returns {Object} - constructed DCS Dict and processed models ast
+    * @private
+    */
+    static parseDecosAndVocabs(extractionDictionary, DCS_VERSION, locale){
```
ekarademir, line 216: No need to save few characters here.
```suggestion
    static parseDecoratorsAndVocabularies(extractionDictionary, DCS_VERSION, locale){
```

ekarademir: also this could be called transform as well, since parse means something different within the concerto library.
```suggestion
    static transformExtractionDictionary(extractionDictionary, DCS_VERSION, locale){
```

packages/concerto-core/lib/decoratorutil.js
```diff
@@ -0,0 +1,409 @@
...
+            }
+            if (decl.$class.endsWith('.MapDeclaration')) {
+                const processedMapDeclData = this.processMapDeclaration(decl, extractionDictionary, removeDecoratorsFromModel, namespace);
+                decl = processedMapDeclData.declaration;
+                extractionDictionary=processedMapDeclData.extractionDictionary;
+            }
+            if (decl.properties) {
+                const processedPropertiesData = this.processProperties(decl.properties, decl.name, extractionDictionary, removeDecoratorsFromModel, namespace);
+                decl.properties = processedPropertiesData.processedProperties;
+                extractionDictionary = processedPropertiesData.extractionDictionary;
```
ekarademir, line 344: These types of providing arguments and then reading the same arguments from the returned data begs for contained state. I would either create a class and keep the state as a private property, the state in this case is `extractionDictionary` and `properties`(maybe), or, I would keep these in a `TransformationContext` object, and pass that object around, instead of calling functions and reassigning them.

ekarademir: I didn't realise that this whole thing was a class, I think you can convert everything into a `DecoratorTransformer` or `DecoratorExtractor` class and accept all the things you accept in `collectExtractedDecoratorsAndProcessedModels` in the constructor, then keep `exctractionDictionary` as an internal state. Once constructed, you can move the logic in `collectExtractedDecoratorsAndProcessedModels` into a method called, `transform` or `extract`.

That way you don't need to pass it around and you can make the class public, since it will work with any model manager, and you can make all these functions that are not being called elsewhere private functions.

ekarademir (approved):

packages/concerto-core/lib/decoratormanager.js
muskanbararia: updated in latest commits

packages/concerto-core/lib/decoratormanager.js
muskanbararia: updated in latest commits.

packages/concerto-core/test/decoratormanager.js
muskanbararia: updated the test case to check for identity post reapplication of same dcs and vocab

packages/concerto-core/test/decoratormanager.js
muskanbararia: here, dcs is an array of decorator command set. I think we will have to iterate over each dcms and apply? Please suggest if there's an alternative.

Jonathan-Casey (approved):

mttrbrts (approved):

Issue #772, opened 2023-12-15
Errors when creating validators in concerto introspect only return a string in a error message , which makes programatic handling the errors dificult


santanu8961:
# Bug Report 🐛

While adding AST models to a `ModelManager` and validating a `string` if there is any error the error only consists of an error message because it gets challenging to detect what kind of error it is, and based on that doing any operation
## Expected Behavior

The error structure should have a `type` to it which can tell the service consuming it what kind of error it is.
## Current Behavior

right now in `concerto/packages/concerto-core/lib/introspect/validator.js` `reportError` only reports an error message, not a type
## Possible Solution

implementing a type with an error message altogether passing a proper error object can fix this behavior.
## Steps to Reproduce

```

const modelManager = new ModelManager({ strict: true, regExp: RE2 });

const modelFiles = [m].map(item => new ModelFile(modelManager, item.metamodel));
modelFiles.sort((a, b) => a.getNamespace().localeCompare(b.getNamespace()));
modelManager.addModelFiles(modelFiles);
```

In this code block I am using a custom regex validator re2 which I am passing to `ModelManager` now if the metamodel has a regex validator that does not satisfy the compliance and throws an error its just a message with information but no type is there, so its difficult to classify which type of error is that and doing some operation like throwing a bad request error is difficult.

### Desktop
 - OS: Windows with WSL
 - Browser: Chrome

PR #773, opened 2023-12-15
fix(error): adding type to error in string validator in introspect

santanu8961:
# Closes <https://github.com/accordproject/concerto/issues/772>

### Changes

Adds type to validator error in introspect validator.

- added a new optional parameter to class `BaseException`  which can take `errorType` as well as `errorMessage` which helps to determine what kind of error is getting thrown from the validator itself. and using that to throw error in `validator`.
- also added this in `NotFoundException` for better complience

### Author Checklist


packages/concerto-core/lib/introspect/stringvalidatorexception.js
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+/**
+* Class representing an validation error related to anykind of string validatoin in introspect
```
sanketshevkar, line 18: nit: *validatoin

packages/concerto-core/lib/introspect/stringvalidatorexception.js
```diff
@@ -0,0 +1,38 @@
...
+* @class
+* @memberof module:concerto-core
+*/
+class StringValidationException extends Error {
+    /**
+     * Create error related to string validation.
+     * @param {string} errormessage - The exception message.
+     * @param {string} errorType - conditional parameter which denoted the kind of string
```
sanketshevkar, line 26: nit: *denoted -> denotes/represents

sanketshevkar:
Should we have a new folder for error classes and not keep it in introspect?

packages/concerto-core/lib/introspect/stringvalidatorexception.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+'use strict';
+
+/**
+* Class representing an validation error related to anykind of string validatoin in introspect
+* @class
+* @memberof module:concerto-core
+*/
+class StringValidationException extends Error {
```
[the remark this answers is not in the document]
mttrbrts: It should be Validator, not Validation. Internally the `regex` keyword is called a `StringValidator`.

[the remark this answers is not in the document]
dselman: Yes, we should ensure we can use similar approach with NumberValidator as well.

packages/concerto-core/lib/introspect/validator.js
```diff
@@ -46,10 +48,11 @@ class Validator {
     /**
      * @param {string} id the identifier of the instance
      * @param {string} msg the exception message
+     * @param {string} errorType the type of error
      * @throws {Error} throws an error to report the message
      */
-    reportError(id, msg) {
-        throw new Error( 'Validator error for field `' + id + '`. ' + this.getFieldOrScalarDeclaration().getFullyQualifiedName() + ': ' + msg );
+    reportError(id, msg, errorType='ValidatorError') {
```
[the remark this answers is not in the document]
dselman: Let's avoid having to use `instanceof` checks - those make assumptions about how packages are required (instanceof can fail if the same code is loaded at different version) which have bitten us in the past.

dselman (changes requested):
I like the motivation for this. I just needs a bit of refactoring to ensure it works with NumberValidator and can be cleanly consumed upstream.

packages/concerto-core/lib/introspect/validator.js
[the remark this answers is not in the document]
santanu8961: Hi @ekarademir I have used the additional argument `errorType` so that it can be propagated through till `StringValidationException` where we can use it to denote what kind of error it is , which can be very useful to do any kind of operation as only error string is quick tricky to detect what error it is trying to throw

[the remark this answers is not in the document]
santanu8961: Hi @dselman , can you please clarify what do you mean by the instance of checks?
Thanks.

packages/concerto-core/lib/introspect/stringvalidatorexception.js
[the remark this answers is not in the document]
santanu8961: Hi @mttrbrts , @ekarademir  I am keeping the name of the class as `StringValidatiorException` and the default error as `ValidatorError` and the regex error as `RegexValidatorException`.
let me know if this works fine . and this is what you meant.

mttrbrts (approved):

packages/concerto-core/lib/introspect/validator.js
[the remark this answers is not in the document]
dselman: I mean client code doing things like:

```
try {
...
}
catch(err) {
if(err instanceof StringValidatorException) {
 ...
}
}
```

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -70,7 +70,7 @@ class StringValidator extends Validator{
                 this.regex = new CustomRegExp(validator.pattern, validator.flags);
             }
             catch(exception) {
-                this.reportError(field.getName(), exception.message);
+                this.reportError(field.getName(), exception.message,'RegexValidatorException');
```
dselman, line 73: I think if we want to go this route then the error codes should be defined and exported and referenced.

E.g. something like:

```
ErrorCodes.js
export const INVALID_REGEX = 1;
```

Somewhere we want to report an error:
```
this.reportError(field.getName(), exception.message, ErrorCodes.INVALID_REGEX);
```

This allows a client to do something like:

```
import {ErrorCodes} from '@accordproject/concerto-core';

try {
...
}
catch(err) {
const errorCode = err?.getErrorCode.?();
switch(errorCode) {
case ErrorCodes.INVALID_REGEX:
  // do something specific here
break;
default:
  // general case
}
```

We should have in mind that concerto-core will be moving to TypeScript and we want these error codes to be type safe.

packages/concerto-core/lib/introspect/stringvalidatorexception.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+'use strict';
+
+/**
+* Class representing an validation error related to anykind of string validation in introspect
+* @class
+* @memberof module:concerto-core
+*/
+class StringValidatorException extends Error {
```
dselman, line 22: This could be more generic - it extends Error and adds an Error Code. We could add an optional error code to concerto-util/BaseException, so that all exceptions can reference an error code.

dselman (changes requested):
I would prefer to see generic Error Code functionality moved into concerto-util/BaseException, and to refactor code to throw Errors that derive from BaseException.

Error codes should be defined constants so that code is type safe.

packages/concerto-core/lib/introspect/stringvalidator.js
```diff
@@ -70,7 +71,7 @@ class StringValidator extends Validator{
                 this.regex = new CustomRegExp(validator.pattern, validator.flags);
             }
             catch(exception) {
-                this.reportError(field.getName(), exception.message);
+                this.reportError(field.getName(), exception.message,ErrorCodes.REGEX_VALIDATOR_EXCEPTION);
```
santanu8961, line 74: Here I have used error codes as constants instead of inline strings for consistency

packages/concerto-core/lib/introspect/validator.js
```diff
@@ -46,10 +47,11 @@ class Validator {
     /**
      * @param {string} id the identifier of the instance
      * @param {string} msg the exception message
+     * @param {string} errorType the type of error
      * @throws {Error} throws an error to report the message
      */
-    reportError(id, msg) {
-        throw new Error( 'Validator error for field `' + id + '`. ' + this.getFieldOrScalarDeclaration().getFullyQualifiedName() + ': ' + msg );
+    reportError(id, msg, errorType=ErrorCodes.DEFAULT_VALIDATOR_EXCEPTION) {
```
santanu8961, line 53: Here I have used error codes as constants instead of inline strings for consistency

packages/concerto-core/lib/introspect/validator.js
```diff
@@ -46,10 +47,11 @@ class Validator {
...
+        throw new BaseException('Validator error for field `' + id + '`. ' + this.getFieldOrScalarDeclaration().getFullyQualifiedName() + ': ' + msg ,errorType);
```
santanu8961, line 54: removed StringValidatorExceptionClass as its not needed anymore I have added a optional param errorType in BaseException which solves the purpose

packages/concerto-util/lib/errorcodes.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
```
santanu8961, line 16: error codes are defined in this file which can be used to implement different errors when ever needed

packages/concerto-util/lib/baseexception.js
```diff
@@ -27,12 +30,14 @@ class BaseException extends Error {
      * Create the BaseException.
      * @param {string} message - The exception message.
      * @param {string} component - The optional component which throws this error.
+     * @param {string} errorType - The optional error code regarding the error
      */
-    constructor(message, component) {
+    constructor(message, component, errorType) {
```
santanu8961, line 35: added a errortype parameter which is optional and provides the errorcode related to the error

packages/concerto-core/lib/introspect/stringvalidator.js
santanu8961: Hi @dselman , as you requested I have changed the structure by doing the following changes to the structure:
1. I no longer am using the StringValidatorException class instead I have added the errorType as an optional param in BaseException and using that.
2. added an errorCodes.js file where I am storing the error type constants and using them directly , these are exposed so the client will also be able to use these , as discussed in the PR

santanu8961:
Hi @dselman , as you requested I have changed the structure by doing the following changes to the structure:
1. I no longer am using the StringValidatorException class instead I have added the errorType as an optional param in BaseException and using that.
2. added an errorCodes.js file where I am storing the error type constants and using them directly , these are exposed so the client will also be able to use these , as discussed in the PR .

requesting your approval on this PR

packages/concerto-core/lib/introspect/stringvalidator.js
sanketshevkar, line 74: nit: lint check in pipeline could throw error here <spacing issue>

packages/concerto-core/types/lib/typenotfoundexception.d.ts
```diff
@@ -7,14 +7,6 @@ export = TypeNotFoundException;
...
  */
 declare class TypeNotFoundException extends BaseException {
-    /**
-     * Constructor. If the optional 'message' argument is not supplied, it will be set to a default value that
-     * includes the type name.
-     * @param {string} typeName - fully qualified type name.
-     * @param {string|undefined} message - error message.
-     * @param {string} component - the optional component which throws this error
-     */
-    constructor(typeName: string, message: string | undefined, component: string);
```
sanketshevkar, line 17: On the main I still see constructor on `TypeNotFoundException.js` but why is this getting deleted here? Is this auto generated?

packages/concerto-core/lib/introspect/validator.js
```diff
@@ -46,10 +47,11 @@ class Validator {
     /**
      * @param {string} id the identifier of the instance
      * @param {string} msg the exception message
+     * @param {string} errorType the type of error
      * @throws {Error} throws an error to report the message
      */
-    reportError(id, msg) {
-        throw new Error( 'Validator error for field `' + id + '`. ' + this.getFieldOrScalarDeclaration().getFullyQualifiedName() + ': ' + msg );
+    reportError(id, msg, errorType=ErrorCodes.DEFAULT_VALIDATOR_EXCEPTION) {
+        throw new BaseException('Validator error for field `' + id + '`. ' + this.getFieldOrScalarDeclaration().getFullyQualifiedName() + ': ' + msg ,undefined,errorType);
```
sanketshevkar, line 54: nit: Space linting, same as above.

packages/concerto-core/types/lib/typenotfoundexception.d.ts
santanu8961: Yes this is actually auto generated after I run build command

packages/concerto-core/lib/introspect/validator.js
santanu8961: resolved

packages/concerto-core/lib/introspect/stringvalidator.js
santanu8961: resolved

sanketshevkar (approved):

dselman (approved):
Good work. Thanks!

PR #794, opened 2024-01-29
refactor(declarations): move declaration uniqueness check to model file

### LINKED ISSUE #785: Declaration validation uniqueness check is O(n^2)
# Discussion 🗣

While validating the concept declarations within the model file, a uniqueness check [runs for each declaration](https://github.com/accordproject/concerto/blob/ced1b3e052de55f9da4d6ab797c864894e41ef04/packages/concerto-core/lib/introspect/classdeclaration.js#L193-L206). This might be acceptable for small model files, but the final logic inadvertently becomes O(n^2). This could be improved to O(n) if we do the uniqueness check where the uniqueness makes more sense, i.e. model file.

[There is already a similar logic](https://github.com/accordproject/concerto/blob/ced1b3e052de55f9da4d6ab797c864894e41ef04/packages/concerto-core/lib/introspect/modelfile.js#L295-L314) for resolving types of properties, at model level.

ekarademir:
# Closes #785 

This logic doesn't need to run for each declaration in the class. It basically calls the model file and fetches declarations and creates a set. There is nothing special about the particular declaration within this logic.

### Changes

- Moved unique name check logic to model file
- Updated the tests (move the tests from class declaration to model file as well)

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/classdeclaration.js
+++ b/packages/concerto-core/lib/introspect/classdeclaration.js
@@ -190,21 +190,6 @@ class ClassDeclaration extends Declaration {
     validate() {
         super.validate();
 
-        const declarations = this.getModelFile().getAllDeclarations();
-        const declarationNames = declarations.map(
-            d => d.getFullyQualifiedName()
-        );
-        const uniqueNames = new Set(declarationNames);
-
-        if (uniqueNames.size !== declarations.length) {
-            const duplicateElements = declarationNames.filter(
-                (item, index) => declarationNames.indexOf(item) !== index
-            );
-            throw new IllegalModelException(
-                `Duplicate class name ${duplicateElements[0]}`
-            );
-        }
-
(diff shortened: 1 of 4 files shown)


packages/concerto-core/lib/introspect/classdeclaration.js
```diff
@@ -190,21 +190,6 @@ class ClassDeclaration extends Declaration {
     validate() {
         super.validate();
 
-        const declarations = this.getModelFile().getAllDeclarations();
```
ekarademir, line 193: There is nothing in this removed code that references the particular declaration (`this`,) and it already calls the model file. So we can just move this logic to model file and remove unnecessary O(n^2) loop.

jamieshorten (approved):

stefanblaginov (approved):

mttrbrts (approved):

Issue #798, opened 2024-02-01
Throw error on validation of model with decorators with type reference if type not present


sanketshevkar:
# Bug Report 🐛

Currently if we add decorator with type reference, it does not check if the type exists within the namespace. We should be throwing error if the type is not imported or not present within the namespace
```
namespace car@1.0.0

import workshop@1.0.0.{Color}

concept Owner {
   o String ownerName
}

// should throw error on model validation as Train is not present in namespace
@colorOptions(Train)
concept CarColor {}
```
Might have to fix this issue first https://github.com/accordproject/concerto/issues/797
## Expected Behavior

Should throw error on validation

## Steps to Reproduce

1. Add the above model to model manager and validate, will not throw error
2. 
3.
4.

subhajit20:
Hey @sanketshevkar  is this the correct file to be changed?
https://github.com/accordproject/concerto/blob/ed6c188d4a03dbc5252a691b51087ad050ee1705/packages/concerto-core/lib/introspect/decorator.js#L33-L45

sanketshevkar:
> Hey @sanketshevkar is this the correct file to be changed?

Yes!

ahmad-kashkoush:
Hey @sanketshevkar , I've Created a PR of My changes, I hope to get a Feedback

dselman:
Fixed in main.

Issue #797, opened 2024-02-01
Need a method to resolve type inference for decorators with type references


sanketshevkar:
# Feature Request 🛍️

In the AST both superType and decorator TypeIdentifiers have a similar shape.
```
"arguments": [
                  {
                    "$class": "concerto.metamodel@1.0.0.DecoratorTypeReference",
                    "type": {
                      "$class": "concerto.metamodel@1.0.0.TypeIdentifier",
                      "name": "Color"
                    }, .....
```

```
"superType": {
              "$class": "concerto.metamodel@1.0.0.TypeIdentifier",
              "name": "Color"
            }
```
But for super class when I do
`declaration.getSuperClass()` I can get a resolved fully qualified type `workshop@1.0.0.Color`

But for reference type decorators I only get `[{ type: 'Identifier', name: 'Color', array: false }]` when I do `decorator.getArguments()` , which I don't think is incorrect.

But shouldn't the decorator class have some internal methods that can resolve the type reference, similar to the one we have in [classDeclaration ](https://github.com/accordproject/concerto/blob/0d7f108d4961f87782436dd689639ce0f35e97f8/packages/concerto-core/lib/introspect/classdeclaration.js#L155)?

RINO-GAELICO:
Hi @sanketshevkar it seems that this issue involves adding a helper method inside the decorator class (packages/concerto-core/lib/introspect/decorator.js) that resolves arguments that are neither Boolean, Numbers nor String to a fully qualified type.
Is my interpretation correct? 
May I be assigned to this issue?

sanketshevkar:
Yes you are correct @RINO-GAELICO. I've assigned this to you!
Thanks for showing interest in our project :)

Also design your solution considering #798 as an extension to this issue.

sanketshevkar:
Also instead of `master` please point your PR towards the https://github.com/accordproject/concerto/tree/v4.0.0 branch.

RINO-GAELICO:
Hi @sanketshevkar,

I've reviewed the documentation and attempted to gain a better understanding of the overall architecture. Could you please provide guidance on the following points?

1. I am considering calling the to-be-implemented helper method within decorator.getArguments(). Does this approach align with the requirements? : 

`if (thing.$class === `${MetaModelNamespace}.DecoratorTypeReference`) {

                        this.arguments.push({
                            **type: this._resolveType(thing),**
                            name: thing.type.name,
                            array: thing.isArray
                        });
                    }`
                    

[...]

PR #813, opened 2024-03-05
Ahmad kashkoush/i798/throw exception

ahmad-kashkoush:

add a throw Error when type is not available in namespace
### Changes

- <ONE>
- <TWO>

### Flags

- <ONE>
- <TWO>

### Related Issues
- Issue #798 

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/decorator.js
+++ b/packages/concerto-core/lib/introspect/decorator.js
@@ -79,6 +79,10 @@ class Decorator {
                 if (thing) {
                     if (thing.$class === `${MetaModelNamespace}.DecoratorTypeReference`) {
                         // XXX Is this really what we want?
+                        if (!MetaModelNamespace[thing.type.name]) {
+                            const errorMessage = `Type '${thing.type.name}' not found within the namespace '${MetaModelNamespace}'.`;
+                            throw new IllegalModelException(errorMessage);
+                        }
                         this.arguments.push({
                             type: 'Identifier',
                             name: thing.type.name,
--- a/packages/concerto-core/lib/introspect/decorator.js
+++ b/packages/concerto-core/lib/introspect/decorator.js
@@ -15,6 +15,7 @@
 'use strict';
 
 const { MetaModelNamespace } = require('@accordproject/concerto-metamodel');
+const IllegalModelException = require('./illegalmodelexception');
 
 // Types needed for TypeScript generation.
 /* eslint-disable no-unused-vars */
--- a/packages/concerto-core/lib/introspect/decorator.js
+++ b/packages/concerto-core/lib/introspect/decorator.js
@@ -79,16 +79,15 @@ class Decorator {
                 if (thing) {
                     if (thing.$class === `${MetaModelNamespace}.DecoratorTypeReference`) {
                         // XXX Is this really what we want?
-                        if (!MetaModelNamespace[thing.type.name]) {
-                            const errorMessage = `Type '${thing.type.name}' not found within the namespace '${MetaModelNamespace}'.`;
-                            throw new IllegalModelException(errorMessage);
-                        }
                         this.arguments.push({
                             type: 'Identifier',
                             name: thing.type.name,
                             array: thing.isArray
                         });
-                    } else {
+                    } else if(thing.$class.includes('DecoratorTypeReference')){
+                        const errorMessage = `Type '${thing.type.name}' not found within the namespace '${MetaModelNamespace}'.`;
+                        throw new IllegalModelException(errorMessage);
+                    }  else {
                         this.arguments.push(thing.value);
                     }
                 }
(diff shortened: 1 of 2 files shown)


mttrbrts (changes requested):
Thank you for your contribution @ahmad-kashkoush.

Please add a unit test case that exercises your new branch.

ahmad-kashkoush:
> Thank you for your contribution @ahmad-kashkoush.
> 
> Please add a unit test case that exercises your new branch.

Ok, thanks.
I'll work on that, I'll type it in the following file, is that right?
https://github.com/accordproject/concerto/blob/ed6c188d4a03dbc5252a691b51087ad050ee1705/packages/concerto-core/test/introspect/decorator.js#L25

ahmad-kashkoush:
@mttrbrts  I've added Unit test for it
image

packages/concerto-core/test/introspect/decorator.js
```diff
@@ -60,6 +61,24 @@ describe('Decorator', () => {
...
+                            name: 'NonExistentType',
+                            isArray: false
+                        }
+                    }
+                ]
+            };
+
+            const errorMessage=`Type '${ast.arguments[0].type.name}' not found within the namespace '${MetaModelNamespace}'. `;
+            (() => new Decorator(parent, ast)).should.throw(IllegalModelException,errorMessage );
+        });
```
mttrbrts, line 81: Is there an existing test that proves the happy path behaviour? If not, please add a test for the success case too.

packages/concerto-core/lib/introspect/decorator.js
```diff
@@ -78,6 +79,10 @@ class Decorator {
                 if (thing) {
                     if (thing.$class === `${MetaModelNamespace}.DecoratorTypeReference`) {
                         // XXX Is this really what we want?
+                        if (!MetaModelNamespace[thing.type.name]) {
```
sanketshevkar, line 82: Are you sure if the `MetaModelNamespace` has the list of all the declarations present within the namespace?

https://github.com/accordproject/concerto-metamodel/blob/main/index.js#L29
https://concerto.accordproject.org/docs/reference/extending-metamodel/

sanketshevkar: What condition will this not throw an error?

sanketshevkar (changes requested):

packages/concerto-core/lib/introspect/decorator.js
ahmad-kashkoush: I've Looked at it and found it string, so it can't do that, I'm currently looking for a way to find whether the type in the metamodel or not.

packages/concerto-core/test/introspect/decorator.js
ahmad-kashkoush: ok, I'll update it to do so

ahmad-kashkoush:
I've fixed mistakes in PR, may I get a feedback @sanketshevkar

packages/concerto-core/lib/introspect/decorator.js
```diff
@@ -83,7 +84,10 @@ class Decorator {
                             name: thing.type.name,
                             array: thing.isArray
                         });
-                    } else {
+                    } else if(thing.$class.includes('DecoratorTypeReference')){
```
sanketshevkar, line 87: I'm skeptical if this resolves the issue.

Have gone through the links I had shared in the last comment?
Can you please describe what is your understanding/interpretation of the problem and how are you trying to solve it?
I feel there are some gaps in interpretation of the task in hand, I can help fill those.

sanketshevkar (changes requested):

packages/concerto-core/lib/introspect/decorator.js
ahmad-kashkoush: What I was trying to do is throw an error on arguments that have a DecoratorTypeReference but are not in the given namespace. 

I've gone now to links you mentioned and It seems to me that I mixed up between "DecoratorTypeReference" and "type reference argument."

What I should be doing is resolving arguments that are neither boolean, nor string, nor numbers and throwing an error if it has a different namespace.

Is my interpretation correct?

sanketshevkar: Yes exactly! So for this we have someone contributing to build a way to resolve if the decorator type reference is present in the namespace or now.
https://github.com/accordproject/concerto/issues/797
You might want to discuss with them as well what approach are they taking.

ahmad-kashkoush: ok, thanks I'll do so.

and for testing I should add this car file in **(packages/concerto-core/test/data/model/)** Directory and throw error test in **(packages/concerto-core/test/modelmanager.js)** is that right?

sanketshevkar: It depends where you implement the validation(resolving decorator type reference ) logic. If you are are planning to add the logic in `decorator.js` then your test should also reside in file that does test for the class and methods present in `decorator.js`

Issue #854, opened 2024-05-29
Update CTO printer for Import Aliasing


ekarademir:
- [ ] Update CTO printer to read aliased imports in the AST and generate the correct CML expression
- [ ] Update printer test suite

Issue #480, opened 2022-08-09
Import Aliasing


dselman:
# Feature Request 🛍️

Allow imported types to be aliased to prevent name collisions.

## Use Case

Allows a namespace X to import types from namespace Y that have the same names as types already present (or imported) into namespace X.

## Possible Solution

```
import { Document as AccordDocument } from org.accordproject.commonmark
```

## Context

In general as models become larger and edited by distributed teams it is impossible to ensure name conflicts do not occur across namespaces.

## Detailed Description

Allow name aliasing during imports. Alternatively we could support using fully-qualified names for properties.

kailash360:
@dselman @mttrbrts 

I went through this issue, and it looks like we can create a separate field in the JSON schema as `alias` to define the alias name for the property. We will always give priority to the alias over the actual considering that it will prevent the conflicting behavior. In the parser we can add the feature to check for the `as <alias>`  and use that value globally in the file. Let me know if this sounds to be a good approach to start with.

kailash360:
Also, considering that it is one of the GSoC ideas, I would love to work on it.

salujajaskeerat:
To implement the alias import, I am thinking of implementing the syntax suggested by @dselman  [above](https://github.com/accordproject/concerto/issues/480#issue-1333123976).  

- Allowing **optional aliases for types** in import statements.
- Example 
     ` import { document as doc1 , cmk } from org.accordproject.commonmark`
      `import { document as doc2 , utl } from org.accord project.utilities`

In the above types, ` cmk, utl ` are not aliased.

mttrbrts:
This looks like a strong design to me and should be familiar to TypeScript / JavaScript / Python.

A more C#-style approach would use namespace aliasing, although this could become confusing if users tried to use it to alias types too. It's likely that this would also force us to allow fully-qualified references to types throughout CML, rather than using short-names as we do today. 

Are there any other alternatives that we should consider?

salujajaskeerat:
Yes the approach we are looking for is close to the typescript syntax. 

A more C#/C++ style approach  , i think is more relevant for aliasing namespaces . Do we want to alias namespaces also ?  

```c# 
using alias=namespace
```
I believe this approach  would require more syntax changes compared to the above design.

ekarademir:
Some test cases to consider
```
import com.example.foo.{As as As}
```

```
import com.example.foo.{Document as Doc1, Doc2}
import com.example.bax.{Document as Doc2}
```

[...]

PR #855, opened 2024-05-31
feat: alias Import rules added

### LINKED ISSUE #853: Update CTO grammar for Import Aliasing

salujajaskeerat:
* Rules for AliasedIdentifier and Types added.
* aliasedTypes filtered in callback of ImportTypes

# Closes #853 #854 

- Update the parser grammar rules to handle the aliased import
- Update printer for aliased imports
- New test cases added

### Related Issues
- Issue #853 
- Issue #480 
- Issue #854 

### Author Checklist


packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1687,7 +1688,10 @@ FromUri
   = FromToken __ u:$URI __ {
     return u;
   }
+  
 
+  
+
```
mttrbrts, line 1694: Please clean up the extra whitespace

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1715,44 @@ ImportType
...
+    = name:$Identifier _ $AsToken _ aliasName:$Identifier{
+      const result = {
+        "$class":"concerto.metamodel@1.0.0.AliasType",
+        name:name,
+        aliasName:aliasName
+      };
+      return result;
+    }
+  
+Types
```
mttrbrts, line 1749: Let's make this rule name more specific. `Types` is a very ambiguous term for us! :-)

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1715,44 @@ ImportType
...
-    return [head, ...tail];
-  }
+AliasedIdentifier 
+    = name:$Identifier _ $AsToken _ aliasName:$Identifier{
+      const result = {
+        "$class":"concerto.metamodel@1.0.0.AliasType",
+        name:name,
+        aliasName:aliasName
+      };
+      return result;
```
mttrbrts, line 1746:
```suggestion
      return {
        "$class":"concerto.metamodel@1.0.0.AliasType",
        name:name,
        aliasName:aliasName
      };
```

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1715,44 @@ ImportType
...
+    = ImportToken __ ns:QualifiedNamespaceDeclaration ".{" _ types:commaSeparatedTypes _ "}" __ u:FromUri? {
+    	const { aliasedTypes, remainingTypes } = types.reduce((acc, type) => {
+        if (type.$class === "concerto.metamodel@1.0.0.AliasType") {
+          acc.aliasedTypes.push(type);
+          acc.remainingTypes.push(type.name);
+        } else {
+          acc.remainingTypes.push(type);
+        }
+        return acc;
+      }, { aliasedTypes: [], remainingTypes: [] });
```
mttrbrts, line 1727: Changing the indentation here would help readability
```suggestion
          if (type.$class === "concerto.metamodel@1.0.0.AliasType") {
            acc.aliasedTypes.push(type);
            acc.remainingTypes.push(type.name);
          } else {
            acc.remainingTypes.push(type);
          }
          return acc;
        }, { aliasedTypes: [], remainingTypes: [] });
```

packages/concerto-cto/lib/printer.js
```diff
@@ -386,9 +386,17 @@ function toCTO(metaModel) {
             case `${MetaModelNamespace}.ImportAllFrom`:
                 result += `\nimport ${imp.namespace}.*`;
                 break;
-            case `${MetaModelNamespace}.ImportTypes`:
-                result += `\nimport ${imp.namespace}.{${imp.types.join(',')}}`;
+            case `${MetaModelNamespace}.ImportTypes`:{
+                let aliasedTypes=new Map();
+                if(Object.prototype.hasOwnProperty.call(imp,'aliasedTypes')){
```
mttrbrts, line 391: Is there a good reason not to do the following here?

```suggestion
                if(imp.aliasedTypes){
```

packages/concerto-cto/test/cto/aliasedImport.cto
```diff
@@ -0,0 +1,9 @@
+namespace test.aliasing
+
+import org.accordproject.{doc1 as d1,doc2 as d2}
+import org.accordproject.files.{file1,file2 as f2,file3}
+import com.example.foo@1.0.4-X.y.z-01.{baz1 as bar1}
+import com.example.foo@1.0.4.{baz2 as bar2} from https://com.example.co.uk/models/com.example.foo.cto
+import com.example.foo@1.0.4.{tee,baz as bar,lorem,ipsum,dolor as freq}
+import com.example.foo@1.0.4.{As as as,as as as}
```
mttrbrts, line 8: ❤️

mttrbrts (commented):
A great start @salujajaskeerat !

packages/concerto-cto/lib/printer.js
salujajaskeerat: I took this suggestion from eslint documentation, and yes, we can achieve the same as you suggested. 

But the difference in both is as follows:

1. **Object.prototype.hasOwnProperty.call(imp, 'aliasedTypes')**: This checks if the object imp has the property aliasedTypes, regardless of whether it is undefined or null. 
2. **imp.aliasedTypes:** This checks if the property aliasedTypes on the **object imp is true** (i.e., not undefined, null, 0, false, NaN, or an empty string).

In our case, we pass an array of objects as value to the field `aliasedTypes` so both would work the same

packages/concerto-cto/lib/parser.pegjs
salujajaskeerat: Sure. 
Are we using any package that does this automatically before committing the code, something like a prettier package?

packages/concerto-cto/lib/parser.pegjs
salujajaskeerat: How about `IdentifierTypeExpression` or simply `TypeExpression`

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1712,43 @@ ImportType
   }
 
 ImportTypes
-    = ImportToken __ ns:QualifiedNamespaceDeclaration ".{" _ types:commaSeparatedIdentifiers _ "}" __ u:FromUri? {
+    = ImportToken __ ns:QualifiedNamespaceDeclaration ".{" _ types:commaSeparatedTypes _ "}" __ u:FromUri? {
```
ekarademir, line 1715: It is interesting that we don't allow spaces before the curly braces or the dot. But yeah, no need to break it now I guess.

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1712,43 @@ ImportType
...
-  = head:$Identifier _ tail:(',' _ @$Identifier _ )*
-  {
-    return [head, ...tail];
-  }
+AliasedIdentifier 
+    = name:$Identifier _ $AsToken _ aliasName:$Identifier{
+      return {
+        "$class":"concerto.metamodel@1.0.0.AliasType",
+        name:name,
+        aliasName:aliasName
```
ekarademir, line 1741:
```suggestion
        name,
        aliasName
```

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1712,43 @@ ImportType
   }
 
 ImportTypes
-    = ImportToken __ ns:QualifiedNamespaceDeclaration ".{" _ types:commaSeparatedIdentifiers _ "}" __ u:FromUri? {
+    = ImportToken __ ns:QualifiedNamespaceDeclaration ".{" _ types:commaSeparatedTypes _ "}" __ u:FromUri? {
+    	const { aliasedTypes, remainingTypes } = types.reduce((acc, type) => {
+          if (type.$class === "concerto.metamodel@1.0.0.AliasType") {
+            acc.aliasedTypes.push(type);
+            acc.remainingTypes.push(type.name);
```
ekarademir, line 1719:
```suggestion
            acc.typeNames.push(type.name);
```
`remainingTypes` threw me off, I had to read the code to understand that they are not actually the remaining types.

or

```suggestion
            acc.types.push(type.name);
```
to keep it aligned with what will be the final field.

packages/concerto-cto/lib/printer.js
```diff
@@ -386,9 +386,17 @@ function toCTO(metaModel) {
             case `${MetaModelNamespace}.ImportAllFrom`:
                 result += `\nimport ${imp.namespace}.*`;
                 break;
-            case `${MetaModelNamespace}.ImportTypes`:
-                result += `\nimport ${imp.namespace}.{${imp.types.join(',')}}`;
+            case `${MetaModelNamespace}.ImportTypes`:{
```
ekarademir, line 389: Are there any tests for this change?

ekarademir (changes requested):

packages/concerto-cto/test/cto/aliasedImport.cto
```diff
@@ -0,0 +1,9 @@
+namespace test.aliasing
+
+import org.accordproject.{doc1 as d1,doc2 as d2}
```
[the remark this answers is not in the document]
salujajaskeerat: These are test cases. I have only added new test case file .  

So basically we first parse the cto file first to get the AST(json) then this AST is converted back using the Printer. 
Then we check if the original cto string is equal to the new obtained cto string. 

Since we checking the cto strings at the end ,which has also been done in the existing testcases. The output format of the Printer is essential. Thus, I have kept the output format of the Printer and input format of the cto file(test case) the same. 

That is why you don't see the spaces.

packages/concerto-cto/lib/parser.pegjs
salujajaskeerat: Yes, in this grammar, I have only modified the rule for `commaSeparatedTypes`.

packages/concerto-cto/lib/printer.js
salujajaskeerat: Yes, Test cases are evaluated for both parser and printer. 

CTO(original) --[parser]--> AST(JSON) --[printer]--> CTO(new)

we check if CTO(original) == CTO(new)

So we are checking whether the printer is working correctly or not and it is true for all the test cases (where aliasing is not done )

packages/concerto-cto/test/cto/aliasedImport.cto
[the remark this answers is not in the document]
ekarademir: Are there any unit tests where the cases can be added?

ekarademir:
From our discussion:
* Prevent aliasing to primitive types
* Prevent aliasing to previously known keywords

salujajaskeerat:
> From our discussion:
> 
> * Prevent aliasing to primitive types
> * Prevent aliasing to previously known keywords

I have handled the case where we are preventing aliases to primitive types. I don't know if we need to hold for the keywords. We can do it.

The parser allows for the import of keywords from a previous version of grammar (without aliasing).
For example:

```
Namespace saluja.org
import org. eca.{import}
```

The old version of grammar would parse the above.

packages/concerto-cto/lib/parser.pegjs
mttrbrts: Not for pegjs files unfortunately. Our linter only works with .js AFAIK

mttrbrts:
> The parser allows for the import of keywords from a previous version of grammar (without aliasing).

Let's keep things simple for now and maintain the existing behaviour. Please create a tracking issue for this behaviour (note that it would be a breaking change so should be tagged against the v4 release).

ekarademir (approved):

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -97,6 +97,10 @@
       namespace: split.join('.')
     };
   }
+  function isPrimitiveType(typeName) {
+        const primitiveTypes = ['Boolean', 'String', 'DateTime', 'Double', 'Integer', 'Long'];
+        return (primitiveTypes.indexOf(typeName) >= 0);
+  }
```
ekarademir, line 103: nitpick: I guess there isn't a formatter for pegjs files, but could you align the indentation amount with the rest of the file?
```suggestion
  function isPrimitiveType(typeName) {
    const primitiveTypes = ['Boolean', 'String', 'DateTime', 'Double', 'Integer', 'Long'];
    return (primitiveTypes.indexOf(typeName) >= 0);
  }
```

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1687,7 +1692,7 @@ FromUri
   = FromToken __ u:$URI __ {
     return u;
   }
-
+
```
ekarademir, line 1695: nitpick: the same goes for this extra whitespace (the new line is fine. )In VSCode you can set to trim trailing whitespaces.

packages/concerto-cto/lib/parser.pegjs
```diff
@@ -1711,22 +1716,46 @@ ImportType
...
 
-commaSeparatedIdentifiers
-  = head:$Identifier _ tail:(',' _ @$Identifier _ )*
-  {
-    return [head, ...tail];
-  }
+AliasedIdentifier 
+    = name:$Identifier _ $AsToken _ aliasedName:$Identifier{
+      if(isPrimitiveType(aliasedName)){
+        throw new Error(`A type cannot be aliased to a Primitive type, here "${name}" is being aliased to "${aliasedName}".`);
```
ekarademir, line 1743: nitpick:
```suggestion
        throw new Error(`A type cannot be aliased to a Primitive type, here "${name}" is being aliased as "${aliasedName}".`);
```

packages/concerto-cto/lib/printer.js
```diff
@@ -386,9 +386,16 @@ function toCTO(metaModel) {
...
                 break;
-            case `${MetaModelNamespace}.ImportTypes`:
-                result += `\nimport ${imp.namespace}.{${imp.types.join(',')}}`;
+            case `${MetaModelNamespace}.ImportTypes`: {
+                const aliasedTypes = imp.aliasedTypes
+                    ? new Map(imp.aliasedTypes.map(({name,aliasedName})=>[name,aliasedName]))
+                    : new Map();
+                const commaSeparatedTypesString = imp.types.map((type) => aliasedTypes.has(type)
+                    ? `${type} as ${aliasedTypes.get(type)}`
+                    : type).join(',');
```
ekarademir, line 395: did you run the formatter on this part? I think the trailing parentheses looks a bit off.
```suggestion
                const commaSeparatedTypesString = imp.types
                    .map(type => (aliasedTypes.has(type) ? `${type} as ${aliasedTypes.get(type)}` : type))
                    .join(',');
```

packages/concerto-cto/lib/parser.pegjs
salujajaskeerat: Yes, this is a nice shortcut. Thanks

packages/concerto-cto/lib/printer.js
salujajaskeerat:
```js
                const aliasedTypes = imp.aliasedTypes
                    ? new Map(
                        imp.aliasedTypes.map(({ name, aliasedName }) => [
                            name,
                            aliasedName,
                        ])
                    )
                    : new Map();
                const commaSeparatedTypesString = imp.types
                    .map((type) =>
                        aliasedTypes.has(type)
                            ? `${type} as ${aliasedTypes.get(type)}`
                            : type
                    )
                    .join(',');
```

Running formatted outputs as above. Made this change

packages/concerto-cto/package.json
```diff
@@ -56,7 +56,7 @@
     "typescript": "4.6.3"
   },
   "dependencies": {
-    "@accordproject/concerto-metamodel": "3.9.1",
+    "@accordproject/concerto-metamodel": "3.10",
```
ekarademir, line 59: I have another PR to update this. So don't worry about it. But the change should be 
```suggestion
    "@accordproject/concerto-metamodel": "3.10.0",
```

mttrbrts (approved):

PR #857, opened 2024-06-07
feat(dcs): optimize decorate models 

### LINKED ISSUE #856: Optimizations for applying decorators on a model
# Discussion 🗣

The current implementation of applyDecorators is not optimized.
Auto-generated model of size 1.6MB took `Execution time: 6696.545083999634 ms` to apply autogenerated vocabulary to the model.
image

image

both these methods seem to be in the stack for most of the time of execution. `executeCommand` is method internally used by `decorateModels`.

## Context

The process of applying decorators on a model today has a very brute force approach wherein we try apply each and every decorator on all the declarations, properties, map elements etc. [[decorateModels](https://github.com/accordproject/concerto/blob/e6da6739c845d8c80598aa18a16179dff9c88571/packages/concerto-core/lib/decoratormanager.js#L214)]
This works fine for small models with small DCS. But for large models with large DCS this process creates a bottle neck.

## Detailed Description

 - Ideally we would need a index of sort wherein we can store all the decorators that target to a specific declaration, property or map-element. For this we could traverse all all the commands and create an index first, then traverse the metamodel object to the find and apply the decorator on that particular element.
 
 **Challenge**: Since a dcs is not tightly linked to a particular model, we need to define a target/targets for each an and every decorator. 
 Each decorator command's target can be defined in a [lot of](https://github.com/accordproject/concerto/blob/e6da6739c845d8c80598aa18a16179dff9c88571/packages/concerto-core/lib/decoratormanager.js#L60) ways (we don't allow certain [combinations](https://github.com/accordproject/concerto/blob/e6da6739c845d8c80598aa18a16179dff9c88571/packages/concerto-core/lib/decoratormanager.js#L334) eg: property and properties together), which makes it very difficult on the first pass to find out which model element/s is that decorator targeted at. 

- So I tried to write up a logic wherein we can map decorators based on the declarations, properties, map elements, namespace and types. Then try to apply them on metamodel element based on this logic. This is not as optimized as the first approach but still better than traversing over each and every model element against all the decorator commands. #857 

**Challenge:** This works, but we face an issue with specificity, in cases of two or more commands in the same DCS which has the same decorator and target the same element but targeted in different way. The decorator which is a part of the last command will get applied. Unless and until we have a way to define specificity for such similar decorators, its very difficult to predict which one will get applied. Also, another question that comes up is that why would we allow a user to define a same decorator (name is same, but arguments could differ) within the same DCS (assuming they target to the same model) without any rule for specificity.

```
{
    "$class" : "org.accordproject.decoratorcommands@0.3.0.DecoratorCommandSet",
    "name" : "web",
    "version": "1.0.0",
    "commands" : [
        {
            "$class" : "org.accordproject.decoratorcommands@0.3.0.Command",
            "type" : "UPSERT",
            "target" : {
                "$class" : "org.accordproject.decoratorcommands@0.3.0.CommandTarget",
                "namespace" : "test@1.0.0",
                "declaration" : "Person",
                "property" : "bio"
            },
            "decorator" : {
                "$class" : "concerto.metamodel@1.0.0.Decorator",
                "name" : "Form",
                "arguments" : [
                    {
                        "$class" : "concerto.metamodel@1.0.0.DecoratorString",
                        "value" : "inputType"
                    },
                    {
                        "$class" : "concerto.metamodel@1.0.0.DecoratorString",
                        "value" : "textArea"
                    }
                ]
            }
        },
        {
            "$class" : "org.accordproject.decoratorcommands@0.3.0.Command",
            "type" : "UPSERT",
            "target" : {
                "$class" : "org.accordproject.decoratorcommands@0.3.0.CommandTarget",
                "type" : "concerto.metamodel@1.0.0.StringProperty"
            },
            "decorator" : {
                "$class" : "concerto.metamodel@1.0.0.Decorator",
                "name" : "Form",
                "arguments" : [
                    {
                        "$class" : "concerto.metamodel@1.0.0.DecoratorString",
                        "value" : "inputType"
                    },
                    {
                        "$class" : "concerto.metamodel@1.0.0.DecoratorString",
                        "value" : "text"
                    }
                ]
            }
        }
    ]
}
```
In this example both decorators have the same name, but the second decorator would get we'll be seen on the final model based on the current logic. With changes proposed in #857 it's difficult to predict which was the last one in the list. The decorators get segregated based on specificity of the command targets.
Specificity of targets defined as:
type > property > mapElements  > declaration > namespace.

This is based on the heuristic that if a type is defined in target, it will take the highest precedence amongst all other target properties. Next highest precedence is for properties and map-elements, then declarations and then namespace.

```
target(namespace) = applyDecoratorTo all decl in namespace
target(namespace + declaration) = applyDecoratorTo to declaration
target(namespace + declaration + property) = applyDecoratorTo  property
target(namespace + declaration + mapElement) = applyDecoratorTo mapElement
target(namespace + declaration + property/mapElement + type) = applyDecoratorTo to type
```
Logic for [executeCommand](https://github.com/accordproject/concerto/blob/e6da6739c845d8c80598aa18a16179dff9c88571/packages/concerto-core/lib/decoratormanager.js#L451)

sanketshevkar:
# Closes #856 

Proposal for optimizing decorate models. More details on #856 

### Changes

- Index all the commands in dcs in their respective maps based on their primary target: namespace, declaration, property, mapElement, types all have their separate maps. Wrap the commands with their index in dcs before indexing them.
- Traverse through the model ast and lookup into the maps for any command related to the targets. Collect them in a list and sort them based on index before applying on the model element.
- Function signatures have more or less stayed the same, except for `executeCommand` but are not a breaking change in nature.
- Logic for `executeCommand` has changed a bit and  `decorateModels` has changed significantly. New private methods have been introduced to support these.
- New test case has added that aims at covering all possible command target combinations in dcs. (though [list](https://docs.google.com/document/d/15SoDFYwkzi0duOw-y4_owyrCT3MDRJCe7SVIa8MD2Ak/edit?usp=sharing) is not exhanustive)

### Performance
- Auto-generated model of size 1.6MB took Execution `time: 6696.545083999634 ms` to apply autogenerated vocabulary to the model with the older implementation.
- New implementation takes `Execution time: 44.81274998188019 ms` for the same operation.
```
const start = performance.now();
DecoratorManager.decorateModels(mm, vocDcs);
const end = performance.now();
console.log(`Execution time: ${end - start} ms`);
```
- 152 times faster. <<hmm this makes me skeptical 💭 >>

#### `total_heap_size`
```
DecoratorManager.decorateModels(mm, vocDcs);
console.log(v8.getHeapStatistics());
```
- `total_heap_size` recorded after an auto-generated model of size 1.6MB  was applied autogenerated vocabulary to the model with the older implementation was `58261504`
- New implementation recorded `total_heap_size` as `61571072`
- 1.06 times more space
- 5% increase in `total_heap_size`

#### `used_heap_size`
```
console.log(v8.getHeapStatistics());
DecoratorManager.decorateModels(mm, vocDcs);
console.log(v8.getHeapStatistics());
```
- Increase in `used_heap_size` recorded after an auto-generated model of size 1.6MB  was applied autogenerated vocabulary to the model with the older implementation was `6541360`
- New implementation recorded increase `used_heap_size` as `13041876`
- 2 times more increase in the increase of `used_heap_space`
- 99.3% increase in increase of `used_heap_size`

### Author Checklist


packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,178 @@ class DecoratorManager {
...
+     * @param {*} decoratorCommandSet the DecoratorCommandSet object
+     * @param {object} [options] - decorator models options
+     * @param {boolean} [options.validate] - validate that decorator command set is valid
+     * with respect to to decorator command set model
+     * @param {boolean} [options.validateCommands] - validate the decorator command set targets. Note that
+     * the validate option must also be true
+     * @param {boolean} [options.migrate] - migrate the decoratorCommandSet $class to match the dcs model version
+     * @returns {ModelManager} a new model manager with the decorations applied
+     */
+    static optimizedDecorateModels(modelManager, decoratorCommandSet, options) {
```
sanketshevkar, line 328: Need to update update the method based on the specificity mentioned in the issue #856

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,178 @@ class DecoratorManager {
...
         return newModelManager;
     }
+
+    /**
+     * Applies all the decorator commands from the DecoratorCommandSet
+     * to the ModelManager.
+     * @param {*} decoratorCommandSet the DecoratorCommandSet object
+     * @returns {Object} a new model manager with the decorations applied
+     */
+    static getDecoratorMaps(decoratorCommandSet) {
```
mttrbrts, line 263: I think that this should be a private function?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,178 @@ class DecoratorManager {
...
+                }
+
+                if (namespaceCommandsMap.has(model.namespace)) {
+                    const commands = namespaceCommandsMap.get(model.namespace);
+                    console.log('name');
+                    commands.forEach(command => {
+                        this.executeDeclarationCommand(model.namespace, decl, command);
+                    });
+                }
+
```
mttrbrts, line 378: This should be outside the declarations.forEach loop

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -253,6 +253,178 @@ class DecoratorManager {
...
+        const mapElementCommandsMap = new Map();
+        const typeCommandsMap = new Map();
+
+        decoratorCommandSet.commands.map(decoratorCommand => {
+            if(decoratorCommand.target.type) {
+                if (typeCommandsMap.has(decoratorCommand.target.type)) {
+                    typeCommandsMap.get(decoratorCommand.target.type).push(decoratorCommand);
+                } else {
+                    typeCommandsMap.set(decoratorCommand.target.type, [{...decoratorCommand}]);
+                }
```
mttrbrts, line 276: You can optimize this a little by removing one of the lookups ...
```suggestion
                const typeCommands = typeCommandsMap.get(decoratorCommand.target.type);
                if (typeCommands) {
                    typeCommands.push(decoratorCommand);
                } else {
                    typeCommandsMap.set(decoratorCommand.target.type, [{...decoratorCommand}]);
                }
              }
```

packages/concerto-core/lib/decoratormanager.js
mttrbrts: The name doesn't seem right, either

mttrbrts (changes requested):
It looks like you're on the right track, @sanketshevkar. I look forward to seeing the perf numbers!

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: https://github.com/accordproject/concerto/issues/858#issuecomment-2157341398

packages/concerto-core/test/dcsIndexWrapper.js
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const DcsIndexWrapper = require('../lib/DcsIndexWrapper');
```
sanketshevkar, line 17: @mttrbrts Coverage for the file this test targets at is showing 0%. Is there anything more I should be doing if I add a new test suite?

packages/concerto-core/api.txt
```diff
@@ -55,17 +55,22 @@ class Concerto {
    + string getNamespace(obj) 
 }
    + object setCurrentTime() 
+class DcsIndexWrapper {
```
mttrbrts, line 58: Do we want this to public?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -200,26 +201,80 @@ class DecoratorManager {
...
+        if (targetCommands) {
+            targetCommands.push(dcsWithIndex);
+        } else {
+            targetMap.set(targetKey, [dcsWithIndex]);
+        }
+    }
+
+
+    /**
+     * Creates five difeerent maps to index decorator command sets by target type and returns them
```
mttrbrts, line 221:
```suggestion
     * Creates five different maps to index decorator command sets by target type and returns them
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -200,26 +201,80 @@ class DecoratorManager {
...
+    static getDecoratorMaps(decoratorCommandSet) {
+        const namespaceCommandsMap = new Map();
+        const declarationCommandsMap = new Map();
+        const propertyCommandsMap = new Map();
+        const mapElementCommandsMap = new Map();
+        const typeCommandsMap = new Map();
+
+        decoratorCommandSet.commands.map((decoratorCommand, index) => {
+            const dcsWithIndex = new DcsIndexWrapper(decoratorCommand, index);
+            if(decoratorCommand.target.type) {
```
mttrbrts, line 235: Is there an advantage to using a switch statement here?

mttrbrts: Also, how do we protect against falsy, but non-null values in these conditions?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -242,16 +297,88 @@ class DecoratorManager {
...
+     * @param {targetKey} targetKey the target key to add the command to
+     * @param {DcsIndexWrapper[]} computedList the computed list to add the command to
+     * @private
+     */
+    static addToComputedList(targetMap, targetKey, computedList) {
+        const targetCommands = targetMap.get(targetKey);
+        if (targetCommands) {
+            targetCommands.forEach((dcsWithIndex) => {
+                computedList.push(dcsWithIndex);
+            });
```
mttrbrts, line 314: Can you push multiple items in one operation?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -242,16 +297,88 @@ class DecoratorManager {
                 });
             }
         }
+    }
+
+    /**
+     * Adds decorator commands with index to the computed list
+     * @param {*} targetMap the target map to add the command to
+     * @param {targetKey} targetKey the target key to add the command to
+     * @param {DcsIndexWrapper[]} computedList the computed list to add the command to
```
mttrbrts, line 306: What's a computed list? The naming of this parameter and function aren't very clear to me.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -242,16 +297,88 @@ class DecoratorManager {
...
+        if (targetCommands) {
+            targetCommands.forEach((dcsWithIndex) => {
+                computedList.push(dcsWithIndex);
+            });
+        }
+    }
+
+    /**
+     * Applies all the decorator commands from the DecoratorCommandSet to the ModelManager,
+     * this is optimized version of decorateModel method and will eventually replace its internal logic.
```
mttrbrts, line 320:
```suggestion
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -242,16 +297,88 @@ class DecoratorManager {
...
+     * @param {ModelManager} modelManager the input model manager
+     * @param {*} decoratorCommandSet the DecoratorCommandSet object
+     * @param {object} [options] - decorator models options
+     * @param {boolean} [options.validate] - validate that decorator command set is valid
+     * with respect to to decorator command set model
+     * @param {boolean} [options.validateCommands] - validate the decorator command set targets. Note that
+     * the validate option must also be true
+     * @param {boolean} [options.migrate] - migrate the decoratorCommandSet $class to match the dcs model version
```
mttrbrts, line 328: I see that you refactored the code above to use unbox these options. What are the risks of inconsistent representation of the same options?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -242,16 +297,88 @@ class DecoratorManager {
...
+    }
+
+    /**
+     * Adds decorator commands with index to the computed list
+     * @param {*} targetMap the target map to add the command to
+     * @param {targetKey} targetKey the target key to add the command to
+     * @param {DcsIndexWrapper[]} computedList the computed list to add the command to
+     * @private
+     */
+    static addToComputedList(targetMap, targetKey, computedList) {
```
mttrbrts, line 309:
```suggestion
    static pushMapValues(array, map, key) {
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -526,14 +646,16 @@ class DecoratorManager {
...
         const { target, decorator, type } = command;
-        if (
-            this.falsyOrEqual(
-                target.property ? target.property : target.properties,
-                [property.name]
-            ) &&
-            this.falsyOrEqual(target.type, [property.$class])
-        ) {
-            this.applyDecorator(property, type, decorator);
+        if(target.properties || target.property || target.type) {
```
mttrbrts, line 649: Does this help with the performance issue that you found?
Are there any non-null, falsy values that need checking here?

packages/concerto-core/test/dcsIndexWrapper.js
mttrbrts: Do you see the same locally, as in the GH Actions build?

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Done

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: fixed

packages/concerto-core/api.txt
mttrbrts: After reviewing the rest of the code, I think not. This is an internal helper class.

packages/concerto-core/test/dcsIndexWrapper.js
sanketshevkar: Locally

sanketshevkar: Moved to `DecoratorManager.js` fixed the issue

packages/concerto-core/api.txt
sanketshevkar: Moved to `DecoratorManager.js`

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Fixed

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: > Is there an advantage to using a switch statement here?

Yes, refactored it to have switch case.

> Also, how do we protect against falsy, but non-null values in these conditions?

They won't get added to their respective maps. `target.[elementTarget]` always needs to be string to be valid, if non-null falsy, it won't be added to any of the maps.

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Done

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Fixed for all lists

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Fixed

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Done

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: > Does this help with the performance issue that you found?

There was an edge case where it applied decorators on map elements. This handles that edge case.

>  Are there any non-null, falsy values that need checking here?

Nope, even if there exists that means that it is a malformed command.

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: for `migrateAndValidate` method it doesn't feel right to me sending them as options as the internal functionality needs at least one of the flags to be passed for it to do something. And moreover its a private method that is used inside the `decorateModel` method.
There are similar couple of private methods that are truly dependent on a flag as the main argument for it to function,

for `executeCommand` I've sent the `enableDecoratorsForNamespace` as an option as its a public function and the flag is actually an optional flag.

mttrbrts (approved):

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -200,26 +235,87 @@ class DecoratorManager {
...
+        } else {
+            targetMap.set(targetKey, [dcsWithIndex]);
+        }
+    }
+
+
+    /**
+     * Creates five different maps to index decorator command sets by target type and returns them
+     * @param {*} decoratorCommandSet the DecoratorCommandSet object
+     * @returns {Object} a new model manager with the decorations applied
```
dselman, line 257: Incorrect JSDoc.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -200,26 +235,87 @@ class DecoratorManager {
...
+    static getDecoratorMaps(decoratorCommandSet) {
+        const namespaceCommandsMap = new Map();
+        const declarationCommandsMap = new Map();
+        const propertyCommandsMap = new Map();
+        const mapElementCommandsMap = new Map();
+        const typeCommandsMap = new Map();
+
+        decoratorCommandSet.commands.map((decoratorCommand, index) => {
+            const dcsWithIndex = new DcsIndexWrapper(decoratorCommand, index);
+            switch (true) {
```
dselman, line 269: Is this  switch supposed to be exhaustive? If yes, add a default case and throw an exception.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -454,26 +620,26 @@ class DecoratorManager {
...
     /**
-     * Executes a Command against a ClassDeclaration, adding
-     * decorators to the ClassDeclaration, or its properties, as required.
+     * Executes a Command against a Declaration, adding
+     * decorators to the Declaration, or its properties, as required.
      * @param {string} namespace the namespace for the declaration
      * @param {*} declaration the class declaration
      * @param {*} command the Command object from the dcs
-     * @param {boolean} [enableDcsNamespaceTarget] - flag to control applying namespace targeted decorators on top of the namespace instead of all declarations in that namespace
+     * @param {*} property the property
```
dselman, line 637: Looks like property is now an optional argument? Would be good to document that.

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -123,6 +123,41 @@ function isUnversionedNamespaceEqual(modelFile, unversionedNamespace) {
     return name === unversionedNamespace;
 }
 
+/**
+ * Helper class to wrap the decorator command and the index of the command
+ * [DecoratorCommandSet](https://models.accordproject.org/concerto/decorators.cto)
+ * @memberof module:concerto-core
+ */
+class DcsIndexWrapper {
```
dselman, line 131: Private?

packages/concerto-core/types/lib/dcsIndexWrapper.d.ts
dselman, line 1: Should not be exported? Private?

packages/concerto-core/types/lib/decoratormanager.d.ts
```diff
@@ -37,8 +37,42 @@ declare class DecoratorManager {
...
+    private static addDcsWithIndexToMap;
+    /**
+     * Creates five different maps to index decorator command sets by target type and returns them
+     * @param {*} decoratorCommandSet the DecoratorCommandSet object
+     * @returns {Object} a new model manager with the decorations applied
+     * @private
+     */
+    private static getDecoratorMaps;
+    /**
+     * Migrate or validate the DecoratorCommandSet object if the options are set as true
```
dselman, line 55: Migrate AND validate?

dselman (changes requested):
Couple of minor changes. Looks good.

packages/concerto-core/types/lib/decoratormanager.d.ts
sanketshevkar: It can also do either of the operation if parameters for the arguments are passed so. It can only validate or it might only migrate and can do both if the parameters are true for both.

The only reason I moved this out of `decorateModels` is to reduce the code in the `decorateModels` method, so is a private method just used in `decorateModels` method.

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: I was under the assumption that if we pass a invalid dcs, it will only get caught if validation is enabled. In the case of a malformed decorator, if we pass (in this case target is undefined, falsy, null or does not match with any of the switch statements) that decorator won't get applied.

This is the older behavior: https://replit.com/@sanketshevkar/Decorator-malformed-dcs#index.js
This differs a bit, I think is was a bug as well. shouldn't `target: false` be invalid dcs, but its passing the syntax validation. In either case it just applies the decorator on top of all the declarations.

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Corrected

packages/concerto-core/types/lib/dcsIndexWrapper.d.ts
sanketshevkar: Yes made the method private

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Yes corrected

dselman (approved):
Bravo!

PR #861, opened 2024-06-15
feat(serializer): add inferClass option

### LINKED ISSUE #482: Optimize JSON (De)serialization (Final Types et al)
# Feature Request 🛍️

## Use Case

Allow modellers to declare that types are "final" and cannot be extended. This allows for optimisation of the serialised JSON, as an array of final types can no longer be polymorphic.

The $class is required in serialised instances because concepts can be extended. We could support the addition of a final keyword which would indicate that a concept could not be extended, meaning that the type of an object can be inferred by introspecting the model.

## Possible Solution

```
namespace acme

final concept Person {
   o String name
}

final concept Company {
   o String name
   o Person[] employees
}

const company = 
{
   "name" : "Clause",
   "employees" : [
      {"name" : "Dan"}, {"name", "Simon"}
   ]
}

// deserialize, passing in the name of the type explicitly
// during deserialization the model is introspected to know that company.employee is an Employee type (and cannot be a subclass).
serialiser.fromJSON( company, ''acme.Company' );
```

dselman:
# Closes #482 #542

Adds the option `inferClass` to the `Serializer` (false by default). When this option is `true` `$class` will only by included in the JSON created for `Resource` objects, in the following circumstances:
1. If this is a root object, *root objects always have a $class attribute* so that they are self-describing
2. If a nested object is not of the same type as the type of its field in the model

In addition, when a `$class` attribute is included it will be shortened (the namespace is removed) when the namespace of the `Resource` is the same as the namespace of the type of the property (see examples).

### Changes
- `JSONGenerator` updated to remove $class from JSON when it can be inferred
- `JSONPopulator` updated to infer $class FQN for objects when it is short or missing
- `fromAst` methods updated to support either short or fully-qualified names for the meta model
- Unit tests

### Flags
- Do we want to only support `inferClass` when model manager is in strict mode?

### Examples

Using these two model files:

```
            namespace org.acme.zoo@1.0.0

            abstract concept Animal {
               o String name
            }

            concept Address {
               o String line1
               o String line2 optional
               o String city
               o String state
               o String country
            }

            // a type that extends Animal, in the same ns as Animal
            concept Dog extends Animal{}

            abstract concept Person {
               o Address address optional // can be inferred from model
               o String name
            }

            concept Owner extends Person {
               o Integer age
            }

            concept Zoo {
               o Person person // $class cannot be inferred from model, as `Person` is abstract
               o Animal[] animals // $class cannot be inferred from model
            }
```

And:

```
            namespace org.acme.cat@1.0.0
            import org.acme.zoo@1.0.0.{Animal}
            // a type that extends Animal in a different namespace
            concept Cat extends Animal{}
```

### Example 1

This example will deserialize with `inferClass=true`. 

1. The `person` property has the `$class` value `Owner` because it is not of the type of the field (`Person`) but is of type `org.acme.zoo@1.0.0.Owner` which is in the same namespace as the field type, so is shortened to just `Owner`.
2. The first entry in the `animals` array has a $class of `Dog`. A $class is required because the type of the property is `Animal` and there are two types that specialise `Animal`: `Cat` and `Dog`, so a discriminator is required.
3. The $class is `Dog` not `org.acme.zoo@1.0.0.Dog` because the type `Dog` and the type `Animal` are defined in the same namespace. During deserialisation the serialised assumes that $class short names are defined in the same namespace as the type of their property in the model.
4. No $class is required for `person.address` as the type of the field is `org.acme.zoo@1.0.0.Address` is the same as the resource.

```
{
                $class: 'org.acme.zoo@1.0.0.Zoo',
                person: {
                    $class: 'Owner',
                    name: 'Dan',
                    age: 42,
                    address: {
                        line1: '1 Main Street',
                        city: 'Boston',
                        state: 'MA',
                        country: 'USA'
                    }
                },
                animals: [
                    { $class: 'Dog', name: 'fido' }
                ]
            }
```

### Example 2

This example will deserialize with `inferClass=true`. 

1. The $class is `org.acme.cat@1.0.0.Cat` because the type `Cat` and the type `Animal` are not defined in the same namespace.

```
{
                $class: 'org.acme.zoo@1.0.0.Zoo',
                person: {
                    $class: 'Owner',
                    name: 'Dan',
                    age: 42
                },
                animals: [
                    {
                        $class: 'org.acme.cat@1.0.0.Cat',
                        name: 'tiddles'
                    }
                ]
            }
```

### Example 3

This example will deserialize with `inferClass=true`.

1. This is for backwards compatibility and shows that $class can still be explicitly provided for `owner` and the first element in the `animals` array.

```
{
                $class: 'org.acme.zoo@1.0.0.Zoo',
                person: {
                    $class: 'org.acme.zoo@1.0.0.Owner',
                    name: 'Dan',
                    age: 42
                },
                animals: [
                    {
                        $class: 'org.acme.zoo@1.0.0.Dog',
                        name: 'fido'
                    }
                ]
            }
```

### Example 4

Here is a meta model instance with `inferClass=true`:

```
{
  "$class": "concerto.metamodel@1.0.0.Model",
  "namespace": "Lorem nisi enim enim.",
  "sourceUri": "Cupidatat officia laborum sunt incididunt.",
  "concertoVersion": "Veniam.",
  "imports": [
    {
      "$class": "ImportAll",
      "namespace": "Duis sint.",
      "uri": "Ea sunt reprehenderit."
    }
  ],
  "declarations": [
    {
      "$class": "MapDeclaration",
      "key": {
        "$class": "StringMapKeyType",
        "decorators": [
          {
            "name": "Incididunt excepteur nostrud enim.",
            "arguments": [
              {
                "$class": "DecoratorString",
                "value": "Cillum occaecat aute aute.",
                "location": {
                  "start": {
                    "line": 38511,
                    "column": 43930,
                    "offset": 23033
                  },
                  "end": {
                    "line": 36502,
                    "column": 39816,
                    "offset": 19486
                  },
                  "source": "Labore tempor et aliquip mollit."
                }
              }
            ],
            "location": {
              "start": {
                "line": 42391,
                "column": 3627,
                "offset": 33958
              },
              "end": {
                "line": 983,
                "column": 40090,
                "offset": 25069
              },
              "source": "Excepteur tempor veniam pariatur."
            }
          }
        ],
        "location": {
          "start": {
            "line": 61309,
            "column": 47612,
            "offset": 47848
          },
          "end": {
            "line": 4788,
            "column": 48475,
            "offset": 41651
          },
          "source": "Non anim nisi ipsum occaecat."
        }
      },
      "value": {
        "$class": "BooleanMapValueType",
        "decorators": [
          {
            "name": "Commodo ut tempor eiusmod.",
            "arguments": [
              {
                "$class": "DecoratorString",
                "value": "Nulla ea sit pariatur incididunt.",
                "location": {
                  "start": {
                    "line": 32231,
                    "column": 39463,
                    "offset": 11115
                  },
                  "end": {
                    "line": 5812,
                    "column": 46231,
                    "offset": 50449
                  },
                  "source": "Ut cupidatat nisi duis elit."
                }
              }
            ],
            "location": {
              "start": {
                "line": 44063,
                "column": 49727,
                "offset": 33761
              },
              "end": {
                "line": 51897,
                "column": 32889,
                "offset": 39320
              },
              "source": "Pariatur laboris adipisicing."
            }
          }
        ],
        "location": {
          "start": {
            "line": 37476,
            "column": 16394,
            "offset": 12273
          },
          "end": {
            "line": 50689,
            "column": 63232,
            "offset": 29302
          },
          "source": "In veniam anim ut."
        }
      },
      "name": "Foo",
      "decorators": [
        {
          "name": "Labore cillum.",
          "arguments": [
            {
              "$class": "DecoratorString",
              "value": "Qui.",
              "location": {
                "start": {
                  "line": 8643,
                  "column": 22128,
                  "offset": 37375
                },
                "end": {
                  "line": 35202,
                  "column": 14646,
                  "offset": 36723
                },
                "source": "Sint eu tempor sint."
              }
            }
          ],
          "location": {
            "start": {
              "line": 63155,
              "column": 29327,
              "offset": 52746
            },
            "end": {
              "line": 58986,
              "column": 61943,
              "offset": 29968
            },
            "source": "Sint excepteur sunt elit."
          }
        }
      ],
      "location": {
        "start": {
          "line": 16561,
          "column": 54945,
          "offset": 30256
        },
        "end": {
          "line": 8976,
          "column": 9505,
          "offset": 43034
        },
        "source": "Labore."
      }
    }
  ],
  "decorators": [
    {
      "name": "Pariatur aute voluptate id.",
      "arguments": [
        {
          "$class": "DecoratorString",
          "value": "Sit.",
          "location": {
            "start": {
              "line": 11878,
              "column": 38094,
              "offset": 16933
            },
            "end": {
              "line": 12942,
              "column": 61905,
              "offset": 17270
            },
            "source": "Anim occaecat proident."
          }
        }
      ],
      "location": {
        "start": {
          "line": 26284,
          "column": 63143,
          "offset": 13162
        },
        "end": {
          "line": 28658,
          "column": 38157,
          "offset": 63188
        },
        "source": "Id nisi."
      }
    }
  ]
}
```

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist


mttrbrts:
> Do we want to only support inferClass when model manager is in strict mode?

Yes, I say so

> Do we want to be stricter in fromJSON with inferClass=false to ensure that $class is always present and is FQN, or is it ok to attempt type inference in all cases?

We need to be backwards compatible. So I say, no for v3, but this sounds reasonable for v4

mttrbrts:
This change has the side effect of making some types implicitly final, if they weren't in scope of the model manager when the serialisation occurred.

For example, it's possible that a namespace is unambiguous to one client, and ambiguous to another.

dselman:
> This change has the side effect of making some types implicitly final, if they weren't in scope of the model manager when the serialisation occurred.
> 
> For example, it's possible that a namespace is unambiguous to one client, and ambiguous to another.

Yes, this one is worrying. I don't see a good solution to this, in as far as adding a new model file to the model manager could have a side-effect on serialisation of types, other than adding an explicit `final` but that would seriously limit the scope of the optimisation. So the assumption is that the model manager at `T0` when JSON was generated by `toJSON` contains the same model files as at `T1` when `fromJSON` is called.

mttrbrts:
In the real world, I see this optimization as hugely beneficial for fixed scope scenarios, such as the following:
1. Serialisation of metamodel instances
2. Serialisation of decorator command sets
3. Serialisation of models in a known closed domain, 

In these cases, it's a fair assumption that the namespace domain is stable (e.g. it only contains the metamodel definitions), so the ambiguous cases should not arise.

Use of an explicit `final` keyword would also likely have most benefit in these same scenarios.

Serialisation of userland models is more dangerous unless the system can also persist the namespace domain and make an assumption like you suggest.

Making everything implicitly `final` (and introducing an explicit `extensible` modifier) would be safer and give us more scope for optimisation, but that's a big breaking change.

mttrbrts:
Adhoc testing shows about 1/3 file size reduction for a 29Mb model definition, and 73Mb Decorator Command Set file.

packages/concerto-core/lib/serializer/jsongenerator.js
```diff
@@ -142,18 +144,40 @@ class JSONGenerator {
...
+        parameters.isRoot = false;
+
         // Walk each property of the class declaration
         const properties = classDeclaration.getProperties();
         for (let index in properties) {
             const property = properties[index];
             const value = obj[property.getName()];
             if (!Util.isNull(value)) {
                 parameters.stack.push(value);
+                parameters.field = property; // save the property to the stack!
```
mttrbrts, line 180: Where do we pop this value?

mttrbrts (commented):
We'll likely need similar changes to the metamodel utils and decorator manager to allow loading and printing of modelfile definitions and dcss with inferred classes.

packages/concerto-core/lib/serializer/jsongenerator.js
```diff
@@ -142,18 +144,40 @@ class JSONGenerator {
...
+        // by default we set the $class on the result
         result.$class = classDeclaration.getFullyQualifiedName();
+
+        // if we are in the inferClass mode and this is not the root instance
+        // we attempt to remove the $class
+        if(this.inferClass && !parameters.isRoot && parameters.field) {
+            const fieldClassDecl = parameters.modelManager.getType(parameters.field.getFullyQualifiedTypeName());
+            // if the $class cannot be unambigiously inferred from the type of the field in the model
+            // we have to include it, but we attempt to shorten it, if the object and the field are in the same ns
+            if(fieldClassDecl.getAssignableClassDeclarations().length > 2) {
```
DS-AdamMilazzo, line 156: It looks like getAssignableClassDeclarations finds all equal or derived types in all known models. If so, then this looks like it keeps the $class when there exist at least three (why not two?) matching types, but that seems like an unreliable property because a new model can extend a concept at any time, leading to the case later on where something that "was unambigious is now ambigious".

I'd recommend omitting the class only when it's exactly equal to the declared field type. That is a stable property that can't break with the addition of new models.

packages/concerto-core/lib/serializer/jsonpopulator.js
```diff
@@ -77,6 +77,51 @@ function validateProperties(properties, classDeclaration) {
...
+    }
+    else {
+        const fqn = field.getFullyQualifiedTypeName();
+        const mm = field.getParent().getModelFile().getModelManager();
+        const classDecl = mm.getType(fqn);
+        const assignable = classDecl.isClassDeclaration?.() ? classDecl.getAssignableClassDeclarations()
+            .filter( a => !a.isAbstract()) : [classDecl];
+        if(assignable.length !== 1) {
+            throw new Error(`The type ${fqn} which was unambigious is now ambigious due to ${assignable.map(a => a.getFullyQualifiedName()).join(',')}`);
+        }
```
DS-AdamMilazzo, line 118: This case worries me. I think we shouldn't omit $class in any case where it could become ambiguous in the future.

packages/concerto-core/lib/serializer/jsongenerator.js
DS-AdamMilazzo, line 156: Do we want the !isAbstract filter here like you have below?

packages/concerto-core/lib/serializer/jsongenerator.js
dselman: The comment is misleading, this is not pushed to the stack, it is just saved to parameters.

packages/concerto-core/lib/serializer/jsongenerator.js
dselman: > I'd recommend omitting the class only when it's exactly equal to the declared field type. That is a stable property that can't break with the addition of new models.

I like this idea — it limits the scope of the optimisation, but should still give us some value for simple cases.

dselman:
Simplified inference logic to assume the type of a field, if $class is missing.

dselman:
Before we can load AST created with `inferClass=true` we will have to update code like this:
https://github.com/accordproject/concerto/blob/5b21276e07bf7416b9ec9a65444eefa001206159/packages/concerto-core/lib/introspect/modelfile.js#L744

Which does not use `Resource`.

mttrbrts:
Nicely done. The size reduction is still around 1/3 for metamodel and DCS, which is good news!

Are you going to patch up the modelfile definition (and similar) in this PR too?

dselman:
> Nicely done. The size reduction is still around 1/3 for metamodel and DCS, which is good news!
> 
> Are you going to patch up the modelfile definition (and similar) in this PR too?

Yes. I will add a test that does the full roundtrip with inferClass true/false on a meta model...

PR #866, opened 2024-06-20
fix(core): falsy default values

### LINKED ISSUE #865: Serialization exception when declaring a default false boolean field
# Bug Report 🐛

Assume a model definition as follows:
```cs
namespace test@1.0.0

concept Animal {
    o Boolean isAlive default=false
}
```

When performing `Serializer.fromJSON` with the instance
```cs
{ "$class": "test@1.0.0.Animal"}
```
The following exception is thrown

```
ValidationException: The instance "test@1.0.0.Animal" is missing the required field "isAlive".
```

Note that the instance does deserialize correctly against the following model.
```
concept Animal {
    o Boolean isAlive default=true
}
```
## Possible Solution

In the ResourceValidator add the following line before throwing the exception, 
```js
                    if (property?.defaultValue !== null){
                        continue;
                    }
```

And change the behaviour of the `process` function in the Field class to
```js

-        if (this.ast.defaultValue) {
+        if (this.ast.defaultValue !== null) {
             this.defaultValue = this.ast.defaultValue;
         } else {
             this.defaultValue = null;
         }
```

Similarly, update the `getDefaultValue` function.

## Detailed Description

When specifying a truthy default value on a field, it becomes implicitly optional. It's not clear if this is intentional.

Note that the Metamodel contains this pattern in several places, e.g. https://github.com/accordproject/concerto-metamodel/blob/main/lib/metamodel.cto#L143-L144

mttrbrts:
# Closes #865

Fixes the edge case of implicit optionals where the default value is falsy.

e.g. 
```cs
concept Animal {
  o Integer ageInYears default=0
}
```

Serialization should assign the default value correctly to 0, but this was previously detected as a missing default value because `0` is falsy in JavaScript.

### Author Checklist


packages/concerto-cto/test/cto/field.cto
```diff
@@ -2,4 +2,8 @@ namespace test.field
 
 concept Foo {
   o String foo default="FOO" optional
-}
\ No newline at end of file
+  o Double doubleDefaultZero default=0.0 optional
+  o Integer intDefaultZero default=0 optional
+  o Long longDefaultZero default=0 optional
+  o Boolean boolDefaultFalse default=false optional
```
ekarademir, line 8: Maybe a case for default value not even set?

packages/concerto-core/test/models/test.js
```diff
@@ -208,10 +208,6 @@ describe('Test Model', function(){
             cObject.model = { 'foo' : 'bar' };
             ( function() {serializer.toJSON(cObject);}).should.throw('Model violation in the "org.acme.Vehicle#YVBVLSFXXAL342374" instance. The field "model" has a value of "{"foo":"bar"}" (type of value: "object"). Expected type of value: "String".');
 
-            // set model to null
```
ekarademir, line 211: Is this redundant now?

ekarademir (approved):

packages/concerto-cto/test/cto/field.cto
mttrbrts: I don't think that we should add a default modifier in the JSON serialization of a model, if the author didn't specify one in the CML. The behaviour is consistent now with this PR.

packages/concerto-core/test/models/test.js
mttrbrts: Yes, the model contains a default modifier, which is now being used properly.
```
o String model default="F150"
```

packages/concerto-core/lib/introspect/field.js
```diff
@@ -99,11 +100,11 @@ class Field extends Property {
     }
 
     /**
-     * Returns the default value for the field or null
+     * Returns the default value for the field or null if there is no default value
      * @return {string | number} the default value for the field or null
      */
     getDefaultValue() {
-        if (this.defaultValue) {
+        if (!Util.isNull(this.defaultValue)) {
```
dselman, line 107: Didn't we already do this on line 87? Why do we need to do it again?

packages/concerto-cto/lib/printer.js
```diff
@@ -147,6 +147,16 @@ function typeFromMetaModel(mm){
     return result;
 }
 
+/**
+ * Returns true if the typeof the object === 'undefined' or
+ * the object === null.
+ * @param {Object} obj - the object to be tested
+ * @returns {boolean} true if the object is null or undefined
+ */
+function isNull(obj) {
```
dselman, line 156: Why not import  Util.isNull from concerto-core? E.g. we could move isNull into ModelUtil and get rid of this Util class.

mttrbrts: I was trying to avoid adding another dependency (and dependency loops). I'll see if there's some way to hoist this into `@accordproject/concerto-util`

Issue #869, opened 2024-06-25
Update concerto-analysis rules to analyse aliased imports


ekarademir:
Related analysers are probably these
* Properties 
* Class declarations that extend an imported type
* ~Import statements~

ekarademir:
* Is aliasing just a syntactic change? So should it be ignored in analysis?
* Or should it at least cause a patch change

Are the following the same
```js
import com.example.foo@1.0.0.{Bar}

concept Animal {
  o Bar leg
}
```

```js
import com.example.foo@1.0.0.{Bar as baz}

concept Animal {
  o baz leg
}
```

salujajaskeerat:
The current version of concerto-analysis only analyses the semantic structure of a model file. Whereas aliasing a type is only a syntactic change.


Also, there is no import factory at present. 

Handling the above would require us to create an import factory.


**Pros of analysing aliased imports:**
- we can label the alias changes as patch changes; this would explicitly tell the user that he has made an alias change. 

**cons of analysing aliased imports**
- aliasing a type is only a syntactic change, it does not change anything semantically

ekarademir:
Our BDFL decided this will be a patch change.

PR #894, opened 2024-08-28
feat(import-alias):patch change for aliased imported types

salujajaskeerat:

This pull request deals with analyser rules for aliased types in concerto. 

### Related Issues
- Issue #869 

### Author Checklist

### DIFF
--- a/packages/concerto-analysis/src/comparers/properties.ts
+++ b/packages/concerto-analysis/src/comparers/properties.ts
@@ -183,11 +183,11 @@ const propertyTypeChanged: ComparerFactory = (context) => ({
         }
         // Up to this point, 'a' and 'b' have identical fully qualified type names with matching versions.
         // Next, we verify whether their types differ locally, which would indicate aliasing.
-        if(a.getType()!=b.getType()){
+        if (a.getType() !== b.getType()) {
             context.report({
-                key:'property-type-aliased',
-                message:`The local type name for "${a.getName()}" in the ${classDeclarationType} "${a.getParent().getName()}" is changed from ${a.getType()} to ${b.getType()}`,
-                element:a
+                key: 'property-type-aliased',
+                message: `The local type name for "${a.getName()}" in the ${classDeclarationType} "${a.getParent().getName()}" is changed from ${a.getType()} to ${b.getType()}`,
+                element: a
             });
             return;
         }
--- a/packages/concerto-analysis/src/comparers/properties.ts
+++ b/packages/concerto-analysis/src/comparers/properties.ts
@@ -181,9 +181,9 @@ const propertyTypeChanged: ComparerFactory = (context) => ({
             });
             return;
         }
-        // Up to this point, 'a' and 'b' have identical fully qualified type names with matching versions.
+        // Up to this point, 'a' and 'b' have identical fully qualified type names with matching major and premajor versions.
         // Next, we verify whether their types differ locally, which would indicate aliasing.
-        if (a.getType() !== b.getType()) {
+        if (!versionDiff && (a.getType() !== b.getType())) {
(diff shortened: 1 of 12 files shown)


salujajaskeerat:
@mttrbrts @ekarademir Kindly review the PR

packages/concerto-analysis/src/comparers/properties.ts
```diff
@@ -173,16 +173,24 @@ const propertyTypeChanged: ComparerFactory = (context) => ({
...
             context.report({
                 key: 'property-type-changed',
                 message: `The ${aType} "${a.getName()}" in the ${classDeclarationType} "${a.getParent().getName()}" changed type from "${aFQTN}" to "${bFQTN}" (type version incompatible)`,
                 element: a
             });
             return;
         }
+        // Up to this point, 'a' and 'b' have identical fully qualified type names with matching versions.
+        // Next, we verify whether their types differ locally, which would indicate aliasing.
+        if (a.getType() !== b.getType()) {
```
ekarademir, line 186:
```suggestion
        if (!versionDiff && a.getType() !== b.getType()) {
```
Since you are removing this condition to check where version hasn't changed, we need to cover it here I think. Because after
```js
(versionDiff === 'major' || versionDiff === 'premajor')
```
check, there are still `patch` and `minor` changes.

salujajaskeerat: Do you mean if there are `minor` or `patch` version  changes , then we should not report the `localtype` changes ?

ekarademir: What I mean is the logic you are adding should not add side effects. With the current change you are not JUST checking the aliasing, but also, minor and patch changes and, you are issuing a `patch` change as a result, for both cases.

If you want to handle minor and patch changes as well, they need to be called out explicitly, with a different rule, separate from aliasing, and before the aliasing check. This is how it's done for major and premajor changes.

Up to now minor and patch changes didn't issue any analysis result. So adding them would require another discussion and would be a breaking change IMO.

ekarademir: Also in your comment you are indicating that `Up to this point, 'a' and 'b' have identical fully qualified type names with matching versions.` But FQNs would be different if there are minor and patch changes.

salujajaskeerat: Ok I understand your point you are saying . But there is one issue with that, if there is a minor or patch change in the version and also some local type changes , then the logic you are saying won't report anything to the user. 
I understand completely what you meant previously that there should be specific rule for that.

salujajaskeerat: Will update the comments as well. 
I think we need a discussion on the minor and patch changes of the version. I feel even if they are not reported we should still report the localType changes . 
We can discuss this in the TechWg call

packages/concerto-core/lib/basemodelmanager.js
```diff
@@ -98,7 +98,7 @@ class BaseModelManager {
         // TODO Remove on release of MapType
         // Supports both env var and property based flag
         this.enableMapType = !!options?.enableMapType;
-        this.importAliasing = !!options?.importAliasing;
+        this.importAliasing = process.env.IMPORT_ALIASING || !!options?.importAliasing;
```
ekarademir, line 101:
```suggestion
        this.importAliasing = process?.env?.IMPORT_ALIASING === 'true' || !!options?.importAliasing;
```

salujajaskeerat: Done

ekarademir (approved):

PR #901, opened 2024-09-07
fix(introspect): redefining imported type (#871)

### LINKED ISSUE #871: Redefining an imported type is allowed
# Bug Report 🐛

When I try to redefine an imported type in the file, it should throw an error as there should be a name conflict. 
For example, below code : 

```js
namespace org.accordproject.color@1.0.0

import org.test.color@1.0.0.{doc} 

concept doc {
    o String detail
}

concept files{
    o doc[]  documents
}
```

Here, I am trying to redefine the concept doc imported above. Should this be allowed?

## Expected Behavior

It should throw an error stating the name conflict. As mentioned above, local type doc has two definitions, i.e. `org.test.color@1.0.0.doc` and `org.accordproject.color@1.0.0.doc`

## Current Behavior

The code gets compiled successfully. Inside the `concept files`, the type `doc` gets resolved to `org.test.color@1.0.0.doc`, which is from the import statement.

## Possible Solution

In `concerto/core` inside the model file validation, we test if the declaration has unique names or not. We don't check with the imported types there. Just checking there with the imports could fix this issue.

Check the validation code here. 

https://github.com/accordproject/concerto/blob/9e61044925e54b72bfdc05a23a15a5062c1ac53d/packages/concerto-core/lib/introspect/modelfile.js#L276

salujajaskeerat:
# Closes #871 

### Related Issues
- Issue #871 

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/introspect/modelfile.js
+++ b/packages/concerto-core/lib/introspect/modelfile.js
@@ -275,13 +275,14 @@ class ModelFile extends Decorated {
         // Validate all of the types in this model file.
         // Check if names of the declarations are unique.
         const uniqueNames = new Set();
+        const concertoTypes = new Set(['Concept', 'Asset', 'Transaction', 'Participant', 'Event']);
         this.declarations.forEach(d => {
             const fqn = d.getFullyQualifiedName();
             const shortName = ModelUtil.getShortName(fqn);
             if (uniqueNames.has(fqn)) {
                 throw new IllegalModelException(`Duplicate class name: ${fqn}`);
             }
-            if (this.importShortNames.has(shortName)) {
+            if (!concertoTypes.has(shortName) && this.importShortNames.has(shortName)) {
                 throw new IllegalModelException(
                     `TypeName Conflict: The type '${shortName}' is being redefined, but it already exists as an imported type from '${this.importShortNames.get(shortName)}'. Please rename your local type or avoid importing conflicting types.`
                 );
--- a/packages/concerto-core/lib/introspect/modelfile.js
+++ b/packages/concerto-core/lib/introspect/modelfile.js
@@ -275,14 +275,13 @@ class ModelFile extends Decorated {
         // Validate all of the types in this model file.
         // Check if names of the declarations are unique.
         const uniqueNames = new Set();
-        const concertoTypes = new Set(['Concept', 'Asset', 'Transaction', 'Participant', 'Event']);
         this.declarations.forEach(d => {
             const fqn = d.getFullyQualifiedName();
             const shortName = ModelUtil.getShortName(fqn);
             if (uniqueNames.has(fqn)) {
                 throw new IllegalModelException(`Duplicate class name: ${fqn}`);
             }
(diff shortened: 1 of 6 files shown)


packages/concerto-core/lib/introspect/modelfile.js
```diff
@@ -275,19 +275,20 @@ class ModelFile extends Decorated {
...
-            d => {
-                const fqn = d.getFullyQualifiedName();
-                if (!uniqueNames.has(fqn)) {
-                    uniqueNames.add(fqn);
-                } else {
-                    throw new IllegalModelException(
-                        `Duplicate class name ${fqn}`
-                    );
-                }
+        const concertoTypes = new Set(['Concept', 'Asset', 'Transaction', 'Participant', 'Event']);
```
muhabdulkadir, line 278: What happens when more concertoTypes are created?

packages/concerto-core/lib/introspect/modelfile.js
```diff
@@ -275,19 +275,20 @@ class ModelFile extends Decorated {
...
+            if (uniqueNames.has(fqn)) {
+                throw new IllegalModelException(`Duplicate class name: ${fqn}`);
             }
-        );
-
+            if (!concertoTypes.has(shortName) && this.importShortNames.has(shortName)) {
+                throw new IllegalModelException(
+                    `TypeName Conflict: The type '${shortName}' is being redefined, but it already exists as an imported type from '${this.importShortNames.get(shortName)}'. Please rename your local type or avoid importing conflicting types.`
+                );
+            }
```
muhabdulkadir, line 289: What would be the short name if this was an aliased type? is that accounted for?

packages/concerto-core/lib/introspect/modelfile.js
salujajaskeerat: I checked if there is any utility function in Model.Util which checks if the type is a concerto-type or not . There is no such function currently . But yes it would be better if have one function which manages this and we register them here

packages/concerto-core/lib/introspect/modelfile.js
salujajaskeerat: Yes it is accounted for . 
imporShortNames is map from local type to fqn . So if it was a aliased name then it is registered in the importShortNames map

packages/concerto-core/lib/introspect/modelfile.js
muhabdulkadir: Let's remove the dependency on `concertoTypes` as we cannot keep track of this list of items.

salujajaskeerat: Remove the dependency on the concerto-type. Also updated the breaking test-cases

dselman (changes requested):
Why is system.Event etc being renamed?

salujajaskeerat:
> Why is system.Event etc being renamed?

Because Event is a concerto type , in the test-case we had a name conflict with Event type being redefined

PR #952, opened 2024-12-02
fix : Modify code generation script to include DCS Concerto model types

### LINKED ISSUE #943: Type definitions for DCS
# Bug Report 🐛

Type definitions for a DCS model are missing. Add generated typescript types of DCS Concerto Model to concerto-types module.

## Expected Behavior

Type of DCS object should be centrally defined in concerto-types.

## Possible Solution

We'll require a code generation script similar to [this](https://github.com/accordproject/concerto/blob/main/packages/concerto-types/scripts/codegen.js) that will convert the [DCS concerto model](https://github.com/accordproject/concerto/blob/3dac62f4690a81fa1754b9d9116cba869e22ac3d/packages/concerto-core/lib/decoratormanager.js#L37) to its corresponding typescript definitions.

Ayush1404:
# Closes #943 

Modifed code generation script to include DCS Concerto model types
### Changes

- <ONE>Modify code generation script to include DCS Concerto model types

### Related Issues
- Issue #943

### DIFF
--- a/packages/concerto-core/lib/decoratormanager.js
+++ b/packages/concerto-core/lib/decoratormanager.js
@@ -812,4 +812,4 @@ class DecoratorManager {
     }
 }
 
-module.exports = DecoratorManager;
+module.exports = {DecoratorManager,DCS_MODEL};
--- a/packages/concerto-core/lib/decoratormanager.js
+++ b/packages/concerto-core/lib/decoratormanager.js
@@ -812,4 +812,5 @@ class DecoratorManager {
     }
 }
 
-module.exports = {DecoratorManager,DCS_MODEL};
+module.exports = DecoratorManager;
+module.exports.DCS_MODEL = DCS_MODEL;
--- a/packages/concerto-types/scripts/codegen.js
+++ b/packages/concerto-types/scripts/codegen.js
@@ -4,7 +4,7 @@ const { ModelManager } = require('@accordproject/concerto-core');
 const { CodeGen: { TypescriptVisitor }} = require('@accordproject/concerto-codegen');
 const { FileWriter } = require('@accordproject/concerto-util');
 const path = require('path');
-const {DCS_MODEL }= require('@accordproject/concerto-core/lib/decoratormanager');
+const { DCS_MODEL }= require('@accordproject/concerto-core');
 
 /**
  * Generate TypeScript files from the metamodel.
--- a/packages/concerto-types/scripts/codegen.js
+++ b/packages/concerto-types/scripts/codegen.js
@@ -4,12 +4,14 @@ const { ModelManager } = require('@accordproject/concerto-core');
 const { CodeGen: { TypescriptVisitor }} = require('@accordproject/concerto-codegen');
 const { FileWriter } = require('@accordproject/concerto-util');
 const path = require('path');
+const {DCS_MODEL }= require('@accordproject/concerto-core/lib/decoratormanager');
 
 /**
  * Generate TypeScript files from the metamodel.
  */
 async function main() {
     const modelManager = new ModelManager({addMetamodel:true, strict: true});
+    modelManager.addCTOModel(DCS_MODEL, 'decoratorcommands@0.4.0.cto');
     const visitor = new TypescriptVisitor();
 
     const fileWriter = new FileWriter(path.resolve(__dirname, '..', 'src', 'generated'));
--- a/packages/concerto-types/src/generated/unions/concerto@1.0.0.ts
+++ b/packages/concerto-types/src/generated/unions/concerto@1.0.0.ts
@@ -26,6 +26,14 @@ import type {
 	IModel,
 	IModels
 } from './concerto.metamodel@1.0.0';
+import type {
+	IDecoratorCommandSetReference,
+	CommandType,
+	ICommandTarget,
+	MapElement,
+	ICommand,
+	IDecoratorCommandSet
+} from './org.accordproject.decoratorcommands@0.4.0';
 
 // interfaces
 export interface IConcept {
@@ -51,7 +59,11 @@ ILongDomainValidator |
 IAliasedType | 
 IImport | 
 IModel | 
-IModels;
+IModels | 
+IDecoratorCommandSetReference | 
+ICommandTarget | 
+ICommand | 
+IDecoratorCommandSet;
 
 export interface IAsset extends IConcept {
    $identifier: string;
--- a/packages/concerto-types/src/generated/unions/org.accordproject.decoratorcommands@0.4.0.ts
+++ b/packages/concerto-types/src/generated/unions/org.accordproject.decoratorcommands@0.4.0.ts
@@ -1,0 +1,47 @@
+/* eslint-disable @typescript-eslint/no-empty-interface */
+// Generated code for namespace: org.accordproject.decoratorcommands@0.4.0
+
+// imports
+import {IDecorator} from './concerto.metamodel@1.0.0';
+import {IConcept} from './concerto@1.0.0';
+
+// interfaces
+export interface IDecoratorCommandSetReference extends IConcept {
+   name: string;
+   version: string;
+}
+
+export enum CommandType {
+   UPSERT = 'UPSERT',
+   APPEND = 'APPEND',
+}
+
+export interface ICommandTarget extends IConcept {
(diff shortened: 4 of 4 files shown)


packages/concerto-types/scripts/codegen.js
```diff
@@ -4,12 +4,14 @@ const { ModelManager } = require('@accordproject/concerto-core');
...
 const { FileWriter } = require('@accordproject/concerto-util');
 const path = require('path');
+const DCS_MODEL = require('@accordproject/concerto-core/lib/decoratormanager').DCS_MODEL;
 
 /**
  * Generate TypeScript files from the metamodel.
  */
 async function main() {
     const modelManager = new ModelManager({addMetamodel:true, strict: true});
+    modelManager.addCTOModel(DCS_MODEL, 'decoratorcommands@0.4.0.cto');
```
sanketshevkar, line 14: Can you please run `npm run codegen` and also commit the generated type files to this PR?

sanketshevkar (changes requested):

Ayush1404:
@sanketshevkar hey , i had to make little changes to make it work , but i have pushed the commits after running codegen script

packages/concerto-types/src/generated/unions/concerto@1.0.0.ts
```diff
@@ -26,6 +26,14 @@ import type {
 	IModel,
 	IModels
 } from './concerto.metamodel@1.0.0';
+import type {
```
sanketshevkar, line 29: Why is this import required?

packages/concerto-types/src/generated/unions/concerto@1.0.0.ts
```diff
@@ -51,7 +59,11 @@ ILongDomainValidator |
 IAliasedType | 
 IImport | 
 IModel | 
-IModels;
+IModels |
```
sanketshevkar, line 62: Is this autogenerated or you've manually added?

Ayush1404: It is auto generated

packages/concerto-types/src/generated/unions/concerto@1.0.0.ts
Ayush1404: It is autogenrated so i dont know exact reason , but thoze types are used below

packages/concerto-types/scripts/codegen.js
```diff
@@ -4,12 +4,14 @@ const { ModelManager } = require('@accordproject/concerto-core');
 const { CodeGen: { TypescriptVisitor }} = require('@accordproject/concerto-codegen');
 const { FileWriter } = require('@accordproject/concerto-util');
 const path = require('path');
+const {DCS_MODEL }= require('@accordproject/concerto-core/lib/decoratormanager');
```
sanketshevkar, line 7: @mttrbrts is there any better way to import the DCS model?

sanketshevkar: Should we just hard code the model here in `concerto-types`?

mttrbrts: I think that this is reasonable for now given that this is the home for the DCS model. However, I would align better with the metamodel to move the model into the @accordproject/concerto-metamodel package, and import from there.

Let's open a separate task for that, given that it would be a breaking change (unless we do some kind of proxying to move it transparently).

mttrbrts (approved):

Ayush1404:
@sanketshevkar hey , sorry the build was failing cause i changed the exports , i have fixed them in the recent commit . What do i do next ?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -813,3 +813,4 @@ class DecoratorManager {
 }
 
 module.exports = DecoratorManager;
+module.exports.DCS_MODEL = DCS_MODEL;
```
sanketshevkar, line 816:
```suggestion
module.exports = { DCS_MODEL, DecoratorManager };
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -813,3 +813,4 @@ class DecoratorManager {
 }
 
 module.exports = DecoratorManager;
```
sanketshevkar, line 815:
```suggestion
module.exports = { DCS_MODEL, DecoratorManager };
```

packages/concerto-core/lib/decoratormanager.js
sanketshevkar, line 816:
```suggestion
```

packages/concerto-types/scripts/codegen.js
sanketshevkar, line 7:
```suggestion
const { DCS_MODEL }= require('@accordproject/concerto-core');
```

sanketshevkar (changes requested):

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Can you please export this using the [index.js](https://github.com/accordproject/concerto/blob/8997f3ce9f240b3d71471b5c78166e2e6a7e9c96/packages/concerto-core/index.js#L94)?

Ayush1404: But doing so will give an error as DecoratorManager is supposed to be exported default in tests that are written

packages/concerto-core/lib/decoratormanager.js
Ayush1404: if we export this like this
`module.exports = { DCS_MODEL, DecoratorManager };`
then the DecoratorManager isn't exported as default which is expected by tests so tests will fail .

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: Hmmm, I'm wondering we could just fix the tests with imports, but could this could break for someone whose not importing through `index.js`. Let me discuss this during today's WG call.

packages/concerto-types/scripts/codegen.js
```diff
@@ -4,12 +4,14 @@ const { ModelManager } = require('@accordproject/concerto-core');
 const { CodeGen: { TypescriptVisitor }} = require('@accordproject/concerto-codegen');
 const { FileWriter } = require('@accordproject/concerto-util');
 const path = require('path');
+const { DCS_MODEL }= require('@accordproject/concerto-core');
```
sanketshevkar, line 7: You'll still have to export it via `index.js`. I don't think this codegen would work without it.

sanketshevkar (changes requested):

packages/concerto-core/lib/decoratormanager.js
sanketshevkar: I just discussed, we decided to move `DCS_MODEL` to `concerto-metamodel` package. I'll raise a PR and cut a release and update it here to be used.

Can you please wait until I move things there?

Ayush1404: @sanketshevkar yeah sure , no problem . I would like to contribute , could you direct me to something where i could ?

sanketshevkar: Thanks @Ayush1404 , you'll have to move the DCS model in this [repo](https://github.com/sanketshevkar/concerto-metamodel) and import it in core and types package in concerto.

Ayush1404: @sanketshevkar , okay do i do it here in this PR or open a separate issue for it ? Also i meant to ask where else could i contribute to concerto or accord project in general .

sanketshevkar:
Hey @Ayush1404,
Couldn't catchup on this issue. Will you be still working on this?

Ayush1404:
@sanketshevkar Hey , sorry for the late reply .Yes i will still be working on this but i am confused about what to do and also should i do it in the same PR or open a seperate issue for it

sanketshevkar:
Issue should stay the same, but you'll have to open up a new PR in concerto-metamodel repo. Move the DCS model definition there from `DecoratorManager` in concerto-core. And import it in `DecoratorManager`.

sanketshevkar:
@Ayush1404 are you working on this?

PR #953, opened 2024-12-03
fix(package): always use latest published version "TEST"

muhabdulkadir:

### Changes

- resolve concert intra repo dependencies to always use latest npm published package

### Author Checklist

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,0 +1,4 @@
+{
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
+  "version": "3.19.9"
+}
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,4 @@
 {
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
-  "version": "3.19.9"
+  "version": "3.19.10"
 }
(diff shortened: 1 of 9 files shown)


packages/concerto-core/package.json
```diff
@@ -64,9 +64,9 @@
     "yargs": "17.7.2"
   },
   "dependencies": {
-    "@accordproject/concerto-cto": "file:../concerto-cto",
+    "@accordproject/concerto-cto": "*",
```
sanketshevkar, line 67: Are we sure we want this? In the future when we release v4, and if some want to install `concerto-core@3.19.9`, won't this forcefully install `concerto-util@4.0.0` and `concerto-cto@4.0.0` which themselves could have a breaking change and thus breaking code for `concerto-core@3.19.9`?

muhabdulkadir: This is a test.

lerna.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
```
sanketshevkar, line 2: We've moved to npm workspaces I believe.

sanketshevkar (changes requested):

lerna.json
sanketshevkar: We don't use lerna anymore.

sanketshevkar: https://github.com/accordproject/concerto/pull/912

PR #977, opened 2025-01-13
chore(aliasing): tests import aliasing

### LINKED ISSUE #972: APIs from concerto-core are breaking for aliased models whose types are resolved.
# Bug Report 🐛

The implementation for the resolution for the previous similar [issue](https://github.com/accordproject/concerto/issues/950) is leading to a few APIs in `concerto-core` and `concerto-cto` breakiing when they try to refer to the resolved type reference for the imported-aliased declarations.

Main issue is that when a model AST is resolved, we loose the reference to the aliased type names instead have the resolved type name, which was replaced.

## Expected Behavior

Resolved model should preserve both the resolved type names and the original type name as specified by the user.

## Current Behavior

Resolved model only replaces the type names used by the user with resolved type names (taking into account aliases).

## Steps to Reproduce

1. Model Decoration
If we try to decorate these models with given DCS, the process fails with error.
```
namespace accord.standard@1.0.0

concept BaseNDA {
    o Date executiondate
}

             +

namespace acme.legal.@1.0.0

import accord.standard@1.0.0.{BaseNDA as NewNDA}

concept BaseNDA extends NewNDA {
    o String person1
}

            +

{
    "commandSet": [
        {
            "type": "UPSERT",
            "target": {
              "namespace": "namespace acme.legal.@1.0.0",
              "declaration": "BaseNDA",
              "$class": "org.accordproject.decoratorcommands@0.3.0.CommandTarget"
            },
            "decorator": {
              "name": "TestMe",
              "arguments": [],
              "$class": "concerto.metamodel@1.0.0.Decorator"
            },
            "$class": "org.accordproject.decoratorcommands@0.3.0.Command"
          }
    ]
}

            =

    Error
```
2. Printing a resolved model from AST, where an declaration extending an aliased declaration would break
If we try to load these models, validate, resolve and print these models, the printing step would result in an error.
```
namespace accord.standard@1.0.0

concept BaseNDA {
    o Date executiondate
}

         +

namespace acme.legal.@1.0.0

import accord.standard@1.0.0.{BaseNDA as NewNDA}

concept BaseNDA extends NewNDA {
    o String person1
}

        =

Error
```
3. Print a resolved model from AST, where a property instantiated using an aliased declaration would change to the base declaration rather than the aliased declaration.
if we try to load these models, validate, resolve and print these models, the processes are successful but we loose reference to the aliased type which is imported in the property instantiation.
```
namespace accord.standard@1.0.0

concept BaseNDA {
    o Date executiondate
}

        +

namespace acme.legal.@1.0.0

import accord.standard@1.0.0.{BaseNDA as NewNDA}

concept NDA {
    o NewNDA newNDA
}

      =

import accord.standard@1.0.0.{BaseNDA as NewNDA}

concept NDA {
    o BaseNDA newNDA
}

```

### Desktop
 - OS: [e.g. macOS]
 - Version: [e.g. 0.22.15]

## Detailed Description

The issue is that `resolveMetamodel` method (`called by ModelManager.toAst(true)`) was modified to replace aliased imports with the actual type names inside  TypeIdentifiers. This is useful for users processing the AST however it is lossy as it is difficult to recover the name as originally used by the user in the CTO file. In a subsequent call to `fromAst` the  resolved AST is used:
1. If the resolved type was not imported then a type not found error will occur
2. If the resolved imported type has the same name as a type  in the current namespace then a conflict will occur, leading to undefined behavior
3. If the resolved imported type is used a super class for a class with the same name, then a conflict will occur
4. Printer.toCTO has not been updated to recover the aliased name, so the roundtrip from [CTO with aliased import -> Resolved AST -> CTO] will be lossy, with aliased type names replaced with resolved type names

### ModelManager.getAst(true) should return: 
   1. A valid instance of the meta model 
   2. Namespaces should be added to TypeIdentifiers so that imports do not need to be inspected
   3. Set the `resolvedName` property to be the name of the type in its host namespace, irrespective of whether aliasing was used for the type or not.
   4. The `name` property should be the type name as specified in the CTO file

To get a valid fully-qualified name for a type identifier the caller should use `${namespace}.${resolvedName}` — this ensures that aliasing is taken into account.

### ModelManager.fromAst(ast) should return:

> No changes here. Undo the changes to fromAst.

### Printer.toCTO(ast) should return:

> No changes here. The name property should be used.

### Metamodel update

   1. Add an optional `resolvedName` to TypeIdentifier
   2. Republish concerto-metamodel
   3. Update models.accordproject.org

### Unit Tests

Alias an import and use wherever a type can be referenced:
- Super type
- Property
- Decorator
- Map key
- May value 

```
namespace child@1.0.0
concept Child {}
scalar MyString extends String
```

```
@test(Kid)
namespace parent@1.0.0

@test(Kid)
import child@1.0.0.{MyString as Str, Child as Kid}

@test(Kid)
concept Child extends Kid {
   @test(Kid)
   o Kid kid
   o Str str
}

@test(Kid)
map KidIndex {
   o Str
   @test(Kid)
   o Kid
}
```

Test that `parent@1.0.0` CTO file roundtrips through getAst(true) and fromAst(ast) and can be used with decorator command set.

sanketshevkar:
# Closes #972 

### Changes

Added to new optionalresolvedName property to TypeIdenifier. This is to preserve the aliased name property which will have the local name for the aliased type. resolvedName will contain the origonal type name. It will get added only for the declrations ot types that "imported and aliased".

With the current implementation the value of the name property is different for a normal ast (local name) vs a resolved ast (original type name). This can be very confusing.

The above changes have been released in `concerto-metamodel@3.11.0`.

This PR mostly focuses on adding the unit tests to concerto apis to reinforce the import aliasing feature.

Tests for the following have been added specifically targeting import aliasing:
- ClassDeclaration
- Property
- Decorator
- Map-Key
- Map-Value
- CTO-ResolvedAST-CTO, a test to confirm that the round trip from CTO to CTO is successfull and produces exactly the same model.

### Author Checklist


packages/concerto-core/test/introspect/scalardeclaration.js
```diff
@@ -304,3 +308,64 @@ describe('ScalarDeclaration', () => {
...
+        const resolvedModelFileChild = new ModelFile(resolvedModelManager, resolvedMetamodelChild, 'child@1.0.0.cto');
+        const resolvedModelFileParent = new ModelFile(resolvedModelManager, resolvedMetamodelParent, 'parent@1.0.0.cto');
+        resolvedModelManager.addModelFiles([resolvedModelFileChild, resolvedModelFileParent], ['child@1.0.0.cto', 'parent@1.0.0.cto']);
+    });
+
+    // describe.only('#getName', () => {
+    //     it('should return the scalar name', () => {
+    //         // This fails, current imports are not declared within a namespace. Is this a bug?
+    //         // Why are we treating imported declrations as foreign objects? Shouldn't they be treated as part of the namespace and declared?
+    //         // Extend this test to map keys and value as well
```
sanketshevkar, line 337: @mttrbrts @dselman can y'all please review this comment?

packages/concerto-core/lib/introspect/property.js
```diff
@@ -109,6 +109,7 @@ class Property extends Decorated {
             this.type = 'String';
             break;
         case `${MetaModelNamespace}.ObjectProperty`:
+            // should we have resolved name here?
             this.type = this.ast.type ? this.ast.type.name : null;
```
sanketshevkar, line 113: Should we expose `resolveName` through the `decorated` instances?

Or is `resolvedName` "only" for downstream systems directly consuming and manipulating the AST?

packages/concerto-core/test/introspect/scalardeclaration.js
sanketshevkar: For the following model
```
namespace car@1.0.0

import person@1.0.0.Driver

concept Driver extends Driver {}
```
This will throw the error that `Drive cannot extend itself` instead of `Driver is already declared.`, we error on similar lines in JS classes.

sanketshevkar: Also with respect to this test, getting the type's classdeclaration from model manager is a bit tricky because, `MyString` is in `child@1.0.0`.

And used in `parent@1.0.0.Child.str` where its aliased as `Str`. `getType` on the property only returns a string value as `Str`. Shouldn't it return the ClassDeclration object.

packages/concerto-core/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "dependencies": {
     "@accordproject/concerto-cto": "3.20.1",
-    "@accordproject/concerto-metamodel": "3.10.4",
+    "@accordproject/concerto-metamodel": "../../../concerto-metamodel",
```
mttrbrts, line 68: Be careful not to commit this change

packages/concerto-core/lib/introspect/declaration.js
```diff
@@ -65,6 +65,7 @@ class Declaration extends Decorated {
             throw new IllegalModelException(`Invalid class name '${this.ast.name}'`, this.modelFile, this.ast.location);
         }
 
+        // should we use resolved name here?
         this.name = this.ast.name;
         this.fqn = ModelUtil.getFullyQualifiedName(this.modelFile.getNamespace(), this.name);
```
mttrbrts, line 70:
```suggestion
        // should we use resolved name here?
        this.name = this.ast.name;
        this.fqn = ModelUtil.getFullyQualifiedName(this.modelFile.getNamespace(), this.ast.resolvedName);
```

I think that the value of this.name should stay consistent, but we should fix the value of fqn

packages/concerto-core/lib/introspect/property.js
mttrbrts: Is the value of this.type valid for an aliased name?

packages/concerto-core/test/introspect/decorator.js
```diff
@@ -76,3 +82,46 @@ describe('Decorator', () => {
...
+        modelManager.addCTOModel(parentModelCTO, 'parent@1.0.0.cto');
+        const resolvedMetamodelChild = modelManager.resolveMetaModel(modelManager.getAst().models[0]);
+        const resolvedMetamodelParent = modelManager.resolveMetaModel(modelManager.getAst().models[1]);
+        resolvedModelManager = new ModelManager({ strict: true, importAliasing: true, enableMapType: true});
+        const resolvedModelFileChild = new ModelFile(resolvedModelManager, resolvedMetamodelChild, 'child@1.0.0.cto');
+        const resolvedModelFileParent = new ModelFile(resolvedModelManager, resolvedMetamodelParent, 'parent@1.0.0.cto');
+        resolvedModelManager.addModelFiles([resolvedModelFileChild, resolvedModelFileParent], ['child@1.0.0.cto', 'parent@1.0.0.cto']);
+    });
+
+    // describe.only('#getArguments', () => {
```
mttrbrts, line 107: Why is this commented out?

packages/concerto-core/test/introspect/scalardeclaration.js
mttrbrts: What is the expected response in the test? and what is the actual value?

mttrbrts (changes requested):

packages/concerto-core/lib/introspect/declaration.js
sanketshevkar: fqn is here is the local fqn for the cto file, right?

```
namespace child

concept Child {}
```
```
namespace parent

import child.{Child as Kid} 

concept Children extends Kid {}
```

The suggest code would result the fqn to be `parent.Child`, instead of `parent.Children`

packages/concerto-core/test/introspect/decorator.js
sanketshevkar: `getArguments` would result 
```
{ type: 'Identifier', name: 'Kid', array: false }
```

I'm not sure if test adds any value. I thought we might have something like arguments[0].getSuperType() and it could return `child@1.0.0.Child` or getType.

packages/concerto-core/test/introspect/scalardeclaration.js
sanketshevkar:
```
namespace child@1.0.0
concept Child {
    o Integer age
}
scalar MyString extends String

```
```
@test(Kid)
namespace parent@1.0.0

import child@1.0.0.{MyString as Str, Child as Kid}

@test(Kid)
concept Child extends Kid {
   @test(Kid)
   o Kid kid
   o Str str
}

concept Student extends Child {
   o Str schoolName
   o Str name
   o Child myChild
}

@test(Kid)
map KidIndex {
   o Str
   @test(Kid)
   o Kid
}
```

From `parent.Child.str` I want to get the scalarDeclration object `Str`, on which I want to invoke the `getName`, `getNamspace` methods and test them.

Since `Str` is not an is imported, we never declare it in the namesapce, because of which we cannot fetch it scalarDeclration object.

sanketshevkar: I'll log an issue to discuss this. This is unrelated to the changes here.

packages/concerto-core/package.json
sanketshevkar: Updated

packages/concerto-core/test/introspect/decorator.js
sanketshevkar: Decoration validation test added to aliased types as decorator arguments.

scripts/coverage.js
```diff
@@ -33,9 +33,9 @@ function copyFiles(files, destDir) {
 }
 
 const lcovs = glob.sync(globPattern).map((dir) => {
-    const packageName = dir.split('/').pop();
+    const packageName = dir.split(path.sep).join('/').split('/').pop();
```
sanketshevkar, line 36: `glob` had to be updated to fix an issue in [rimraf](https://github.com/isaacs/rimraf/issues/306), which was causing `concerto-types` build to fail.

`glob` itself had a breaking change for the `sync` [API](https://github.com/isaacs/node-glob/issues/480) a few releases back because of which coverge reprting was failing for windows environment in the PR build pipeline.

So I had to fix the script.

packages/concerto-core/test/data/aliasing/parent.json
```diff
@@ -0,0 +1,498 @@
...
+            "end": {
+              "offset": 162,
+              "line": 11,
+              "column": 1,
+              "$class": "concerto.metamodel@1.0.0.Position"
+            }
+          }
+        }
+      ],
+      "location": {
```
ekarademir, line 175: Could you remove the locations from the ast please?

packages/concerto-core/test/introspect/decorator.js
```diff
@@ -76,3 +82,34 @@ describe('Decorator', () => {
...
+        const resolvedModelFileChild = new ModelFile(resolvedModelManager, resolvedMetamodelChild, 'child@1.0.0.cto');
+        const resolvedModelFileParent = new ModelFile(resolvedModelManager, resolvedMetamodelParent, 'parent@1.0.0.cto');
+        resolvedModelManager.addModelFiles([resolvedModelFileChild, resolvedModelFileParent], ['child@1.0.0.cto', 'parent@1.0.0.cto']);
+    });
+
+    describe('#validate', () => {
+
+        it('should be able get validate a decorator whose argument is an imported type which is aliased', () => {
+            const classDeclaration = resolvedModelManager.getType('parent@1.0.0.Child');
+            const decorator = classDeclaration.getDecorators()[0];
```
ekarademir, line 111: Could you also check the decorator for the whole model?

ekarademir (approved):

packages/concerto-core/test/data/aliasing/parent.json
sanketshevkar: [This](https://github.com/accordproject/concerto/blob/cc54de136c4f01d918182c3b09f9858720c71ece/packages/concerto-core/test/introspect/metamodel.js#L215) test is doing a doing a check on the parsed cto model with the ast.

packages/concerto-core/test/introspect/decorator.js
sanketshevkar: Yes works!
Added a new test to validate that.

packages/concerto-core/test/data/aliasing/parent.cto
```diff
@@ -0,0 +1,24 @@
+@test(Kid)
+namespace parent@1.0.0
+
+import child@1.0.0.{MyString as Str,Child as Kid}
```
mttrbrts, line 4: Perhaps it's jet lag, but I find these examples hard to parse. Let's give the identifiers more descriptive names to make the tests easier to understand.

mttrbrts (approved):

packages/concerto-core/test/data/aliasing/parent.cto
sanketshevkar: Done!

PR #986, opened 2025-02-04
Update the ModelManager to handle multiple metamodel versions.

### LINKED ISSUE #556: Fully support metamodel versioning
Metamodel versioning has been introduced, but the entire codebase is hardcoded to `concerto.metamodel@1.0.0`, and so we can only support metamodels using this specific version. This limits us when making metamodel changes, as it is not currently possible to deliver a new metamodel version into Concerto and have Concerto still process existing metamodels that use the old version.

This feature should extend Concerto such that we can support existing `1.0.0` version metamodels and a new `1.1.0` version metamodel at the same time, as well as defining the strategy for how we eventually support a new `2.0.0` version. Will we support both `1.x` and `2.x` metamodels at the same time, or will we introduce a new major version of Concerto?

Atul-khadse:
# Closes #556 

- This PR adds metamodel versioning support to the Concerto codebase, allowing for the validation and processing of multiple metamodel versions (1.0.0 and 1.1.0). It introduces a dynamic versioning strategy to ensure backward compatibility and prepare for future versions.  

### Changes

- Added metamodelVersions array to ModelManager to store supported metamodel versions.
- Implemented validateMetamodelVersion method to check the version before adding a model.
- Refactored addModel to extract the version from ast.$class and validate it.
- Enhanced error messages for unsupported versions.

### Flags

- Future versions (2.x) will require a separate strategy to ensure backward compatibility.
- This PR only supports 1.0.0 and 1.1.0 for now

### Related Issues
- Issue #556

### DIFF
--- a/packages/concerto-core/lib/modelmanager.js
+++ b/packages/concerto-core/lib/modelmanager.js
@@ -70,6 +70,7 @@ class ModelManager extends BaseModelManager {
      */
     constructor(options) {
         super(options, ctoProcessFile(options));
+        this.metamodelVersions = ['1.0.0', '1.1.0']; // Add supported metamodel versions here
     }
 
     /**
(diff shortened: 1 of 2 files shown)


packages/concerto-core/lib/modelmanager.js
```diff
@@ -88,6 +89,30 @@ class ModelManager extends BaseModelManager {
...
+
+    /**
+     * Adds a model in AST format to the ModelManager.
+     * @param {object} ast - The AST of the model.
+     * @param {string} [fileName] - An optional file name to associate with the model file.
+     * @param {boolean} [disableValidation] - If true then the model files are not validated.
+     * @throws {IllegalModelException}
+     * @return {ModelFile} The newly added model file (internal).
+     */
+    addModel(ast, cto, fileName, disableValidation) {
```
sanketshevkar, line 111: This would override the `addModel` in `BaseModelManager`.

packages/concerto-core/lib/modelmanager.js
```diff
@@ -88,6 +89,30 @@ class ModelManager extends BaseModelManager {
         return this.addModel(cto, cto, fileName, disableValidation);
     }
 
+    /**
+     * Validates the metamodel version.
+     * @param {string} version - The metamodel version to validate.
+     * @throws {Error} If the version is not supported.
+     */
+    validateMetamodelVersion(version) {
+        if (!this.metamodelVersions.includes(version)) {
```
sanketshevkar, line 98: Thanks a lot for the PR. But we first need to decide on the strategy how do we support multiple versions of metamodel first and then make relavant changes.

Happy to hear you opinions on the issue discssion board. Thanks!

Atul-khadse: @sanketshevkar Thank you for the feedback! That makes sense. I'll share my thoughts on the strategy for supporting multiple metamodel versions on the issue discussion board. Looking forward to collaborating on this further!

PR #993, opened 2025-03-13
feat(decorators): added support for multiple decorators

### LINKED ISSUE #992: Decorator command in DCS should support multiple decorators for a single target.
# Feature Request 🛍️

A single model element target can have multiple decorators. Currently in DCS, we can only have one decorator per target. The request is to support multiple decorators per target.

Current behaviour
```
"commandSet": {
    "name": "test.dcs",
    "version": "1.0.0",
    "commands": [
      {
        "type": "UPSERT",
        "target": {
          "namespace": "test@1.0.0",
          "declaration": "Car",
          "$class": "org.accordproject.decoratorcommands@0.3.0.CommandTarget"
        },
        "decorator": {
          "name": "Hide",
          "arguments": [],
          "$class": "concerto.metamodel@1.0.0.Decorator"
        },
        "$class": "org.accordproject.decoratorcommands@0.3.0.Command"
      }
    ],
    "$class": "org.accordproject.decoratorcommands@0.3.0.DecoratorCommandSet"
  }
```

New behaviour
```
  "commandSet": {
    "name": "test.dcs",
    "version": "1.0.0",
    "commands": [
      {
        "type": "UPSERT",
        "target": {
          "namespace": "test@1.0.0",
          "declaration": "Car",
          "$class": "org.accordproject.decoratorcommands@0.3.0.CommandTarget"
        },
        "decorators": [
            {
            "name": "Hide",
            "arguments": [],
            "$class": "concerto.metamodel@1.0.0.Decorator"
          },
          {
            "name": "Term",
            "arguments": [],
            "$class": "concerto.metamodel@1.0.0.Decorator"
          }
        ],
        "$class": "org.accordproject.decoratorcommands@0.3.0.Command"
      }
    ],
    "$class": "org.accordproject.decoratorcommands@0.3.0.DecoratorCommandSet"
  }
```

## Use Case

Support multiple decorators per target.

## Possible Solution

Update the DCS model and decoratorManager

riyadhami:
# Closes #992

This PR adds support for handling multiple decorators per target in Concerto models. Currently, only single decorators are handled, and this enhancement allows parsing and processing of multiple decorators

### Changes

- Modified `DecoratorManager.js`, `DecoratorExtractor.js`, and `VocabularyManager.js` to support parsing and processing of multiple decorators.
- Updated related functions to handle decorator arrays instead of single instances.

### Flags

- Need to add additional test coverage based on maintainers' input.

### Related Issues
- Issue #992 

### Author Checklist

### DIFF
--- a/packages/concerto-core/lib/decoratormanager.js
+++ b/packages/concerto-core/lib/decoratormanager.js
@@ -77,11 +77,11 @@ enum MapElement {
 }
 
 /**
- * Applies a decorator to a given target
+ * Applies decorators to a given target
  */
 concept Command {
     o CommandTarget target
-    o Decorator decorator
+    o Decorator[] decorators
     o CommandType type
     o String decoratorNamespace optional
 }
@@ -687,11 +687,15 @@ class DecoratorManager {
      * @param {*} command the Command object from the dcs
      */
     static executeNamespaceCommand(model, command) {
-        const { target, decorator, type } = command;
+        const { target, decorators, type } = command;
         if (Object.keys(target).length === 2 && target.namespace) {
             const { name } = ModelUtil.parseNamespace( model.namespace );
             if(this.falsyOrEqual(target.namespace, [model.namespace, name])) {
-                this.applyDecorator(model, type, decorator);
+                if (decorators && decorators.length > 0) {
+                    decorators.forEach(dec => {
+                        this.applyDecorator(model, type, dec);
+                    });
+                }
             }
         }
     }
@@ -708,7 +712,7 @@ class DecoratorManager {
      * org.accordproject.decoratorcommands model
      */
     static executeCommand(namespace, declaration, command, property, options) {
-        const { target, decorator, type } = command;
+        const { target, decorators, type } = command;
         const { name } = ModelUtil.parseNamespace( namespace );
         if (this.falsyOrEqual(target.namespace, [namespace,name]) &&
             this.falsyOrEqual(target.declaration, [declaration.name])) {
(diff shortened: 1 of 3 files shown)


packages/concerto-core/lib/decoratormanager.js
```diff
@@ -77,11 +77,11 @@ enum MapElement {
 }
 
 /**
- * Applies a decorator to a given target
+ * Applies decorators to a given target
  */
 concept Command {
     o CommandTarget target
-    o Decorator decorator
+    o Decorator[] decorators
```
mttrbrts, line 84: This is a breaking change to the DCS metamodel. How can you make this change in a backwards-compatible way?

mttrbrts (commented):
This PR is looking strong, let's figure out how we can make the change to be backwards compatible.

I'm also starting to get concern about the complexity / performance of the execute command, but that should be solved separately.

riyadhami:
Hey, thanks for reviewing the pr. I made the changes so that decorator turns into decorators and instead of an object gives me an array. From the breaking change, I can see that it requires ''Decorator decorator" and not "Decorator[] decorators".  Also, all test cases are made to be working with "decorator" and "decorator Object" so those will always fail from what I can see. Maybe something we could do about this too? (There are a lot of them).  How do you want me to implement this backwards compatibility?

One approach I can think of is : Making "Decorator decorator" @deprecated with just a message saying use decorators instead. We could also maybe just mention the breaking change explicitly as a version increment(multiple decorators support), providing a migration path and documentation.

What do you have in mind? Maybe I could get some lead here as to what exactly to do.

mttrbrts:
> Hey, thanks for reviewing the pr. I made the changes so that decorator turns into decorators and instead of an object gives me an array. From the breaking change, I can see that it requires ''Decorator decorator" and not "Decorator[] decorators". Also, all test cases are made to be working with "decorator" and "decorator Object" so those will always fail from what I can see. Maybe something we could do about this too? (There are a lot of them). How do you want me to implement this backwards compatibility?
> 
> One approach I can think of is : Making "Decorator decorator" @deprecated with just a message saying use decorators instead. We could also maybe just mention the breaking change explicitly as a version increment(multiple decorators support), providing a migration path and documentation.
> 
> What do you have in mind? Maybe I could get some lead here as to what exactly to do.

@riyadhami I recommend adding `o Decorator[] decorators optional`, and giving primacy to `decorator` over `decorators`.

That way, any existing applications that use the library (and our tests) will continue to work unchanged, but where applications want to opt-in to the new feature they can choose to do that.

riyadhami:
I've made the requested changes. Should I open a new PR or continue working on this one?

PR #1000, opened 2025-03-21
fix(analysis): detect changes for various model modifications

### LINKED ISSUE #719: Concerto analysis changes
# Bug Report 🐛

Concerto analysis returns no change for:
- Changing a default value (maybe also adding one?)
- changing a range to be less restrictive
- adding/removing optional from a property
- Importing a different minor or patch version

## Expected Behavior

- Removing `optional` from property should be a breaking change.
- The rest should probably be patch or minor
- Possibly add way to extend config to add your own rules to use

## Steps to Reproduce

1.
2.
3.
4.

### Desktop
 - OS: [e.g. macOS]
 - Version: [e.g. 0.22.15]

khaledGadelhaQ:
# Closes #719

### Changes

This PR addresses issue #719 by adding logic to detect changes in the following scenarios:
- [x] Adding/removing `optional ` from a property
- [x] Adding/removing default values from a property
- [x]  Changing the value of default values for properties
- [x] Adding/removing default values from scalars
- [x] Changing the value of default values for scalars

Each change is covered by **unit tests** to ensure correctness. This ensures that the `concerto-analysis` package correctly identifies and reports these changes, improving the accuracy of model comparisons.

---

### **Testing**
- All new changes are covered by unit tests.
- Verified against sample `.cto` files to ensure accurate detection of changes.

---
### Flags

NA

### Screenshots or Video
NA

### Related Issues
- Issue #719 

### Author Checklist
- [x] Merging to `main` from `fork:khaledGadelhaQ/issue-719/fix-analysis-changes-detection`

### DIFF
--- a/packages/concerto-analysis/src/compare-config.ts
+++ b/packages/concerto-analysis/src/compare-config.ts
@@ -53,6 +53,8 @@ export const defaultCompareConfig: CompareConfig = {
         'optional-property-added': CompareResult.PATCH,
         'required-property-removed': CompareResult.MAJOR,
         'optional-property-removed': CompareResult.MAJOR,
+        'optional-to-required-property': CompareResult.MAJOR,
+        'required-to-optional-property': CompareResult.PATCH,
         'namespace-changed': CompareResult.ERROR,
         'enum-value-added': CompareResult.PATCH,
         'enum-value-removed': CompareResult.MAJOR,
--- a/packages/concerto-analysis/src/compare-config.ts
+++ b/packages/concerto-analysis/src/compare-config.ts
@@ -72,6 +72,9 @@ export const defaultCompareConfig: CompareConfig = {
         'scalar-default-value-added': CompareResult.PATCH,
         'scalar-default-value-removed': CompareResult.MAJOR,
         'scalar-default-value-changed': CompareResult.MAJOR,
+        'property-default-value-added': CompareResult.PATCH,
+        'property-default-value-removed': CompareResult.MAJOR,
+        'property-default-value-changed': CompareResult.MAJOR,
     },
 };
 
@@ -177,3 +180,6 @@ export class CompareConfigBuilder {
         return this;
     }
 }
+
+export const DEFAULT_COMPARER_FACTORIES_COUNT = defaultCompareConfig.comparerFactories.length;
+export const DEFAULT_RULES_COUNT = Object.keys(defaultCompareConfig.rules).length;
--- a/packages/concerto-analysis/src/compare-config.ts
+++ b/packages/concerto-analysis/src/compare-config.ts
@@ -69,6 +69,9 @@ export const defaultCompareConfig: CompareConfig = {
         'scalar-validator-added': CompareResult.MAJOR,
         'scalar-validator-removed': CompareResult.PATCH,
         'scalar-validator-changed': CompareResult.MAJOR,
+        'scalar-default-value-added': CompareResult.PATCH,
+        'scalar-default-value-removed': CompareResult.MAJOR,
+        'scalar-default-value-changed': CompareResult.MAJOR,
     },
 };
 
--- a/packages/concerto-analysis/src/compare-config.ts
+++ b/packages/concerto-analysis/src/compare-config.ts
@@ -69,12 +69,12 @@ export const defaultCompareConfig: CompareConfig = {
         'scalar-validator-added': CompareResult.MAJOR,
         'scalar-validator-removed': CompareResult.PATCH,
         'scalar-validator-changed': CompareResult.MAJOR,
-        'scalar-default-value-added': CompareResult.PATCH,
+        'scalar-default-value-added': CompareResult.MINOR,
         'scalar-default-value-removed': CompareResult.MAJOR,
-        'scalar-default-value-changed': CompareResult.MAJOR,
-        'property-default-value-added': CompareResult.PATCH,
+        'scalar-default-value-changed': CompareResult.PATCH,
+        'property-default-value-added': CompareResult.MINOR,
         'property-default-value-removed': CompareResult.MAJOR,
-        'property-default-value-changed': CompareResult.MAJOR,
+        'property-default-value-changed': CompareResult.PATCH,
     },
 };
 
--- a/packages/concerto-analysis/src/comparers/properties.ts
+++ b/packages/concerto-analysis/src/comparers/properties.ts
@@ -236,4 +236,36 @@ const propertyValidatorChanged: ComparerFactory = (context) => ({
     }
 });
 
-export const propertyComparerFactories = [propertyAdded, propertyRemoved, propertyTypeChanged, propertyValidatorChanged];
+const propertyOptionalChanged: ComparerFactory = (context) => ({
+    compareProperty: (a, b) => {
+        if (!a || !b) {
+            return;
+        }
+        const aIsOptional = a.isOptional();
+        const bIsOptional = b.isOptional();
+        if (aIsOptional === bIsOptional) {
+            return;
+        }
+        if(aIsOptional && !bIsOptional) {
+            const classDeclarationType = getDeclarationType(a.getParent());
+            const type = getPropertyType(a);
+            context.report({
+                key: 'optional-to-required-property',
+                message: `The optional ${type} "${a.getName()}" was changed to be required in the ${classDeclarationType} "${a.getParent().getName()}"`,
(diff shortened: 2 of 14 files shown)


packages/concerto-analysis/src/compare-config.ts
```diff
@@ -67,6 +69,9 @@ export const defaultCompareConfig: CompareConfig = {
         'scalar-validator-added': CompareResult.MAJOR,
         'scalar-validator-removed': CompareResult.PATCH,
         'scalar-validator-changed': CompareResult.MAJOR,
+        'scalar-default-value-added': CompareResult.PATCH,
```
mttrbrts, line 72: @ekarademir Do you agree with this classification? Is this expected to be a backwards-compatible change in all scenarios?

For example, where a client application relies on `null` or `undefined` values by default?
We also have the scenario of implicit defaults for some property types, e.g. required Boolean properties

mttrbrts (commented):
Thanks @khaledGadelhaQ, this is a strong contribution. 

Once you get an approval from a maintainer, let's merge the PR. Additional comparison cases can be added in separate PRs.

packages/concerto-analysis/src/compare-config.ts
```diff
@@ -67,6 +69,9 @@ export const defaultCompareConfig: CompareConfig = {
...
+        'scalar-default-value-removed': CompareResult.MAJOR,
+        'scalar-default-value-changed': CompareResult.MAJOR,
```
ekarademir, line 74: We have extended the comparer internally in DocuSign and the extended rules have been in use quite a while now, so we have a healthy amount of data to back the rationale. Our rules are as follows:
```
'scalar-default-value-added': CompareResult.MINOR,
'scalar-default-value-removed': CompareResult.MINOR,
'scalar-default-value-changed': CompareResult.PATCH,
```

* Changing the default value of the scalar won't break any validation, it would even mean that, previous default value was a mistake so the change could be patching something missing.
* Adding a default value is also not a breaking change, but since it should be a larger change than a small patch. So we set that to minor.
* Removing a default value might be a major change, but having it as minor, didn't cause any problems. We can argue that we don't have optional declarations (we have optional properties,) so having a default value for a scalar declaration is practically making it optional, and removing it would make it non-optional and that is a breaking change. If we decide on MAJOR, we would update the internal rules too.

packages/concerto-analysis/src/compare-config.ts
ekarademir: > For example, where a client application relies on `null` or `undefined` values by default? We also have the scenario of implicit defaults for some property types, e.g. required Boolean properties

I think semantically, adding a default value is not a breaking change. It's akin to making a required field an optional field. If clients are relying on `null` or `undefined`, I would assume they are already treating the field as optional.

packages/concerto-analysis/src/comparers/properties.ts
```diff
@@ -236,4 +236,36 @@ const propertyValidatorChanged: ComparerFactory = (context) => ({
     }
 });
 
-export const propertyComparerFactories = [propertyAdded, propertyRemoved, propertyTypeChanged, propertyValidatorChanged];
+const propertyOptionalChanged: ComparerFactory = (context) => ({
```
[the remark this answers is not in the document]
khaledGadelhaQ: Yes, you can invoke this analysis by comparing two properties by adding an optional property to one field.
`old.cto`
old
`new.cto`
new

When you run the analysis, it will give you the following report:
result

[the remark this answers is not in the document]
khaledGadelhaQ: The event `propertyAdded` will be triggered when a new property is added. The event `propertyRemoved` will be triggered when a property is removed. The event `propertyTypeChanged` will be triggered when the type of a property is changed. The event `propertyValidatorChanged` will only compare validators, meaning that none of the previous analyses will cause a higher scoped change to be triggered.

packages/concerto-analysis/src/compare-config.ts
khaledGadelhaQ: The same thing applies to properties when adding/removing or changing a default value.
The `concerto-analysis` does not detect any change when adding/removing or changing a default value should we add rules for those cases? something like: 
` 'property-default-value-added',
   'property-default-value-removed', 
    'property-default-value-changed'`

khaledGadelhaQ:
@mttrbrts  Thank you for the feedback. I have completed the changes for this pull request based on the previous suggestions. I will add the remaining comparison cases in separate pull requests.

I look forward to any code reviews and will make further changes as needed. I will be waiting for any code reviews and will apply the changes as required.

packages/concerto-analysis/src/compare-config.ts
ekarademir: > should we add rules for those cases? something like:
> ` 'property-default-value-added', 'property-default-value-removed', 'property-default-value-changed'`

That would tie up the PR nicely actually.

packages/concerto-analysis/src/comparers/properties.ts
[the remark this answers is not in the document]
ekarademir: I think with or without this change the result of the analysis won't change (3 means MAJOR) however no harm being more descriptive.

packages/concerto-analysis/src/compare-config.ts
ekarademir: What is your rationale keeping the "default value changed" to MAJOR and "default value added" to PATCH @khaledGadelhaQ ?

khaledGadelhaQ: @ekarademir  It depends on the client application. Changing an existing default value is breaking because any code relying on the original default will behave differently. Serialization/deserialization outputs change, that's why it's `MAJOR`, while adding a default value will be seen as non-breaking if the property was previously `undefined`/`null`. This was my reasoning behind that , I think it depends on how your application uses properties with no default values.

packages/concerto-analysis/src/comparers/properties.ts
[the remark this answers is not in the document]
khaledGadelhaQ: if you removed `propertyOptionalChanged` the analysis will not report any changes in the two models.

packages/concerto-analysis/src/compare-config.ts
khaledGadelhaQ: @ekarademir @mttrbrts,  Are we agreed on keeping this change as `PATCH`?

khaledGadelhaQ:
@ekarademir Could you please review the latest commits?

packages/concerto-analysis/src/compare-config.ts
dselman: LGTM

packages/concerto-analysis/src/compare-config.ts
dselman: Some notes from me:
- Default values do not change the type system (i.e. the type of a property is the same, regardless of whether it has a default value, or not). In that sense they are not breaking (e.g. generated code generation is unchanged). This implies that these changes are POTENTIALLY minor/patch from a model pov, but we should also consider serialized instances of the models:

Serialization and RPC:
- Adding a default value means that clients no longer have to supply a value, but they still can... => MINOR (because this feels like a functional change, not a fix)
- Removing a default value on a **required** property means that clients now have to supply a value, and in the past they didn't have to. Runtime breaking change. Serialised instances are invalid. => MAJOR
- Removing a default value on an **optional** property has no impact on clients or serialised instances => MINOR
- Changing the default value of a property won't break existing clients or serialised instances => PATCH (could be considered a fix)
- Adding a default value on a scalar => MINOR
- Removing the default value on a scalar depends on whether the **usage** of the scalar is optional, so the conservative thing to do is to say that this could be a breaking change. => MAJOR
- Changing the default value of a scalar => PATCH

Side note: we should double check that the default value passes the other facets we have on the property. E.g. the default value for an Integer must fall within the `range` checks. I don't remember whether we do this when we validate the model.

khaledGadelhaQ: I have added the new changes. Thanks for the feedback. I also tested whether the default value passes the other facets we have on the property, and it did not pass the checks! I will open an issue for that.

khaledGadelhaQ:
@mttrbrts, any updates on this PR?

mttrbrts (approved):

ekarademir (approved):

PR #1001, opened 2025-03-22
fix(parser): collect errors without halting and include file IDs

### LINKED ISSUE #999: Improve Error Handling in parseModels Function
# Discussion 🗣
The current implementation of the `parseModels` function in the parser module lacks error handling. When parsing an array of model files, if one file fails to parse, the entire function throws an exception immediately. This prevents the processing of subsequent files, which could be valid, and makes debugging and maintenance more difficult.

## Context
In environments where multiple model files are processed, it's common that one or a few files might contain errors while the others are correctly formatted. Immediate termination upon encountering an error not only halts the overall process but also leaves users without any parsed results from valid files.

Current implementation in `parserMain.js`:
```javascript
function parseModels(files, options) {
    const result = {
        $class: `${MetaModelNamespace}.Models`,
        models: [],
    };
    files.forEach((modelFile) => {
        let metaModel = Parser.parse(modelFile, options);
        result.models.push(metaModel);
    });
    return result;
}
```

## Proposed improvements

- Add a way to collect errors rather than throwing immediately
- Include file identifiers in error messages

fuyalasmit:
# Closes #999 

### Changes

- Modified `parseModels` to collect errors in an array instead of throwing immediately
- Added file identifiers (`file_${index}` or explicit `fileName`) to error messages

### Flags

- **File Identification:**  
  Errors use generated `file_${index}` IDs (e.g., `file_1`) instead of actual filenames.  
  This is because the input `files` array may contain raw CTO strings (legacy) without filenames.  

### Author Checklist
- [x] Merging to `main` from `fuyal:asmit/999`

### DIFF
--- a/packages/concerto-cto/lib/parserMain.js
+++ b/packages/concerto-cto/lib/parserMain.js
@@ -44,17 +44,36 @@ function parse(cto, fileName, options) {
  * Parses an array of model files
  * @param {string[]} files - array of cto files
  * @param {Object} [options] - an optional options parameter
- * @param {string} [options.skipLocationNodes] - when true location nodes will be skipped in the metamodel AST
- * @return {*} the AST / metamodel
+ * @param {boolean} [options.skipLocationNodes] - when true location nodes will be skipped in the metamodel AST
+ * @return {Object} an object with:
+ *   - `$class` {string}: the AST/metamodel namespace
+ *   - `models` {Object[]}: successfully parsed metamodels
+ *   - `errors` {Object[]}: parse errors, each containing:
+ *       - `file` {string}: file identifier
+ *       - `message` {string}: error message
+ *       - `location` {Object} [optional]: error location, if available
  */
 function parseModels(files, options) {
     const result = {
         $class: `${MetaModelNamespace}.Models`,
         models: [],
+        errors: [],
     };
-    files.forEach((modelFile) => {
-        let metaModel = Parser.parse(modelFile, options);
-        result.models.push(metaModel);
+    files.forEach((modelFile, index) => {
+        const fileId = `file_${index}`; // for unique file id
+        try {
+            let metaModel = parse(modelFile, fileId, options);
+            result.models.push(metaModel);
+        } catch (err) {
+            const errorDetails = {
+                file: err.fileName || fileId,
+                message: err.message,
+            };
+            if (err.fileLocation) {
+                errorDetails.location = err.fileLocation;
+            }
+            result.errors.push(errorDetails);
+        }
     });
     return result;
 }
--- a/packages/concerto-cto/test/parserMain.js
+++ b/packages/concerto-cto/test/parserMain.js
@@ -61,8 +61,23 @@ describe('parser', () => {
             {
                 $class: 'concerto.metamodel@1.0.0.Models',
                 models: getCTOFiles().map(({ ast }) => JSON.parse(ast)),
+                errors: [],
             }
         );
+    });
+
+    it('Should handle errors in multiple files', () => {
+        const files = [
+            getCTOFiles()[0].content, // Valid (file_0)
+            'namespace invalid\nconcept Bad { o String bad', // Invalid (file_1)
+            getCTOFiles()[1].content, // Valid (file_2)
+        ];
+        const mm = Parser.parseModels(files, { skipLocationNodes: true });
+        mm.models.length.should.equal(2);
+        mm.errors.length.should.equal(1);
+        mm.errors[0].file.should.equal('file_1');
+        mm.errors[0].message.should.match(/Expected .+ but/);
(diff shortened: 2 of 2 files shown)


fuyalasmit:
@mttrbrts can you please review this ?

packages/concerto-cto/lib/parserMain.js
```diff
@@ -44,17 +44,36 @@ function parse(cto, fileName, options) {
...
+        } catch (err) {
+            const errorDetails = {
+                file: err.fileName || fileId,
+                message: err.message,
+            };
+            if (err.fileLocation) {
+                errorDetails.location = err.fileLocation;
+            }
+            result.errors.push(errorDetails);
+        }
```
mttrbrts, line 76: Is there a way to make this backwards-compatible for existing client applications?

For example, if a client relies upon the parser throwing errors today, then after merging, they will miss errors that are aggregated into the `error` array.

We do have the option of making a major version bump in the SDK but try to avoid that where possible.

fuyalasmit: For making it backwards-compatible, I modified `parseModels` to collect all parsing errors in the `errors` array (as before) but also to throw an aggregate error if any errors occur after processing all files.

This maintains the original behavior of throwing an error on failure, so existing clients won’t break.

Please take a look and let me know if this resolves your concern.

mttrbrts: Thanks @fuyalasmit, that's an improvement. 

Is `result.errors` still needed in the return object? If it's empty, then it's never used?

fuyalasmit: @mttrbrts Do we want `parseModels` to:
- Process all files, collecting errors.
- Throw one exception with all error details if any occur.
- Return `{ $class, models }` only if all files parse successfully?

This removes `errors` from the return object for backward compatibility. 

Updated code (comments show old approach):

```js
function parseModels(files, options) {
    const result = {
        $class: `${MetaModelNamespace}.Models`,
        models: [],
        // errors: [],
    };
    const errors = [];
    files.forEach((modelFile, index) => {
        const fileId = `file_${index}`; // for unique file id
        try {
            let metaModel = parse(modelFile, fileId, options);
            result.models.push(metaModel);
        } catch (err) {
            const errorDetails = {
                file: err.fileName || fileId,
                message: err.message,
            };
            if (err.fileLocation) {
                errorDetails.location = err.fileLocation;
            }
            // result.errors.push(errorDetails);
            errors.push(errorDetails);
        }
    });
    // if (result.errors.length > 0) {
    if (errors.length > 0) {
        // const errorMessages = result.errors.map(e=> `Error in ${e.file}: ${e.message}`).join('; ');
        const errorMessages = errors.map(e=> `Error in ${e.file}: ${e.message}`).join('; ');
        // const aggregateError = new Error(`Parsing errors occurred in ${result.errors.length} files: ${errorMessages}`);
        const aggregateError = new Error(`Parsing errors occurred in ${errors.length} files: ${errorMessages}`);
        // aggregateError.errors = result.errors; // individual errors
        aggregateError.errors = errors; // individual errors
        throw aggregateError;
    }
    return result;
}
```

fuyalasmit:
made all the required changes, kindly review once.

packages/concerto-cto/lib/parserMain.js
```diff
@@ -44,18 +44,45 @@ function parse(cto, fileName, options) {
...
+            };
+            if (err.fileLocation) {
+                errorDetails.location = err.fileLocation;
+            }
+            errors.push(errorDetails);
+        }
     });
+    if (errors.length > 0) {
+        const errorMessages = errors.map(e=> `Error in ${e.file}: ${e.message}`).join('; ');
+        const aggregateError = new Error(`Parsing errors occurred in ${errors.length} files: ${errorMessages}`);
```
mttrbrts, line 82:
```suggestion
        const errorMessages = errors.map(e=> `\tError in ${e.file}: ${e.message}`).join('\n');
        const aggregateError = new Error(`Parsing errors occurred in ${errors.length} files:\n${errorMessages}`);
```

mttrbrts (approved):
Thanks @fuyalasmit. I just made one final suggestion to better format the output messages.

In future, it would be helpful to extend the signature of the parseModels function to allow a client application to pass in real file names (such as in the Concerto CLI).

fuyalasmit:
> Thanks @fuyalasmit. I just made one final suggestion to better format the output messages.
> 
> In future, it would be helpful to extend the signature of the parseModels function to allow a client application to pass in real file names (such as in the Concerto CLI).

Passing the real name would be a good idea to be implemented in future. 
Thank you for the suggestion, I have committed the changes. PR should be good to go.

fuyalasmit:
.

Issue #982, opened 2025-01-30
If a declaration is named "properties" then we can't write a valid vocabulary for it


jamieshorten:
# Bug Report 🐛
If we have a model like
```
@Test2('world')
namespace com.example.e2e.a@1.0.1
@Term('A nice properties term')
concept properties {
  @Term('A cool string')
  o String someString
}
```
and run `concerto extract-decorators --models MODEL.cto --removeDecoratorsFromSource true` we get the following `vocabulary.yml`:
```
locale: en
namespace: com.example.e2e.a@1.0.1
declarations:
  - properties: A nice properties term
    properties:
      - someString: A cool string
```
Correct me if I'm missing something but that is an invalid YAML file due to the duplicated `properties` key. Trying to apply this vocab gives:
```
concerto decorate --models com.example.e2e.a.cto --vocabulary vocabulary_0.yml
10:29:53 - ERROR: YAMLParseError: Map keys must be unique at line 5, column 5:

  - properties: A nice properties term
    properties:
```

## Expected Behavior
I'm not sure how/if we should handle this case or if there is a way to write a vocab for the `properties` concept in the above model.

ansh7432:
Hii @jamieshorten please assign this issue to me i wanna work on this

ansh7432:
Hii @jamieshorten @sanketshevkar gentle reminder can I work on this ?

ansh7432:
hi @jamieshorten @sanketshevkar gentle reminder..

PR #1011, opened 2025-04-06
Fix: Print YAML key duplication in decorator extraction

abhicodes01:
Ensure that property decorators are correctly named using Term_<locale> format, 
and prevent incorrect or duplicate application of decorators like 'Term_properties'. 
This resolves rendering issues during vocabulary decoration due to malformed 
decorator names.

Signed-off-by: Abhijeet SIngh Parihar <abhijeetsinghparihar756@gmail.com>

Closes: #982

### DIFF
--- a/packages/concerto-vocabulary/lib/vocabularymanager.js
+++ b/packages/concerto-vocabulary/lib/vocabularymanager.js
@@ -333,7 +333,7 @@ class VocabularyManager {
                                 }
                             });
                         }
-                        else if(term.localeCompare('properties')) {
+                        else if(term === 'properties') {
                             decoratorCommandSet.commands.push({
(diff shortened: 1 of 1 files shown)


packages/concerto-vocabulary/lib/vocabularymanager.js
```diff
@@ -333,7 +333,7 @@ class VocabularyManager {
                                 }
                             });
                         }
-                        else if(term.localeCompare('properties')) {
+                        else if(term === 'properties') {
```
sanketshevkar, line 336: Hi @abhicodes01 
Thanks for you interest.
Can you please explain in detail what is this change about? Is there any issue associated with this?

PR #1012, opened 2025-04-11
fix(filter): prevent modifying models in place

### LINKED ISSUE #987: BaseModelManager filter method, edits AST in place
# Bug Report 🐛

`BaseModelManager` filter operation inadvertently manipulates the AST of any model file in place but then also returns a new filtered model manager. It should do one of them, not both.

This causes a problem, if the model ASTs are kept in memory and reused. Because the operation manipulates those objects in memory without advertising it.

## Steps to Reproduce

1. Create several models that depend on each other in memory
2. Apply a filter operation that would result in removal of imports
3. Inspect the models in memory

### Desktop
 - OS: [e.g. macOS]
 - Version: [e.g. 0.22.15]

fuyalasmit:
# Closes #987

Ensures `filter` method returns a new model manager with filtered results instead of modifying the original models

### Changes

- Updated `filter` to use deep copies.  

### Author Checklist
- [x] Merging to `main` from `fuyalasmit/concerto:asmit/987`

### DIFF
--- a/packages/concerto-core/lib/basemodelmanager.js
+++ b/packages/concerto-core/lib/basemodelmanager.js
@@ -849,8 +849,15 @@ class BaseModelManager {
         const modelManager = new BaseModelManager({...this.options}, this.processFile);
         const removedFqns = []; // the list of FQN of types that have been removed
 
+        // deep copy of the model files
+        const copiedModelFiles = Object.values(this.modelFiles).map(modelFile => {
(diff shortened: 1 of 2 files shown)


fuyalasmit:
@ekarademir is this the correct approach of solving #987 ?
could you please review this ?

fuyalasmit:
.

PR #1014, opened 2025-04-28
Improve Deep Copy Performance in Decorator Manager

yashkp1234:
This PR improves the performance of deep copy operations in the decorator manager component of concerto-core.

Changes:
- Refactored deep copy implementation in decorator manager to improve performance
- Component affected: concerto-core

Please review the changes and provide feedback on the performance improvements.


packages/concerto-core/lib/decoratormanager.js
```diff
@@ -349,9 +353,8 @@ class DecoratorManager {
      * @private
      */
     static pushMapValues(array, map, key) {
-        const targetCommands = map.get(key);
-        if (targetCommands) {
-            array.push(...targetCommands);
+        for (const value of map.get(key) || []) {
+            array.push(value);
         }
     }
```
muhabdulkadir, line 359: Why push an empty array?

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -389,21 +392,21 @@ class DecoratorManager {
                     };
                 })
                 : []);
-        }).flat().filter(i => i.namespace);
+        }).filter(i => i.namespace);
         const { namespaceCommandsMap, declarationCommandsMap, propertyCommandsMap, mapElementCommandsMap, typeCommandsMap }  = this.getDecoratorMaps(decoratorCommandSet);
         const ast = modelManager.getAst(true, true);
-        const decoratedAst = JSON.parse(JSON.stringify(ast));
+        const decoratedAst = rfdc(ast);
```
ekarademir, line 398: How about `structuredClone`? https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone it's a standard function that is also available in NodeJS.

mttrbrts:
Benchmarking framework useful for #1003 too.

packages/concerto-core/lib/decoratormanager.js
mttrbrts: https://github.com/davidmarkclements/rfdc/issues/35

ekarademir (approved):

packages/concerto-core/lib/decoratormanager.js
muhabdulkadir: My point was that using the copy is more efficient for large object.

PR #1022, opened 2025-06-11
chore(core): decorateModels perf config

sanketshevkar:

### Changes

- decorateModels getAst resolution to be triggered optionally, if the consumer are confident that their models are already resolved, they can use the `options.disableMetamodelResolution` to avoid the extra overhead.
- decorateModels fromAst validation to be triggered optionally, if the consumers are confident that their decorator within DCS and the DCS objects themselves are valid, also the models are also valid they can use the `options.disableMetamodelValidation` to avoid the extra overhead. The validation was the step which ensured that the model elements do not have two decorators with same name. This check was replicated at the method where decorators where `APPENDED`.
- Semver validation was also made to be optionally triggered. There are places within the code base where Semver validation was done multiple times under the pretext of parsing the namespace. with the option `options.disableValidation` in parseNamespace method this extra overhead can be avoided.

Before:
 ```
 modelManager = DecoratorManager.decorateModels(modelManager, dcsContent, {
  validate: false, // We validated on write, so no need to pay the cost of validation again here
});
 ```
The decorateModels method took **39-42ms** with the sample data

After:
```
modelManager = DecoratorManager.decorateModels(modelManager, dcsContent, {
  validate: false, // We validated on write, so no need to pay the cost of validation again here
  disableMetamodelResolution: true,
  disableMetamodelValidation: true,
  enableDcsNamespaceTarget: true
});
```
The decorateModels with the following config takes **17-20ms** with the sample data.

That's **48% to 59%** improvement.

Flame Graphs
- Before
Screenshot 2025-06-11 at 4 20 08 PM

- After
Screenshot 2025-06-11 at 4 19 33 PM

### Author Checklist


ekarademir (dismissed):

packages/concerto-core/lib/modelutil.js
```diff
@@ -114,29 +114,34 @@ class ModelUtil {
      * its name and version parts. The version of the namespace
      * (if present) is parsed using semver.parse.
      * @param {string} ns the namespace to parse
+     * @param {object} [options] optional parsing options
+     * @param {boolean} [options.disableVersionValidation] if false, the version will be validated
```
mttrbrts, line 118: How about disableVersionParsing?

mttrbrts (approved):
Great work, Sanket. Happy to merge once we have consensus on naming.

ekarademir (commented):
From @dselman

> Are these two new options orthogonal? Can we combine them as a single option?

From myself, I'm fine with disable or skip actually. Apparently `disableX` has been used elsewhere too.

ekarademir:
> From @dselman
> 
> > Are these two new options orthogonal? Can we combine them as a single option?
> 
> From myself, I'm fine with disable or skip actually. Apparently `disableX` has been used elsewhere too.

So if you can try to consolidate the option if it makes sense, that would be good.

packages/concerto-core/lib/modelutil.js
sanketshevkar: `SemVer.parse` and `SemVer.validate` are two different methods and playing around I think they are independent and validate method being the most expensive one, that's why I chose valiation.

mttrbrts: You're correct, however, the reason that we do validation is to handle malformed versions gracefully before parsing them. The returned object contains the parsed version by default, and that's the main client perceivable change, IMO.

sanketshevkar: Understood, I've made the change.
If `disableVersionValidation` is set to true, it just returns the name of the model.
```
{
 name: org.accordproject.test
}
```

packages/concerto-core/lib/decoratormanager.js
```diff
@@ -370,10 +371,22 @@ class DecoratorManager {
...
      * @param {boolean} [options.enableDcsNamespaceTarget] - flag to control applying namespace targeted decorators on top of the namespace instead of all declarations in that namespace
+     * @param {boolean} [options.fastMode] - optional flag to disable both metamodel resolution and validation, only use if you are sure that the model manager has fully resolved models
+     * @param {boolean} [options.disableMetamodelResolution] - flag to disable metamodel resolution, only use if you are sure that the model manager has fully resolved models
+     * @param {boolean} [options.disableMetamodelValidation] - flag to disable metamodel validation, only use if you are sure that the models and decorators are already validated
      * @returns {ModelManager} a new model manager with the decorations applied
      */
     static decorateModels(modelManager, decoratorCommandSet, options) {
 
+        if (options?.fastMode) {
+            if (options?.disableMetamodelResolution === false || !options?.disableMetamodelValidation === false || options?.enableDcsNamespaceTarget === false) {
```
sanketshevkar, line 382: Added a new fastMode for model decoration. It sets `disableMetamodelResolution, disableMetamodelValidation, enableDcsNamespaceTarget` to true if unset, or undefined. If either of them is false then it will throw error.

mttrbrts (approved):

PR #1039, opened 2025-07-16
Fix: Resolve default-ruleset Publish Failure

Ahmed-Gaper:
This PR addresses the recent failure of the `concerto-linter` package to publish to npm. The root cause was an incorrect dependency declaration for `@accordproject/concerto-linter-default-ruleset`.

Previously, `concerto-linter` attempted to depend on `@accordproject/concerto-linter-default-ruleset` with a fixed version (`"1.0.0"`). This created a circular dependency issue where the linter couldn't publish because its dependency wasn't yet available on npm.

The solution involves two key changes:

1.  **Update Dependency Path:** The `concerto-linter-default-ruleset` dependency in `package.json` has been updated to use a **local file path**:

    ```json
    "@accordproject/concerto-linter-default-ruleset": "file:./default-ruleset"
    ```
    This ensures that `default-ruleset` is always included directly within `concerto-linter`, resolving the immediate publish blockage.

2.  **Automate Release Versioning:** The `scripts/bump_version.js` file has been modified. This script will now automatically **replace the `file:` path with the correct stable version** of `concerto-linter-default-ruleset` during future stable releases. This maintains the correct dependency structure for published versions while allowing for local development and build processes.

Additionally, to improve modularity and prevent similar issues in the future:

* **Decoupled Default Ruleset Unit Tests:** The unit tests within `default-ruleset` have been refactored to be **loosely coupled**, removing their direct dependency on `concerto-linter`. This makes the `default-ruleset` more independent and robust.

These changes collectively ensure that `concerto-linter` can publish successfully and that its dependencies are handled correctly both during development and in production releases.

### Author Checklist
- [x] Merging to `main` from `Ahmed-Gaper/ruleset-publish-failure`

### DIFF
--- a/packages/concerto-linter/package.json
+++ b/packages/concerto-linter/package.json
@@ -33,7 +33,7 @@
     "license": "Apache-2.0",
     "dependencies": {
         "@accordproject/concerto-core": "3.21.0",
-        "@accordproject/concerto-linter-default-ruleset": "1.0.0",
+        "@accordproject/concerto-linter-default-ruleset": "file:./default-ruleset",
         "find-up": "5.0.0",
         "@stoplight/spectral-cli": "6.15.0",
         "@stoplight/spectral-core": "1.20.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -15504,7 +15504,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@accordproject/concerto-core": "3.21.0",
-        "@accordproject/concerto-linter-default-ruleset": "1.0.0",
+        "@accordproject/concerto-linter-default-ruleset": "file:./default-ruleset",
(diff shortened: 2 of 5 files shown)


packages/concerto-linter/package.json
```diff
@@ -33,7 +33,7 @@
     "license": "Apache-2.0",
     "dependencies": {
         "@accordproject/concerto-core": "3.21.0",
-        "@accordproject/concerto-linter-default-ruleset": "1.0.0",
+        "@accordproject/concerto-linter-default-ruleset": "file:./default-ruleset",
```
jamieshorten, line 36: I'm fine with this for now to unblock the pipeline, but I'd prefer not to rely on a file reference going forward.

jamieshorten (approved):

PR #1044, opened 2025-07-26
feat(concerto-core): implement YAML to JSON conversion for DCS

### LINKED ISSUE #1040: Implement YAML to JSON conversion for Decorator Command Set (DCS)
This issue tracks the implementation of the `yamlToJson()` function, which will convert DCS YAML string into its corresponding JSON format.

This is the reverse of the `jsonToYaml()` function implemented in #1037 (tracked by #1032).

The goal is to allow full bidirectional conversion between the DCS YAML and JSON formats.

---

### References

- JSON → YAML implementation: [#1032](https://github.com/accordproject/concerto/issues/1032), [#1037](https://github.com/accordproject/concerto/pull/1037)
- YAML spec & format rules: [#1024](https://github.com/accordproject/concerto/issues/1024)
-  [Design doc](https://docs.google.com/document/d/1LYhqw-IsEvHd0WuPbUsNU3PIBOElWjznuRv5pbtLzS4/edit?usp=sharing)

### Tasks

- Implement `yamlToJson()` function in `dcsconverter.js`
- Write unit tests for different YAML input cases

fuyalasmit:
# Closes #1040

This PR implements YAML to JSON conversion support for Decorator Command Sets (DCS), completing the bidirectional transformation feature.

- Implemented `yamlToJson()` function in `lib/dcsconverter.js` for converting simplified YAML DCS into JSON
- Exposed `yamlToJson()` via `DecoratorManager`
- Type definitions updated in `dcsconverter.d.ts` and `decoratormanager.d.ts`
- Comprehensive unit tests in `test/dcsconverter.js` and `test/decoratormanager.js`

---

### Author Checklist
- [x] Commit messages follow [AP format](https://github.com/accordproject/techdocs/blob/master/DEVELOPERS.md#commit-message-format)
- [x] Merging to `main` from `fuyalasmit:gsoc/#1040`

### DIFF
--- a/packages/concerto-core/lib/dcsconverter.js
+++ b/packages/concerto-core/lib/dcsconverter.js
@@ -22,6 +22,7 @@ const { MetaModelNamespace } = require('@accordproject/concerto-metamodel');
  * handles the target field of a command
  * @param {object} target the value of target
  * @returns {object} the simplified target object
+ * @private
  */
 function handleTarget(target){
     const targetKeys = Object.keys(target);
@@ -43,6 +44,7 @@ function handleTarget(target){
  * @param {object} [argument.type] the type identifier for type reference arguments
  * @param {boolean} [argument.isArray] whether the type reference is an array
  * @returns {object} simplified argument object
+ * @private
  */
 function handleArguments(argument){
     const mapClassToType = {
@@ -72,6 +74,7 @@ function handleArguments(argument){
  * @param {string} decorator.name the name of the decorator
  * @param {object[]} [decorator.arguments] the list of arguments
  * @returns {object} simplified decorator object with name and arguments
+ * @private
  */
 function handleDecorator(decorator){
     return {
@@ -89,6 +92,7 @@ function handleDecorator(decorator){
  * @param {object} command.target the target to apply decorator to
  * @param {object} command.decorator the decorator to apply
  * @returns {object} simplified commands object
+ * @private
  */
 function handleCommands(command){
     return {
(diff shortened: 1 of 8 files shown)


fuyalasmit:
I've utilized the existing `MetaModelNamespace` from `@accordproject/concerto-metamodel` to avoid hardcoding of the mm namespace: `concerto.metamodel@1.0.0`

packages/concerto-core/lib/dcsconverter.js
```diff
@@ -121,4 +122,92 @@ function jsonToYaml(dcsJson){
...
+ * handles each argument of the decorator
+ * converts simplified argument objects back to full JSON representation
+ * @param {string} MetaModelNamespace - the metamodel namespace
+ * @param {object} argument - the simplified argument object
+ * @param {string} [argument.type] - the argument type for primitive values
+ * @param {*} [argument.value] - the argument value
+ * @param {object} [argument.typeReference] - the type reference object for complex types
+ * @returns {object} - the fully qualified argument object after restoring required fields
+ */
+function restoreArguments(MetaModelNamespace, argument){
```
sanketshevkar, line 136: I think we should make this method private. Look at a few of the methods in decoratormanager.js file, you'll see how we use the private key word.

We use the jsdoc comments to generate types. Over there the method definitions gets marked as private. Although JS doesn't have the concept of `access-modifiers` for classes and methods similar to what a tightly typed language like Java would have. But would I suggest we follow the pattern.

sanketshevkar (changes requested):

packages/concerto-core/lib/dcsconverter.js
fuyalasmit: Just to clarify, if we make this method private, should we also do the same for other similar helper methods?

Basically, should only `jsonToYaml ` and `yamlToJson ` be considered public from this module?

fuyalasmit: Updated it accordingly.

PR #1055, opened 2025-08-12
feat(ci/cd): Added code to restricting the conflict between declared names and imported names

Anshukumar123975:
# Description:
This PR adds changes in the modelfile.js.
This allows the conformance tests to pass on a rule which states that an error should be thrown when there is a conflict between declared names and imported names.

### DIFF
--- a/.github/workflows/conformance-test.yml
+++ b/.github/workflows/conformance-test.yml
@@ -21,6 +21,11 @@ jobs:
         with:
           node-version: 18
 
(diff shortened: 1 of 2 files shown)


ekarademir:
this test is failing https://github.com/Anshukumar123975/concerto/blob/80f98241ee5afdb65328149daa170cd2acf0a5fa/packages/concerto-core/test/introspect/introspector.js#L85-L109

PR #1102, opened 2026-01-07
feat(concerto-util): add MonetaryUtil support

rockaxorb13:
### Description
This PR adds a new `MonetaryUtil` class to the `@accordproject/concerto-util` package. This utility provides functionality to convert numbers into their written word representation (e.g., converting `100` to `"one hundred"`), utilizing the `n2words` library.

### Motivation
As discussed in the previous meeting, there is a need for monetary text conversion support within the library to facilitate the upcoming implementation of the `Monetary` type in the core model. This utility lays the groundwork for that feature.

### Changes
* **New File:** Added `lib/monetary.js` containing the `MonetaryUtil` class.
* **Dependencies:** Added `n2words` to `package.json`.
* **Exports:** Exposed `MonetaryUtil` via `index.js` for external consumption.

### Verification
* Validated that the new utility correctly converts numbers to text strings using the default locale ('en').
* Confirmed the package installs and exports the new class correctly.

### Next Steps
Once merged, this utility will be used by the `concerto-core` logic to handle monetary serialization and human-readable formatting.

### DIFF
--- a/packages/concerto-util/package.json
+++ b/packages/concerto-util/package.json
@@ -8,6 +8,7 @@
   "dependencies": {
     "@supercharge/promise-pool": "1.7.0",
     "debug": "4.3.7",
+    "n2words": "1.13.0",
     "slash": "3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


packages/concerto-util/package.json
```diff
@@ -54,6 +54,7 @@
   "dependencies": {
     "@supercharge/promise-pool": "1.7.0",
     "debug": "4.3.7",
+    "n2words": "^3.0.0",
```
dselman: Please use a pinned/exact package version to prevent supply chain attacks.

dselman (changes requested):
LGTM - just one small change required for the dependency on n2words. Can you also check that this webpacks ok and the increase in the bundle size?

PR #1105, opened 2026-01-10
feat(core&util): migrate packages to typescript & fix circular dependencies 

rockaxorb13:
## Summary
This PR migrates the `concerto-core` package from JavaScript to TypeScript. It addresses critical circular dependency issues discovered during the migration, modernizes the build configuration, and achieves **99% test coverage**.

## Key Changes

1.  **TypeScript Migration:**
    * Renamed all source files in `src/` from `.js` to `.ts`.
    * Updated `package.json` and `tsconfig.json` to support the new TypeScript build pipeline.
    * Added type annotations (using `any` for this initial phase) to satisfy the compiler and enable future strict typing without breaking changes.

2.  **Circular Dependency Fix:**
    * **Issue:** A circular dependency existed between `BaseModelManager` and its helper models (`RootModel`, `DecoratorModel`). `BaseModelManager` required the models at the top level, while the models (implicitly or explicitly) depended on the manager or shared utilities, causing `require` to return empty objects at runtime in the test environment.
    * **Fix:** Refactored `addRootModel` and `addDecoratorModel` in `BaseModelManager.ts` to use **lazy loading**. The `require` calls for `rootmodel.ts` and `decoratormodel.ts` were moved *inside* the methods to ensure they are only loaded when needed, breaking the initialization cycle.
    * **Safety:** Added runtime safety checks to verify that the loaded modules correctly export the expected factory functions (`getRootModel`, `getDecoratorModel`) before execution.

3.  **Namespace Registration Fix:**
    * **Issue:** The `DecoratorModel` was loading but failing to register its namespace (`concerto.decorator@1.0.0`) correctly, causing "No registered namespace" errors in dependent tests.
    * **Fix:** Updated `decoratormodel.ts` to explicitly set the `namespace` property on the AST object before returning it, ensuring `ModelFile` correctly indexes it in the `ModelManager`.

4.  **Test Infrastructure:**
    * Updated `npm test` script to use `ts-node` with the correct project configuration (`tsconfig.build.json`).
    * Achieved **99% test coverage**, confirming that the refactoring did not break existing functionality.

5. **Enhance TypeScript definitions and improve type safety across the codebase**

## Breaking Changes
* None. The public API remains backward compatible. Internal file extensions have changed to `.ts`.

## Checklist
* [x] Build passes (`npm run build`)
* [x] Tests pass (`npm test`) with 99% coverage
* [x] DCO Signed-off

Here is a screenshot showing the test coverage after my modifications:
Screenshot

---

## Detailed Error & Resolution Log:

* **Error:** `TypeError: getDecoratorModel is not a function`
    * **Cause:** Circular dependency causing `require('./decoratormodel')` to return an empty object `{}` during `BaseModelManager` initialization.
    * **Solution:** Implemented lazy loading for model helpers inside `BaseModelManager` methods using `require` inside the function scope.

* **Error:** `IllegalModelException: No registered namespace for type "concerto.decorator@1.0.0.DotNetNamespace"`
    * **Cause:** The `DecoratorModel` AST was missing the explicit `namespace` property expected by `ModelFile` during registration.
    * **Solution:** Explicitly set `ast.namespace` in `src/decoratormodel.ts` to match the CTO definition.

* **Error:** `Failed to load getDecoratorModel. Got: object`
    * **Cause:** `ts-node` ambiguity between `decoratormodel.ts` (code) and `decoratormodel.json` (data) caused it to load the JSON file instead of the source code.
    * **Solution:** Updated `require` calls to explicitly use the `.ts` extension (`require('./decoratormodel.ts')`) in the test environment to force loading the source code.

### DIFF
--- a/packages/concerto-util/src/loaders/httpfileloader.ts
+++ b/packages/concerto-util/src/loaders/httpfileloader.ts
@@ -15,6 +15,7 @@
 'use strict';
 
 import type { FileLoader } from './fileloader';
+const debug = require('debug')('concerto:HTTPFileLoader');
 
 // Ensure fetch is recognized if not in global types
 declare const fetch: (input: string | URL, init?: RequestInit) => Promise<{ ok: boolean; status: number; text(): Promise<string> }>;
@@ -62,8 +63,7 @@ class HTTPFileLoader<T> implements FileLoader<T> {
             };
         }
 
-        // eslint-disable-next-line no-console
-        console.log(requestUrl);
+        debug('loading', requestUrl);
         const response = await fetch(requestUrl, options);
         if (!response.ok) {
             throw new Error(`HTTP request failed with status: ${response.status}`);


rockaxorb13:
Hey@mttrbrts,
Now that the TypeScript migration for core is technically complete and passing locally, I'm looking at the next steps for the v4 Alpha goals.

I am stuck resolving this CI circular dependency, but alongside it, should I look into:

1. Enabling strict: true in tsconfig to catch potential null-checks we missed?
2. Starting the ESM migration (shifting from CommonJS to ES Modules), since we now have a clean TS base?

I'm happy to pick up whichever is higher priority for the v4 release while this PR goes through review.

rockaxorb13:
Thanks alot @dselman and @mttrbrts!

I have resolved the CI failures locally. As discussed, I upgraded to Node v24 and ran a clean build/test cycle. Here is a summary of the changes I made to get everything passing:

Summary of Fixes:

1. CommonJS Compatibility: I ensured concerto-core/index.ts retains the original CommonJS export style (export = ...) as requested.
2. concerto-analysis Tests: The analysis tests were failing because they tried to import the new TypeScript classes via default imports. I updated compare.test.ts and compare-utils.test.ts to use require() calls with eslint-disable. This ensures the tests pass without requiring a breaking API change in the core library.

rockaxorb13:
@mttrbrts yes, right away, I am on it!

rockaxorb13:
Hi @mttrbrts ! Thanks for the review. Really appreciate your time :)

About the commonJS vs ESM situation, I brought up the same concern on a workgroup call 3 weeks ago, I believe at that time @dselman told me that I keep the commonJS syntax in for the time being. 

I'm ready to implement any further changes if need be.

rockaxorb13:
@mttrbrts No, I didn't bring up the typing issues, though if need be I can make the required changes for the same

mttrbrts (approved):

PR #1144, opened 2026-02-28
fix(core): resolve incorrect version check against process.env.npm_package_version

### LINKED ISSUE #1142: Bug[v4]: isCompatibleVersion uses process.env.npm_package_version causing version check crashes in downstream packages
# Bug Report 🐛
### Description
While testing the migration of downstream packages (specifically `template-engine`) to `@accordproject/concerto-core@4.0.0-beta.1`, model validation unexpectedly fails with a version mismatch error.

The engine throws the following error when evaluating models:
`Error: ModelFile expects Concerto version ^3.0.0 but this is 2.7.1`

`2.7.1` is the version of the calling package (`template-engine`), not the installed `concerto-core` package.

### Root Cause
In `packages/concerto-core/src/introspect/modelfile.ts`, the `isCompatibleVersion()` method checks the engine's version using `process.env.npm_package_version`. 

In Node.js, this environment variable evaluates to the version of the project that initiated the `npm` process (the parent project), rather than the version of the `concerto-core` dependency itself. 

**Proof of Concept:** To confirm this was the root cause, I temporarily spoofed the `template-engine` version in its root `package.json` to `3.99.9`. Running the test suite immediately bypassed the crash, mathematically proving the version check was reading the parent directory's package file.

### Proposed Fix
Replace `process.env.npm_package_version` with a dynamically scoped `require` statement to strictly read `concerto-core`'s own `package.json`. 

I will open a PR with this fix targeting the `v4.0.0` branch shortly!

rockaxorb13:
Fixes #1142 

### Description
This PR resolves a critical version-checking bug introduced during the v4 migration. 

Currently, `ModelFile.isCompatibleVersion()` relies on `process.env.npm_package_version`. When `concerto-core` is installed as a dependency in downstream packages (like `template-engine` or `cicero`), this environment variable resolves to the parent package's version, causing valid models to falsely fail the semver check and crash the application.

### Changes Made
* Removed `process.env.npm_package_version` from `packages/concerto-core/src/introspect/modelfile.ts`.
* Replaced it with a dynamically scoped `require('../../package.json')` inside the `isCompatibleVersion` method to guarantee `concerto-core` evaluates against its own true version.

### Testing
* Verified locally that downstream packages can now successfully parse and validate models requiring `^3.0.0` or `^4.0.0`.
* The `concerto` monorepo test suite passes cleanly (`npm ci && npm run build && npm test`).

### DIFF
--- a/packages/concerto-core/src/introspect/modelfile.ts
+++ b/packages/concerto-core/src/introspect/modelfile.ts
@@ -30,6 +30,7 @@ const MapDeclaration = require('./mapdeclaration');
 const ModelUtil = require('../modelutil');
 const Globalize = require('../globalize');
 const Decorated = require('./decorated');
+const packageJson = require('../../package.json');
 
 // Types needed for TypeScript generation.
 /* eslint-disable no-unused-vars */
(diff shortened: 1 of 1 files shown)


rockaxorb13:
Thanks @sanketshevkar! I have made changes to the PR, could you please let me know if this format is fine? Or is there any additional change I am supposed to make?

PR #1184, opened 2026-04-20
fix(vocabulary): improve property existence checks and enhance falsy value handling in tests

### LINKED ISSUE #1183: Vocabulary term lookups silently discard valid falsy values (empty string, null)
# Bug Report 🐛

  Vocabulary term lookups silently discard valid falsy values (empty string, null) and return null instead.

  ## Behavior

| Case | YAML value for myProp       | getTerm result | Expected  | Bug?                                        |
| ---- | --------------------------- | -------------- | --------- | ------------------------------------------- |
| 1    | "" (empty string)           | null           | ""        | Yes                                         |
| 2    | null                        | null           | null      | Ambiguous (null-not-found vs null-as-value) |
| 3    | "null" (string literal)     | "null"         | "null"    | No (control)                                |
| 4    | (no value — YAML undefined) | null           | undefined | Yes                                         |

  ## Possible Solution

  1. Replace `d[propertyName]` with `propertyName in d` when searching for a property by name — checks key existence instead of value truthiness
  2. Replace `if (term)` with `if (term !== null && term !== undefined)` when deciding whether to return a found term or fall back

  ## Steps to Reproduce

`script`:
```js
'use strict';

const { VocabularyManager } = require('./index.js');

const vm = new VocabularyManager();

// ── Case 1: empty string value ────────────────────────────────────────────────
const emptyStringVocab = `
locale: en
namespace: org.test@1.0.0
declarations:
  - MyDecl: 'My Declaration'
    properties:
      - myProp: ""
`;

vm.addVocabulary(emptyStringVocab);
const result1 = vm.getTerm('org.test@1.0.0', 'en', 'MyDecl', 'myProp');

console.log('=== Case 1: myProp value is "" (empty string) ===');
console.log('getTerm result :', JSON.stringify(result1));   // returns null (BUG)
console.log('Expected       :', JSON.stringify(''));         // should be ""
console.log('Bug present    :', result1 !== '', '\n');

// ── Case 2: explicit null value ───────────────────────────────────────────────
const vm2 = new VocabularyManager();

const nullValueVocab = `
locale: en
namespace: org.test@1.0.0
declarations:
  - MyDecl: 'My Declaration'
    properties:
      - myProp: null
`;

vm2.addVocabulary(nullValueVocab);
const result2 = vm2.getTerm('org.test@1.0.0', 'en', 'MyDecl', 'myProp');

console.log('=== Case 2: myProp value is null ===');
console.log('getTerm result :', JSON.stringify(result2));   // returns null (ambiguous — was not found OR was set to null)
console.log('Expected       :', JSON.stringify(null));       // null, but for the wrong reason
console.log('Ambiguous (null-because-not-found vs null-as-value):', result2 === null, '\n');

// ── Case 3: non-empty value (control — must still work) ───────────────────────
const vm3 = new VocabularyManager();

const normalVocab = `
locale: en
namespace: org.test@1.0.0
declarations:
  - MyDecl: 'My Declaration'
    properties:
      - myProp: "null"
`;

vm3.addVocabulary(normalVocab);
const result3 = vm3.getTerm('org.test@1.0.0', 'en', 'MyDecl', 'myProp');

console.log('=== Case 3: myProp value is "null" (control) ===');
console.log('getTerm result :', JSON.stringify(result3));
console.log('Expected       :', JSON.stringify('null'));
console.log('Correct        :', result3 === 'null', '\n');

// ── Case 4: property with no value (YAML undefined) ──────────────────────────
const vm4 = new VocabularyManager();

const undefinedValueVocab = `
locale: en
namespace: org.test@1.0.0
declarations:
  - MyDecl: 'My Declaration'
    properties:
      - myProp:
`;

vm4.addVocabulary(undefinedValueVocab);
const result4 = vm4.getTerm('org.test@1.0.0', 'en', 'MyDecl', 'myProp');

console.log('=== Case 4: myProp has no value (YAML undefined) ===');
console.log('getTerm result :', JSON.stringify(result4));    // returns null (BUG — same as case 2)
console.log('Expected       :', JSON.stringify(undefined));   // undefined (key exists, value absent)
console.log('Bug present    :', result4 !== undefined, '\n');

```

  Context (Environment)

  This affects any use of VocabularyManager where vocabulary authors intentionally use empty strings or null to represent "no label" or "intentionally blank" terms. The silent discard
  makes it impossible to distinguish between a missing term and a deliberately empty one.

  Desktop

  - OS: any
  - Browser: N/A (Node.js library)
  - Version: concerto-vocabulary (current main)

  Detailed Description

  The root cause is two truthiness checks in the vocabulary lookup path:

  In vocabulary.js, when searching for a property by name:
  // Before — reads the value, fails if falsy
  decl.properties.find(d => d[propertyName])

  // After — checks key existence regardless of value
  decl.properties.find(d => propertyName in d)

  In vocabularymanager.js, when deciding whether to return or fall back:
  // Before — treats "" and null as "not found"
  if (term)

  // After — only falls back if truly nothing was found
  if (term !== null && term !== undefined)

  The two bugs compound: even if a property with a falsy value is found in vocabulary.js, it gets discarded again by the truthiness check in vocabularymanager.js.

  Possible Implementation

  See above under Possible Solution — both are single-line changes with no impact on the existing API surface or behaviour for non-falsy values.

mon7792:
# Closes #1183 

  Fix vocabulary term lookup failing when term value is falsy (empty string, `0`, `false`, `null`)

  ### Changes
  - Use `propertyName in d` instead of `d[propertyName]` in `vocabulary.js` to correctly detect property existence regardless of value
  - Use `term !== null && term !== undefined` instead of `if (term)` in `vocabularymanager.js` to avoid skipping valid falsy terms
  - Add test cases in `vocabulary.js` and `vocabularymanager.js` test suites covering falsy vocabulary term values

  ### Flags
  - The `if (term)` pattern in `VocabularyManager` caused fallback to a parent/default vocabulary even when the term was explicitly set to an empty string or other falsy value — this is a
  silent correctness bug, not a crash

  ### Screenshots or Video
  N/A

  ### Related Issues
  - Issue #<NUMBER>

  ### Author Checklist

  Two things you need to fill in manually:
  1. Issue number — if there's a GitHub issue tracking this bug, add it to Closes #<NUMBER> and Related Issues
  2. DCO sign-off — if your commits don't have --signoff, you'll need to amend them before the PR can be merged

  To check if your commit has a sign-off:
  git log -1 --format="%B"
  If it's missing Signed-off-by:, let me know and I'll help you add it.


packages/concerto-vocabulary/lib/vocabularymanager.js
```diff
@@ -242,7 +242,7 @@ class VocabularyManager {
         if (voc) {
             term = voc.getTerm(declarationName, propertyName, identifier);
         }
-        if (term) {
+        if (term !== null && term !== undefined) {
```
muhabdulkadir:
```suggestion
        if (!!term) {
```

packages/concerto-vocabulary/lib/vocabularymanager.js
```diff
@@ -271,7 +271,7 @@ class VocabularyManager {
         if (voc) {
             term = voc.getElementTerms(declarationName, propertyName);
         }
-        if (term) {
+        if (term !== null && term !== undefined) {
```
muhabdulkadir:
```suggestion
        if (!!term) {
```

packages/concerto-vocabulary/lib/vocabulary.js
```diff
@@ -140,7 +140,7 @@ class Vocabulary {
             return identifier ? decl[identifier] : decl[declarationName];
         }
         else {
-            const property = decl.properties ? decl.properties.find(d => d[propertyName]) : null;
+            const property = decl.properties ? decl.properties.find(d => propertyName in d) : null;
```
muhabdulkadir: Why change this behaviour? This can have an adverse effect on other consumers of this.

PR #1199, opened 2026-04-29
fix(*): add v3 model compatibility support and disable adding, creating and updating unversioned models

muhabdulkadir:

Concerto v4 currently rejects models that declare a concerto version range targeting v3, even when those models are otherwise valid and semantically compatible with the v4 runtime. As a result, existing v3 models fail at load time and cannot be reused in v4 environments.

This change restores practical backward compatibility for v3-authored models while keeping strict validation for unsupported versions.

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist

### DIFF
--- a/packages/concerto-core/src/introspect/modelfile.ts
+++ b/packages/concerto-core/src/introspect/modelfile.ts
@@ -718,6 +718,11 @@ class ModelFile extends Decorated {
         this.namespace = ast.namespace;
         this.version = nsInfo.version;
 
+        // In v4, all non-system models must declare a namespace version (e.g., @1.0.0)
+        if (!this.version && !this.isSystemModelFile()) {
+            throw new Error(`Cannot create a ModelFile with an unversioned namespace: ${ast.namespace}. All models must specify a version (e.g., @1.0.0).`);
+        }
+
         // Make sure to clone imports since we will add built-in imports
         const imports = ast.imports ? ast.imports.concat([]) : [];
 
--- a/packages/concerto-core/src/introspect/modelfile.ts
+++ b/packages/concerto-core/src/introspect/modelfile.ts
@@ -673,14 +673,13 @@ class ModelFile extends Decorated {
      * Models targeting an older major version are considered backward-compatible
      * with newer runtimes (e.g. a model declaring "^3.0.0" loads under v4).
      */
-isCompatibleVersion() {
+    isCompatibleVersion() {
         if (this.ast.concertoVersion) {
             if (semver.satisfies(packageJson.version, this.ast.concertoVersion, { includePrerelease: true })) {
                 this.concertoVersion = this.ast.concertoVersion;
             } else {
-                // Allow models that target an older major version to load under a newer runtime.
-                const minVersion = semver.minVersion(this.ast.concertoVersion);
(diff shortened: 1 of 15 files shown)


mttrbrts (commented):
IMO we should be stronger here @muhabdulkadir, we should specifically allow v3 versions to run with the v4 runtime, but nothing else. e.g. v0.82 is not compatible with v3

packages/concerto-core/test/semver.js
```diff
@@ -56,7 +56,7 @@ describe('Semantic Versioning', () => {
             (() => {
                 modelManager = new ModelManager();
                 modelManager.addCTOModel('namespace test', 'test.cto');
-            }).should.throw(/Cannot add an unversioned namespace/);
+            }).should.throw(/Cannot create a ModelFile with an unversioned namespace/);
```
muhabdulkadir, line 61: When `addCTOModel()` is called, it internally calls `addModelFile()`, which parses the CTO by calling `ModelFile.fromAst()`. Since I added the strict validation at the `ModelFile.fromAst()` level (to enforce it at all entry points, not just ModelManager), the error is now thrown there first.

packages/concerto-core/src/introspect/modelfile.ts
```diff
@@ -669,14 +669,21 @@ class ModelFile extends Decorated {
...
         if (this.ast.concertoVersion) {
             if (semver.satisfies(packageJson.version, this.ast.concertoVersion, { includePrerelease: true })) {
                 this.concertoVersion = this.ast.concertoVersion;
             } else {
-                throw new Error(`ModelFile expects Concerto version ${this.ast.concertoVersion} but this is ${packageJson.version}`);
+                // Allow v3 models to load under newer runtimes
+                if (semver.minSatisfying(['3.0.0'], this.ast.concertoVersion)) {
+                    this.concertoVersion = this.ast.concertoVersion;
+                } else {
+                    throw new Error(`ModelFile expects Concerto version ${this.ast.concertoVersion} but this is ${packageJson.version}`);
```
jamieshorten: This would say it expects a version of v4.0.0 right? rather than the 3.0.0 that is actually supported, right?

jamieshorten: Do we want to reference the supported language version rather?

muhabdulkadir: I'm not sure I fully grasp this, but in the event that the model is not `>=v3` then the error would be `ModelFile expects Concerto version vX but this is v4`

jamieshorten: Right, I was hoping we could be more specific, like: "This version of Concerto supports a language version of v3.0.0 or greater, but this model is for v2.1.3"

Overall I think the current message isn't to bad though.

muhabdulkadir: Not a bad idea. https://github.com/accordproject/concerto/pull/1199/commits/f0e2c7d882db30f2b11645475789c97c9c82cdab

jamieshorten (approved):

mttrbrts (approved):

PR #1204, opened 2026-05-02
fix(concerto-linter): validate namespace version across all models

### LINKED ISSUE #1203: Namespace version validation only checks first model
The namespace version validation in namespace-version.ts currently checks only the first model using $.models[0].namespace.

This causes incorrect behavior when multiple models are present, as other models are not validated.

Expected Behavior
Validation should apply to all models, not just the first one.

Proposed Fix
Update the selector from:

$.models[0].namespace

to:

$.models[*].namespace

I’d like to work on fixing this issue and add a test case to cover multi-model validation.

atulXdev:
# Closes #1203

The `namespace-version` rule in the default ruleset used `$.models[0].namespace` which only checked the first model in the AST. This means when linting multiple Concerto models together, only the first model's namespace was validated for versioning — any subsequent models without a version silently passed.

### Changes
- Changed JSONPath `given` from `$.models[0].namespace` to `$.models[*].namespace` in the namespace-version rule, consistent with all other rules in the default ruleset
- Added a multi-model test case that constructs an AST with two models (one valid, one missing version) to verify the fix

### Flags
- This is a 1-line behavioral fix + test — no breaking changes
- All existing tests continue to pass
- Consistent with the pattern used by all 8 other rules in the default ruleset

### Screenshots or Video
N/A

### Related Issues
- Closes #1203

### Author Checklist
- [x] Merging to `main` from `fork:fix/namespace-version-all-models`

### DIFF
--- a/packages/concerto-linter/default-ruleset/src/namespace-version.ts
+++ b/packages/concerto-linter/default-ruleset/src/namespace-version.ts
@@ -22,7 +22,7 @@ import { pattern } from '@stoplight/spectral-functions';
  */
 export default {
     description: 'namespace should specify a version.',
-    given: '$.models[0].namespace',
+    given: '$.models[*].namespace',
     message: 'namespace \'{{value}}\' should specify a version.',
     severity: 0, // 0 = error, 1 = warning, 2 = info, 3 = hint
     then: {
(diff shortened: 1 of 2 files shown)


mttrbrts (approved):
Thanks for your contribution @atulXdev

PR #1215, opened 2026-05-07
fix(concerto-core): add opt-out for reserved system type name clashes

muhabdulkadir:

Adds a targeted compatibility opt-out for reserved Concerto system type name clashes so dependent systems can migrate without losing other validation protections.

By default, behavior is unchanged. When enabled, only clashes against reserved system types are allowed.

### Changes

- Added a new ModelManager option: `dangerouslyAllowReservedSystemTypeNamesInUserModels ` (default false)
- Wired option into declaration clash validation:
  -  keeps existing clash validation
  -  skips only when clash is with reserved system types in concerto namespace:
      -  Concept, Asset, Transaction, Participant, Event
- Added tests for:
  -  default behavior still rejecting system type-name clashes
  -  opt-out allowing reserved system type-name clashes
  -  opt-out still rejecting non-system ambiguous type clashes
-  Documented the new option in constructor JSDoc.

### Flags

- <ONE> This is intentionally scoped as a compatibility escape hatch.
- <TWO> Validation remains strict for non-system type collisions.
- <THREE> No API behavior changes unless `dangerouslyAllowReservedSystemTypeNamesInUserModels ` is explicitly enabled.

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist

### DIFF
--- a/packages/concerto-core/src/introspect/declaration.ts
+++ b/packages/concerto-core/src/introspect/declaration.ts
@@ -84,8 +84,8 @@ class Declaration extends Decorated {
 
         // #648 - check for clashes against imported types
         if (modelFile.isImportedType(this.getName())) {
-            const dangerouslyAllowReservedSystemTypeNames = Boolean(modelFile.getModelManager()?.options?.dangerouslyAllowReservedSystemTypeNames);
-            if (dangerouslyAllowReservedSystemTypeNames && this.isReservedSystemTypeImport(modelFile, this.getName())) {
+            const dangerouslyAllowReservedSystemTypeNamesInUserModels = Boolean(modelFile.getModelManager()?.options?.dangerouslyAllowReservedSystemTypeNamesInUserModels);
+            if (dangerouslyAllowReservedSystemTypeNamesInUserModels && this.isReservedSystemTypeImport(modelFile, this.getName())) {
                 return;
             }
 
--- a/packages/concerto-core/src/introspect/declaration.ts
+++ b/packages/concerto-core/src/introspect/declaration.ts
@@ -17,6 +17,7 @@
 const Decorated = require('./decorated');
 const ModelUtil = require('../modelutil');
 const IllegalModelException = require('./illegalmodelexception');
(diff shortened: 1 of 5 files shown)


packages/concerto-core/src/types.ts
```diff
@@ -18,6 +18,7 @@ export interface ModelManagerOptions {
     regExp?: RegExp;
     metamodelValidation?: boolean;
     addMetamodel?: boolean;
+    allowReservedSystemTypeNames?: boolean;
```
jamieshorten: lets change this to something like `dangerouslyAllowReservedSystemTypeNames` to make clear that this is risky. We should also make clear that this is a transitional parameter and will be removed at some point.

muhabdulkadir: https://github.com/accordproject/concerto/pull/1215/commits/976e567faad05434137305d01a701fc4215e27fd

packages/concerto-core/src/introspect/declaration.ts
```diff
@@ -80,13 +80,44 @@ class Declaration extends Decorated {
...
     validate(...args) {
         super.validate(...args);
+        const modelFile = this.getModelFile();
 
         // #648 - check for clashes against imported types
-        if (this.getModelFile().isImportedType(this.getName())){
+        if (modelFile.isImportedType(this.getName())) {
+            const dangerouslyAllowReservedSystemTypeNames = Boolean(modelFile.getModelManager()?.options?.dangerouslyAllowReservedSystemTypeNames);
+            if (dangerouslyAllowReservedSystemTypeNames && this.isReservedSystemTypeImport(modelFile, this.getName())) {
+                return;
```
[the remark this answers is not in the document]
muhabdulkadir: I investigated this and can confirm this is pre-existing behavior. I ran probe scripts in both v3 and v4; the metamodel AST correctly records the import declarations in both runtimes. The issue is purely in the `getType()` lookup precedence, which always favours the system namespace.

mttrbrts (approved):

PR #1236, opened 2026-05-25
fix(concerto-core): handle RelationshipMapValueType in map value type processing

muhabdulkadir:

This change fixes a gap in map value type introspection where `RelationshipMapValueType` was not handled in `MapValueType.processType()`. As a result, `this.type` could remain unset and `getType()` could return `undefined` for relationship map values. The fix adds the missing switch branch and routes it through the same `TypeIdentifier` handling.

### Related Issues
- Issue #<NUMBER>
- Pull Request #<NUMBER>

### Author Checklist

### DIFF
--- a/packages/concerto-core/src/introspect/mapvaluetype.ts
+++ b/packages/concerto-core/src/introspect/mapvaluetype.ts
@@ -91,6 +91,7 @@ class MapValueType extends Decorated {
         let decl;
         switch(this.ast.$class) {
         case `${MetaModelNamespace}.ObjectMapValueType`:
+        case `${MetaModelNamespace}.RelationshipMapValueType`:
 
             // ObjectMapValueType must have TypeIdentifier.
             if (!('type' in ast)) {
--- a/packages/concerto-core/test/introspect/mapdeclaration.js
+++ b/packages/concerto-core/test/introspect/mapdeclaration.js
@@ -143,6 +143,13 @@ describe('MapDeclaration', () => {
 
         it('should validate when map value is an identified concept declaration', () => {
             let decl = introspectUtils.loadLastDeclaration('test/data/parser/mapdeclaration/mapdeclaration.goodvalue.declaration.identified.concept.cto', MapDeclaration);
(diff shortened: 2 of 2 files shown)


packages/concerto-core/src/introspect/mapvaluetype.ts
```diff
@@ -91,6 +91,7 @@ class MapValueType extends Decorated {
         let decl;
         switch(this.ast.$class) {
         case `${MetaModelNamespace}.ObjectMapValueType`:
+        case `${MetaModelNamespace}.RelationshipMapValueType`:
```
jamieshorten, line 94: what's the fall through behaviour for switch in JS/TS? This reads like `ObjectMapValueType` case now does nothing. Can we make this clearer?

muhabdulkadir: the behaviour of both `ObjectMapValueType` and `RelationshipMapValueType` are exactly the same, this is the simplest was to write this.

packages/concerto-core/test/introspect/mapdeclaration.js
```diff
@@ -143,6 +143,13 @@ describe('MapDeclaration', () => {
 
         it('should validate when map value is an identified concept declaration', () => {
             let decl = introspectUtils.loadLastDeclaration('test/data/parser/mapdeclaration/mapdeclaration.goodvalue.declaration.identified.concept.cto', MapDeclaration);
+            expect(decl.getValue().getType()).to.equal('Person');
+            decl.validate();
+        });
+
+        it('should validate when map value is a relationship declaration', () => {
+            let decl = introspectUtils.loadLastDeclaration('test/data/parser/mapdeclaration/mapdeclaration.badvalue.declaration.relationship.cto', MapDeclaration);
```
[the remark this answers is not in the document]
muhabdulkadir: the is correct. The model representation in the `mapdeclaration.badvalue.declaration.relationship.cto` file is correct.

jamieshorten (approved):

PR #1242, opened 2026-05-28
fix(*): clone imports in ModelFile.filter to prevent mutation

muhabdulkadir:

`ModelFile.filter` used a shallow spread `({ ...this.ast })` to create the filtered AST, which shared import objects with the original model file. `BaseModelManager.filter` then mutated `imp.types` in-place on those shared objects, corrupting the source model files

### Author Checklist

### DIFF
--- a/packages/concerto-core/src/introspect/modelfile.ts
+++ b/packages/concerto-core/src/introspect/modelfile.ts
@@ -886,6 +886,7 @@ class ModelFile extends Decorated {
         const ast = {
             ...this.ast,
             declarations: declarations,
+            imports: this.ast.imports?.map(imp => ({...imp})),
         };
         if (ast.declarations?.length > 0){
             return new ModelFile(modelManager, ast, undefined, this.fileName);


mttrbrts:
Is there a risk that we see corrupted imports because of the same issue with another shallow spread here.

muhabdulkadir:
> Is there a risk that we see corrupted imports because of the same issue with another shallow spread here.

The shallow clone is sufficient because the only mutation site `(imp.types = imp.types.filter(...))` is a property reassignment, not an in-place array mutation. No other code path mutates nested properties on the filtered AST's imports.

packages/concerto-core/src/introspect/modelfile.ts
mttrbrts, line 882: We likely have the same issue here

muhabdulkadir: No, this is fine. The `declaration.ast` references are placed into a new declarations array `(line 875)` and then spread into a new AST object `(line 887: ...this.ast)`. The `new ModelFile` constructed at `line 892` reads from these AST nodes to build its own object model but doesn't mutate them.

mttrbrts (approved):

PR #1246, opened 2026-06-02
fix(concertino): avoid browser import failure

### LINKED ISSUE #1245: concertino: fs.readFileSync at module load time breaks browser/bundler environments
## Summary

`@accordproject/concertino` calls `fs.readFileSync` at **module load time** in `dist/index.js` to parse the AJV validation schema. This fails in browser environments (Vite, webpack, etc.) where `fs` is unavailable, causing the entire package import to throw — including the non-validating `convertToConcertino` function which has no need for the schema at all.

## Affected code

`packages/concertino/dist/index.js` (generated from `src/index.ts`):

```js
const schema = JSON.parse((0, fs_1.readFileSync)((0, path_1.join)(__dirname, 'spec/concertino.schema.json'), 'utf-8'));
// ...
this.validate = this.ajv.compile(schema); // inside ConcertinoConverter constructor
```

The `readFileSync` is executed unconditionally when the module is first imported, not lazily when validation is requested.

## Impact

Any browser-based tooling importing `@accordproject/concertino` sees:

```
Module "fs" has been externalized for browser compatibility.
Cannot access "fs.readFileSync" in client code.
```

This silently breaks `ConcertinoConverter` and `convertToConcertino` in browser builds even though `convertToConcertino` itself has no file-system dependency.

## Workaround

Import from the subpath that has no `fs` dependency:

```js
import { convertToConcertino } from "@accordproject/concertino/dist/concertinoSerializer";
```

This works but relies on an internal, undocumented path.

## Suggested fix

Move the schema load (and AJV compilation) inside `ConcertinoConverter.isValid()`, which is the only method that actually needs it. `fromConcertoMetamodel` and `toConcertoMetamodel` can remain file-system-free:

```ts
// Lazy-load once on first call to isValid()
private getValidator() {
  if (!this.validate) {
    const schema = JSON.parse(fs.readFileSync(...));
    this.validate = this.ajv.compile(schema);
  }
  return this.validate;
}
```

Alternatively, the JSON schema could be inlined as a plain JS object import (no `fs` needed) or the package could ship a separate browser-safe entry point that omits validation.

## Environment

- `@accordproject/concertino` v4.1.3
- Vite v6.4.2 (browser build)

Rishabh060105:
# Closes #1245
Ensure `@accordproject/concertino` can be imported safely in browser and bundler environments by removing top-level filesystem access from the package entrypoint and compiling the AJV validator only when validation is requested.

### Changes
- Replace eager schema loading via `fs.readFileSync(...)` with a JSON module import and lazy validator initialization inside `ConcertinoConverter`.
- Add regression coverage for lazy validator compilation and cached reuse, and enable `resolveJsonModule` for the `concertino` TypeScript build.

### Flags
- The package root no longer imports `fs` or `path`, which avoids browser bundler failures when consumers only need `fromConcertoMetamodel` or `toConcertoMetamodel`.
- Validation-related work and any schema-compilation failures now move from `ConcertinoConverter` construction time to the first `isValid()` call.

### Screenshots or Video
- N/A

### Related Issues
- Issue #1245
- Pull Request #1235

### Author Checklist
- [x] Merging to `main` from `Rishabh060105:issue-1245-concertino-browser-import`

### DIFF
--- a/packages/concertino/src/index.ts
+++ b/packages/concertino/src/index.ts
@@ -20,11 +20,10 @@
 /* eslint-disable valid-jsdoc */
 import { IModels } from '@accordproject/concerto-metamodel';
 import { IConcertino } from './spec/concertino.metamodel@4.0.0-alpha.2';
+import schema from './spec/concertino.schema.json';
 import { convertToConcertino } from './concertinoSerializer';
 import { convertToMetamodel } from './metamodelSerializer';
 import Ajv, { ValidateFunction } from 'ajv';
-import { readFileSync } from 'fs';
-import { join } from 'path';
 
 /**
  * Conversion options for Concertino format.
@@ -36,14 +35,12 @@ export interface ConcertinoOptions {
   version?: string;
 }
 
-const schema = JSON.parse(readFileSync(join(__dirname, 'spec/concertino.schema.json'), 'utf-8'));
-
 /**
  * Concertino utility class for converting between Concerto metamodel and Concertino format.
  */
 export class ConcertinoConverter {
     private options: ConcertinoOptions;
-    private validate: ValidateFunction;
+    private validate?: ValidateFunction;
     private ajv: Ajv;
 
     /**
@@ -56,7 +53,6 @@ export class ConcertinoConverter {
             ...options
         };
         this.ajv = new Ajv();
-        this.validate = this.ajv.compile(schema);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Rishabh060105:
Hi @mttrbrts,  The change removes the top-level `fs/path` schema read from the package root, imports the schema as JSON instead, and compiles the AJV validator lazily on the first `isValid()` call. `fromConcertoMetamodel` and `toConcertoMetamodel` remain unchanged in behaviour, and I added regression coverage to verify lazy compilation and validator reuse.

mttrbrts (approved):

PR #1253, opened 2026-06-05
perf(concerto-core): optimize   ModelManager and ModelFile filter   operations

muhabdulkadir:

- `BaseModelManager.filter` previously constructed each ModelFile 2–3× per file (once to filter, once for import cleanup, plus re-validation). Now exactly 1× — each avoided construction skips a full AST parse + semantic validation pass, which dominates wall-clock cost.
- `ModelFile.filter` is now self-contained: returns a fully correct ModelFile with cleaned imports, requiring only a predicate and target model manager.
- Added `disableValidation` option to `BaseModelManager.filter`.
- - Replaced O(M × R × T) nested import cleanup with O(N + T) linear approach using predicate-based import validation
- When all types in an `ImportTypes` entry were filtered out, the empty entry remained in the AST, causing the CTO printer to emit invalid `import ns.{}` syntax. Now removes the entire import when no types remain.
- Handles ImportTypeFrom (single named import with URI) in addition to ImportType.

### Complexity improvement

- **Before**: for each model file, iterated all removed FQNs and called `getImports()` (which rebuilds the full FQN list each call)

- **After**: O(N + T) — linear in total declarations plus total import type references using `O(1) getModelFile + getLocalType` lookups

Where 
  - N = total declarations, 
  - T = total import type references, 
  - M = model files,
  - R = removed declarations, 

### Author Checklist

### DIFF
--- a/packages/concerto-core/src/basemodelmanager.ts
+++ b/packages/concerto-core/src/basemodelmanager.ts
@@ -852,9 +852,11 @@ class BaseModelManager {
      * ModelFiles with no declarations after filtering will be removed.
      *
      * @param {FilterFunction} predicate - the filter function over a Declaration object
+     * @param {Object} [options] - options for the filter method
+     * @param {boolean} [options.disableValidation] — If true then the model files are not validated
      * @returns {BaseModelManager} - the filtered ModelManager
      */
-    filter(predicate){
+    filter(predicate, options?){
         const modelManager = new BaseModelManager({...this.options}, this.processFile);
         const filteredModels: ModelFileInstance[] = [];
 
@@ -868,7 +870,7 @@ class BaseModelManager {
             }
         }
 
-        modelManager.addModelFiles(filteredModels);
+        modelManager.addModelFiles(filteredModels, undefined, options?.disableValidation);
         return modelManager;
     }
 }
--- a/packages/concerto-core/src/basemodelmanager.ts
+++ b/packages/concerto-core/src/basemodelmanager.ts
@@ -209,9 +209,9 @@ class BaseModelManager {
              throw new Error(`Failed to load getDecoratorModel. Got: ${typeof getDecoratorModel}`);
         }
         const {decoratorModelAst, decoratorModelCto, decoratorModelFile} = getDecoratorModel();
-        
+
         const m = new ModelFile(this, decoratorModelAst, decoratorModelCto, decoratorModelFile);
- 
+
         this.addModelFile(m, decoratorModelCto, decoratorModelFile, true);
     }
 
@@ -856,59 +856,19 @@ class BaseModelManager {
      */
     filter(predicate){
         const modelManager = new BaseModelManager({...this.options}, this.processFile);
-        const removedFqns = []; // the list of FQN of types that have been removed
+        const filteredModels: ModelFileInstance[] = [];
 
-        // remove the types from model files, populating removedFqns
-        let filteredModels = Object.values(this.modelFiles)
-            .map((modelFile: ModelFileInstance) => modelFile.filter(predicate, modelManager, removedFqns))
-            .filter(Boolean);
-
-        // remove concerto model files - as these are automatically added
-        // when we recreate the model manager below
-        filteredModels = filteredModels.filter(mf => !mf.isSystemModelFile());
-
-        // now update filteredModels to remove any imports of removed types
-        const modelsWithValidImports = filteredModels.map( modelFile => {
-            const ast = modelFile.getAst();
-            let modified = false;
-            removedFqns.forEach( removedFqn => {
-                const ns = ModelUtil.getNamespace(removedFqn);
-                const isSystemImport = ns.startsWith('concerto@') || ns === 'concerto';
-                if(!isSystemImport && modelFile.getImports().includes(removedFqn)) {
-                    const removeName = ModelUtil.getShortName(removedFqn);
-                    const removeNamespace = ModelUtil.getNamespace(removedFqn);
-                    ast.imports = ast.imports.filter(imp => {
-                        const remove = ModelUtil.getShortName(imp.$class) === 'ImportType' &&
-                            imp.name === removeName &&
-                            imp.namespace === removeNamespace;
-                        if(remove) {
-                            modified = true;
-                        }
-                        return !remove;
-                    });
-                    ast.imports.forEach( imp => {
-                        if(imp.namespace === removeNamespace) {
-                            if(ModelUtil.getShortName(imp.$class) === 'ImportTypes') {
-                                imp.types = imp.types.filter((type) => {
-                                    const remove = (type === removeName);
-                                    if(remove) {
-                                        modified = true;
-                                    }
-                                    return !remove;
-                                });
-                            }
-                        }
-                    });
-                }
-            });
-            if(modified) {
-                return new ModelFile(this, ast, undefined, modelFile.fileName);
+        for (const modelFile of Object.values(this.modelFiles) as ModelFileInstance[]) {
+            if (modelFile.isSystemModelFile()) {
+                continue;
             }
-            else {
-                return modelFile;
+            const filtered = modelFile.filter(predicate, modelManager);
+            if (filtered) {
+                filteredModels.push(filtered);
             }
-        });
-        modelManager.addModelFiles(modelsWithValidImports);
+        }
+
+        modelManager.addModelFiles(filteredModels);
         return modelManager;
     }
 }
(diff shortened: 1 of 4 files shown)


packages/concerto-core/src/basemodelmanager.ts
```diff
@@ -856,59 +856,19 @@ class BaseModelManager {
...
-                return new ModelFile(this, ast, undefined, modelFile.fileName);
+        const filteredModels: ModelFileInstance[] = [];
+
+        for (const modelFile of Object.values(this.modelFiles) as ModelFileInstance[]) {
+            if (modelFile.isSystemModelFile()) {
+                continue;
             }
-            else {
-                return modelFile;
+            const filtered = modelFile.filter(predicate, modelManager);
```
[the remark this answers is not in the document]
muhabdulkadir: this is redundant because system modelfile is already filtered out.

packages/concerto-core/src/introspect/modelfile.ts
```diff
@@ -867,31 +867,56 @@ class ModelFile extends Decorated {
...
+                if (ModelUtil.getShortName(imp.$class) === 'ImportType') {
+                    const sourceFile = sourceManager.getModelFile(imp.namespace);
+                    if (!sourceFile) {
+                        return true;
+                    }
+                    const decl = sourceFile.getLocalType(imp.name);
+                    return !decl || predicate(decl);
+                }
+                return true;
+            });
```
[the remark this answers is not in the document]
muhabdulkadir: `ImportTypeFrom` is handled similar to `ImportType`. This case is already handled.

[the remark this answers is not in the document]
muhabdulkadir: Added a test to verify it

packages/concerto-core/src/introspect/modelfile.ts
```diff
@@ -867,31 +867,58 @@ class ModelFile extends Decorated {
...
         const ast = {
             ...this.ast,
-            declarations: declarations,
+            declarations,
             imports: this.ast.imports?.map(imp => ({...imp})),
         };
-        if (ast.declarations?.length > 0){
-            return new ModelFile(modelManager, ast, undefined, this.fileName);
+
+        if (ast.imports) {
```
mttrbrts, line 895: **Behavioral change: the system-import guard is gone.**

The old `BaseModelManager.filter` explicitly never stripped system imports:
```js
const isSystemImport = ns.startsWith('concerto@') || ns === 'concerto';
if(!isSystemImport && ...) { /* remove */ }
```
Here there's no such guard — `predicate(decl)` is run against the resolved source declaration regardless of namespace, so an explicit `import concerto@1.0.0.Concept` gets stripped from the AST whenever the predicate rejects system types (which a namespace-scoped predicate typically does).

I tested this locally and it's **benign**: `validateModelFiles()` still passes and `getImports()` still lists the concerto types, because the system import is re-added implicitly when the filtered manager is rebuilt. So this is a note, not a blocker — but it's an untested change in behavior. Worth either restoring the `concerto@`/`concerto` short-circuit for clarity, or adding a test that filters a file with an explicit system import so the intent is pinned down.

packages/concerto-core/src/introspect/modelfile.ts
```diff
@@ -867,31 +867,58 @@ class ModelFile extends Decorated {
...
      */
-    filter(predicate, modelManager, removedDeclarations){
-        let declarations: any[] = []; // ast for all included declarations
-        this.declarations?.forEach( declaration => {
-            const included = predicate(declaration);
-            if(!included) {
-                removedDeclarations.push(declaration.getFullyQualifiedName());
-            }
-            else {
+    filter(predicate, modelManager){
```
mttrbrts, line 877: The `@private` tag was dropped (along with the `removedDeclarations` param). Was making `filter` a documented/public method intentional? It's only called from `BaseModelManager.filter`. No type-surface impact (there was never a TS `private` modifier, so it's already in the `.d.ts`), but removing `@private` does surface it in the generated jsdoc. If it's meant to stay internal, worth keeping the tag.

packages/concerto-core/src/introspect/modelfile.ts
```diff
@@ -867,31 +867,58 @@ class ModelFile extends Decorated {
...
+        const declarations: any[] = [];
+        for (const declaration of this.declarations) {
```
mttrbrts, line 879: Minor: the previous code used `this.declarations?.forEach(...)`. Dropping the optional chain means this `for...of` throws if `declarations` is ever undefined rather than returning `null`. In practice `declarations` is always initialized in the constructor, so this is likely fine — just flagging the dropped guard.

packages/concerto-core/src/introspect/modelfile.ts
```diff
@@ -867,31 +867,58 @@ class ModelFile extends Decorated {
...
+                    if (!sourceFile) {
+                        return true;
+                    }
+                    const decl = sourceFile.getLocalType(imp.name);
+                    return !decl || predicate(decl);
+                }
+                return true;
+            });
+
+            ast.imports = ast.imports.filter(imp => {
```
mttrbrts: Nit: the two `ast.imports.filter(...)` passes (ImportType, then ImportTypes) could be a single pass with a `switch`/branch on the short name — one iteration, slightly less to read. Optional.

Also a contract note worth a doc line: the predicate is now invoked on declarations from *other* files (the import sources), not just this file's own declarations. That's correct for any pure predicate, but the JSDoc should probably state the predicate must be total/side-effect-free across imported types too.

mttrbrts (commented):
Nice refactor — making `ModelFile.filter` self-contained reads much more clearly than threading a shared `removedFqns` array through `BaseModelManager.filter`, and the complexity win is real (per-import map lookups via `getModelFile`/`getLocalType` instead of the nested re-scan).

I checked this out locally: `concerto-core` builds, `npm run lint` is clean, and the full `modelfile.js` + `modelmanager.js` suites pass (165 passing), including the pre-existing `ModelManager#filter` "remove imports for filtered types" regression test. Pruning the whole `ImportTypes` entry when its last type is filtered out (`return imp.types.length > 0`) is also a nice improvement over the old code, which left behind an empty `import ns.{}`.

A few things worth a look before merge — one behavioral change and some minor nits (details inline). Nothing blocking from my side.

mttrbrts:
One note on the complexity framing in the description — the big-O is accurate but it measures the lighter half of the work, so I'd lead with a different headline.

The `O(M × R × I)` → `O(N + T)` claim holds (I verified `getModelFile` is an O(1) object lookup and `getLocalType` an O(1) `Map` lookup, so the per-import resolution really is linear). If anything it understates the old cost: the old cleanup called `getImports()` *inside* the `removedFqns.forEach`, and `getImports()` rebuilds the full FQN list every call, so it was closer to `O(M × R × T)`.

But that big-O only covers the **import-bookkeeping layer**. The cost that dominates both the old and new code is `new ModelFile(ast, …)` reconstruction plus `addModelFiles(...)` validation — each is ≥ O(N) with much heavier constants (re-parsing the AST, resolving types, full semantic validation) than the import scan. The asymptotic win on the bookkeeping is genuine, but it's not what moves the wall-clock.

The real benefit is the one in your second bullet: **collapsing 2–3× `ModelFile` constructions per file down to exactly 1×.** Each avoided construction skips a full parse + validate pass, and that's the expensive thing. I'd promote that to the headline and treat the `O(N + T)` line as supporting detail.

Minor, while you're touching the description: in `M × R × I`, `R` is *total* removed declarations but `I` is *per-file* imports — the symbols aren't on the same footing. And the new path now invokes the predicate `O(N + T)` times (once per declaration **plus** once per resolved import type) vs `O(N)` before, since the old cleanup used string matching rather than re-running the predicate — fine for an O(1) predicate, worth a footnote if predicates can be heavy.

mttrbrts (approved):

packages/concerto-core/src/introspect/modelfile.ts
muhabdulkadir: Adding this back.

packages/concerto-core/src/introspect/modelfile.ts
muhabdulkadir: Adding this back.

packages/concerto-core/src/introspect/modelfile.ts
muhabdulkadir: the guard isn't needed, `this.declarations` is initialised so should never be `undefined`.

packages/concerto-core/src/introspect/modelfile.ts
muhabdulkadir: would combine this.

PR #1262, opened 2026-06-14
feat(concerto): add tree-shakeable package entrypoints

### LINKED ISSUE #1261: feat: add tree-shaking support for web bundles
## Summary

The Concerto packages are not currently optimized for tree-shaking in web bundles. This means consumers using bundlers (webpack, rollup, esbuild, Vite, etc.) will receive the full library even when they only use a small subset of its functionality.

### 1. Missing `sideEffects` declaration (all 7 packages)

None of the packages declare `"sideEffects"` in their `package.json`. Without this, bundlers must conservatively assume every module has side effects and cannot safely eliminate unused code.

### 2. Missing `exports` field (all 7 packages)

No package uses the `"exports"` field, which is required for modern conditional entry points (e.g. resolving ESM vs CJS based on the consumer's environment).

### 3. No ESM distribution

The `"browser"` field in `concerto-core`, `concerto-cto`, and `concerto-util` points to UMD bundles (`dist/concerto-*.js`), which are not tree-shakeable. The remaining 4 packages have no browser target at all.

### 4. CommonJS-only output in 4 packages

`concerto-util`, `concerto-analysis`, `concertino`, and `concerto-linter` compile to CommonJS only. CommonJS cannot be statically analysed by bundlers for tree-shaking.

## Real Side Effect: `dayjs.extend()` calls

Several files in `concerto-core` call `dayjs.extend(plugin)` at **module load time**:

- `datetimeutil.ts`
- `jsonpopulator.ts`
- `valuegenerator.ts`
- `resourcevalidator.ts`
- `factory.ts`
- `typed.ts`

These are genuine side effects. If any of these files are eliminated by a tree-shaker, dayjs plugins (UTC, quarterOfYear, duration, minMax) won't be registered, causing silent runtime failures. This must be addressed before marking packages as side-effect-free.

### Step 1: Consolidate dayjs plugin registration

Move all `dayjs.extend()` calls into a single module (e.g. `src/dayjs-setup.ts`) imported once, and list it in the `sideEffects` array:

```json
// concerto-core/package.json
"sideEffects": ["dist/dayjs-setup.js"]
```

All other packages can then safely use:

```json
"sideEffects": false
```

### Step 2: Add `exports` field

```json
"exports": {
  ".": {
    "import": "./dist/index.mjs",
    "require": "./dist/index.js",
    "browser": "./dist/concerto-core.esm.js"
  }
}
```

> No deep imports were found in the codebase, so adding an `exports` field will not break existing consumers.

### Step 3: Generate ESM output

Add a dual-build step (e.g. rollup or `tsc` with `module: ESNext`) to emit `.mjs` / ESM bundles alongside the existing CJS output. Replace the UMD `browser` bundles with ESM equivalents.

### Step 4: Upgrade remaining packages to dual CJS+ESM

Update the TypeScript configs for `concerto-util`, `concerto-analysis`, `concertino`, and `concerto-linter` to emit both CommonJS and ESM.

## Backwards Compatibility

All proposed changes are backwards compatible:
- Existing `dist/index.js` (CJS) files are unchanged
- ESM output is additive
- `"sideEffects"` metadata is ignored by tools that don't support it
- `"exports"` field is safe because no deep subpath imports exist in the codebase

## Affected Packages

- `@accordproject/concerto-util`
- `@accordproject/concerto-core`
- `@accordproject/concerto-cto`
- `@accordproject/concerto-analysis`
- `@accordproject/concerto-vocabulary`
- `@accordproject/concertino`
- `@accordproject/concerto-linter`

Rishabh060105:
# Closes #1261

Adds tree-shakeable package metadata, additive ESM entrypoints, and browser-bundle validation while preserving existing CommonJS and UMD browser bundle compatibility.

### Changes
- Add `module`, conditional `exports`, and `sideEffects` metadata for the issue-listed packages.
- Add an explicit esbuild-based ESM build step that emits `dist/esm/index.mjs` without changing package-wide `type` semantics.
- Consolidate Concerto Core `dayjs.extend()` calls into a shared `dayjs-setup` module.
- Add Playwright browser E2E coverage for the generated UMD bundles and wire it into Ubuntu CI.
- The additive ESM build preserves existing `main`, `typings`, and browser UMD entrypoints.

### Flags
- None

### Screenshots or Video
N/A

### Related Issues
- Issue #1261

### Author Checklist
- [x] Merging to `main` from `Rishabh060105/issue-1261-tree-shaking`


Rishabh060105:
Hi @mttrbrts,

I avoided changing package-wide `"type"` semantics and did not rely on plain `tsc` ESM output. Instead, the packages keep their existing CJS `dist/index.js`, and an explicit ESM build step now emits `dist/esm/index.mjs`.

I also kept the existing browser UMD bundles intact, added `exports` / `module` / `sideEffects` metadata, consolidated the `dayjs.extend()` setup, and added Playwright browser coverage for the generated bundles.

One thing I wanted to call out: I kept a `./dist/*` export as a compatibility escape hatch so existing deep imports don’t break when `exports` is added. I found a few current consumers(template-engine using `dist/*` paths), so removing it now would likely be breaking. In the long term, we may want to narrow or remove that once consumers have migrated to the documented package root entrypoints.

mttrbrts:
## Verification findings

Tested this PR locally by building all packages and running a Rollup + `@rollup/plugin-node-resolve` bundle against the new `dist/esm/index.mjs` files. Summary of what works and what doesn't.

### ✅ What works

- All ESM builds generate correctly (`dist/esm/index.mjs` for every package)
- The `exports` conditional field correctly routes bundlers to the ESM file via the `"import"` condition
- **Package-level tree shaking works** — a web app that imports only from `concerto-util` produces a 59 KB bundle and correctly excludes `concerto-core` entirely

### ❌ Bug: `sideEffects` in `concerto-core` blocks tree shaking

`packages/concerto-core/package.json` lists `./dist/esm/index.mjs` in `sideEffects`:

```json
"sideEffects": ["./dist/dayjs-setup.js", "./dist/esm/index.mjs"]
```

Marking the ESM entrypoint itself as side-effectful tells bundlers "always include this entire file" — which is the opposite of tree-shaking. The real side effect is the `dayjs.extend()` calls in `dayjs-setup.ts`, not the whole index.

I've opened a fix in draft PR #1263 that:
1. Removes `./dist/esm/index.mjs` from `sideEffects`
2. Adds a second esbuild pass to emit `dist/esm/dayjs-setup.js` as a separate file so the `sideEffects` field can reference it correctly

### ⚠️ Per-export granular tree shaking does not work (known limitation)

Importing 1 export vs all 40 from `concerto-core` produces effectively identical bundle sizes (35-byte difference). The root cause is structural: the ESM build is a pre-bundled CJS-in-ESM blob using `__commonJS`/`__esm` wrappers, which downstream bundlers cannot statically analyse.

```
import { ModelManager }                 → 1,462,043 bytes
import { ModelManager, Serializer, … }  → 1,462,008 bytes
```

This is because the source uses `import X = require()` / `export = X` TypeScript CJS patterns throughout. Fixing it requires migrating source files to native ESM syntax — a larger follow-on PR. The path is straightforward (the `exports` dual-condition field already provides the right infrastructure), but it touches every source file.

### Build order note

Running `npm run build` in `packages/concerto-core` alone fails until `concerto-cto`'s ESM dist exists first, because the new `exports.import` field now resolves to `dist/esm/index.mjs` in the workspace symlink before it's been built. CI likely handles this via dependency order, but it's worth noting for local development docs.

Rishabh060105:
Hi @mttrbrts, I think it is pointing to the `main` instead of  `Rishabh060105/issue-1261-tree-shaking`.

PR #1276, opened 2026-06-30
fix(concerto-vocabulary): look up terms by own property only

### LINKED ISSUE #1241: VocabularyManager.getTerm returns Object.prototype built-ins for property names that shadow Object.prototype (hasOwnProperty, constructor, toString, …)
`VocabularyManager.getTerm` returns a built-in `Object.prototype` method (a `Function`) instead of the YAML vocabulary string when a property name matches a key on `Object.prototype` — such as `hasOwnProperty`, `constructor`, or `toString`.

## Root Cause

In `vocabulary.js` line 143:

```js
const property = decl.properties.find(d => d[propertyName]);
```

Each element `d` in `decl.properties` is a plain `{}` object. For any `propertyName` that exists on `Object.prototype`, `d[propertyName]` walks the prototype chain and returns the built-in method — which is truthy. As a result:

1. `find` always returns the **first** element in the array (all plain objects inherit these)
2. `property[propertyName]` on line 144 then returns the built-in `Object.prototype` method
3. The caller receives a `Function`, not the vocabulary string

> **Note:** the fix proposed in #1184 (`propertyName in d`) does **not** fix this — `'hasOwnProperty' in {}` also returns `true` via the prototype chain. The correct fix is `Object.prototype.hasOwnProperty.call(d, propertyName)`.

## Steps to Reproduce

```js
const { VocabularyManager } = require('@accordproject/concerto-vocabulary');
const vm = new VocabularyManager();

const vocab = `
locale: en
namespace: org.test@1.0.0
declarations:
  - Book: A book
    properties:
      - hasOwnProperty: Custom term
      - constructor: Custom constructor term
      - toString: Custom toString term
`;

vm.addVocabulary(vocab);

console.log(vm.getTerm('org.test@1.0.0', 'en', 'Book', 'hasOwnProperty'));
// Expected: "Custom term"
// Actual:   [Function: hasOwnProperty]  ← Object.prototype.hasOwnProperty

console.log(vm.getTerm('org.test@1.0.0', 'en', 'Book', 'constructor'));
// Expected: "Custom constructor term"
// Actual:   [Function: Object]          ← Object.prototype.constructor

console.log(vm.getTerm('org.test@1.0.0', 'en', 'Book', 'toString'));
// Expected: "Custom toString term"
// Actual:   [Function: toString]        ← Object.prototype.toString
```

## Affected Property Names (non-exhaustive)

Any name present on `Object.prototype`:
`hasOwnProperty`, `constructor`, `toString`, `valueOf`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString`

## Possible Solution

Replace the plain property access on line 143 of `vocabulary.js` with an own-property check:

```js
// Before — traverses the prototype chain; shadowed by Object.prototype built-ins
const property = decl.properties.find(d => d[propertyName]);

// After — checks only own properties; immune to prototype shadowing
//         and also fixes falsy-value coercion (see #1183) as a side-effect
const property = decl.properties.find(d => Object.prototype.hasOwnProperty.call(d, propertyName));
```

This single-line change also subsumes the falsy-value fix from #1183 because `Object.prototype.hasOwnProperty.call` tests key existence, not value truthiness — so `""`, `false`, and `0` are handled correctly too.

## Relationship to #1183

#1183 covers falsy-value coercion (`""`, `false`, `0` → `null`). The proposed fix there (`propertyName in d`) would partially address falsy values but would **not** resolve this prototype pollution bug. Using `Object.prototype.hasOwnProperty.call(d, propertyName)` fixes **both** issues with a single change.

## Environment

- Package: `@accordproject/concerto-vocabulary`
- Runtime: Node.js (any version)
- Concerto version: current `main`

apoorva-01:
# Closes #1241

`Vocabulary.getTerm` and `getElementTerms` matched a property's term with `propertyName in d`, which walks the prototype chain. A property named after an `Object.prototype` member (`hasOwnProperty`, `constructor`, `toString`, `valueOf`) matched every declaration object and returned a built-in `Function` instead of the term, or a `Function` instead of `null` for an undeclared name.

### Changes

- Both methods now use an own-property check, `Object.prototype.hasOwnProperty.call(d, propertyName)`, so a term comes back only for a property actually declared on the object.
- Tests in `test/vocabulary.js` cover `constructor`/`toString`/`hasOwnProperty` resolving to their own terms and undeclared prototype names returning `null`; they fail on revert. Suite (95), lint and coverage green.

### Flags

- Used `hasOwnProperty.call` rather than `Object.hasOwn`, which reads cleaner but isn't in the package's `es2020` TS lib.
- Fixed `getElementTerms` alongside `getTerm` since it has the identical defect, happy to narrow to `getTerm`.
- Also covers the property-existence half of #1183.

### Related Issues

- Issue #1241
- Issue #1183 (partly addressed)

### Author Checklist

### DIFF
--- a/packages/concerto-vocabulary/src/vocabulary.ts
+++ b/packages/concerto-vocabulary/src/vocabulary.ts
@@ -131,7 +131,7 @@ class Vocabulary {
             return identifier ? decl[identifier] : decl[declarationName];
         }
         else {
-            const property = decl.properties ? decl.properties.find((d: any) => propertyName in d) : null;
+            const property = decl.properties ? decl.properties.find((d: any) => Object.prototype.hasOwnProperty.call(d, propertyName)) : null;
             return property ? identifier ? property[identifier] : property[propertyName] : null;
         }
     }
@@ -154,7 +154,7 @@ class Vocabulary {
             return decl;
         }
         else {
-            const property = decl.properties ? decl.properties.find((d: any) => propertyName in d) : null;
+            const property = decl.properties ? decl.properties.find((d: any) => Object.prototype.hasOwnProperty.call(d, propertyName)) : null;
             return property;
         }
     }
--- a/packages/concerto-vocabulary/test/vocabulary.js
+++ b/packages/concerto-vocabulary/test/vocabulary.js
@@ -191,4 +191,56 @@ describe('Vocabulary', () => {
             terms.label.should.equal('');
         });
     });
+
+    describe('getTerm / getElementTerms - property names that shadow Object.prototype', () => {
+        let voc;
+        beforeEach(() => {
+            voc = new Vocabulary({}, {
+                locale: 'en',
+                namespace: 'org.proto@1.0.0',
+                declarations: [
+                    {
+                        Book: 'A book',
+                        properties: [
+                            { title: 'The title' },
+                            { constructor: 'Constructor term' },
+                            { toString: 'ToString term' },
+                            { hasOwnProperty: 'HasOwnProperty term' }
+                        ]
+                    }
+                ]
+            });
+        });
+
+        it('getTerm - returns the term for a property named constructor', () => {
+            const term = voc.getTerm('Book', 'constructor');
+            term.should.equal('Constructor term');
+        });
+
+        it('getTerm - returns the term for a property named toString', () => {
+            const term = voc.getTerm('Book', 'toString');
+            term.should.equal('ToString term');
+        });
+
+        it('getTerm - returns the term for a property named hasOwnProperty', () => {
+            const term = voc.getTerm('Book', 'hasOwnProperty');
+            term.should.equal('HasOwnProperty term');
+        });
+
+        it('getTerm - returns null for an undeclared Object.prototype property name', () => {
+            const term = voc.getTerm('Book', 'valueOf');
+            should.equal(term, null);
+        });
+
+        it('getElementTerms - returns the property for a name that shadows Object.prototype', () => {
+            const terms = voc.getElementTerms('Book', 'toString');
+            should.exist(terms);
+            terms.toString.should.equal('ToString term');
+        });
+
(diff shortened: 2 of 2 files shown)


mttrbrts:
## pr-autopilot review

Small, well-scoped fix in `concerto-vocabulary`: term/property lookups now use `Object.prototype.hasOwnProperty.call(d, propertyName)` instead of the `in` operator, so a property object no longer matches names it merely inherits from `Object.prototype` (e.g. `toString`, `constructor`, `hasOwnProperty`). Both affected methods (`getTerm`, `getElementTerms`) were updated consistently, and the new tests exercise the shadowing cases plus a negative case for an undeclared prototype-inherited name. No API surface, docs, or dependency changes.

**Risk breakdown**

- Blast radius: 0.33 (confidence 0.85)
- Security: 0.13 (confidence 0.62)
- API compatibility: 0.71 (confidence 0.83)
- Test adequacy: 0.97 (confidence 0.90)
- Description quality: 1.00 (confidence 1.00)
- Composite risk: 29.4/100 (medium)

**Decision:** needs-review -- gates failed: not_first_time_contributor: author association 'FIRST_TIME_CONTRIBUTOR' is not trusted

---
_Automated review by pr-autopilot (kind: bugfix, policy: 2026.09.23-3+70e377528a92). This is not a substitute for human review._

mttrbrts:
**Summary**

Fixes a bug where `Vocabulary.getTerm`/`getElementTerms` used `propertyName in d`, which searches the prototype chain, causing properties named like `constructor`, `toString`, or `hasOwnProperty` to be resolved against `Object.prototype` instead of the declaration's own properties. The fix switches to `Object.prototype.hasOwnProperty.call(d, propertyName)` to restrict matching to own properties only.

The fix is small, correctly scoped, and comes with solid regression tests covering the previously-mishandled property names and a negative case for an undeclared property that also happens to shadow a prototype method.

No issues found. Approving.

---
*This is an automated review.*

PR #1287, opened 2026-07-28
fix: correct expected HTTP status in modelmanager test

### LINKED ISSUE #1286: Fix failing modelmanager test on bad URL
The test expects a 400 status for a bad URL on github, but the github request actually returns 404. This causes the test to fail. The assertion should be updated to expect 404.

aadityakumarsah:
Fixes #1286. The test expects 400 but the HTTP request returns 404. Updated the assertion to match the actual result.


junaidjmomin:
Thanks for working on this. I reproduced the failure locally: `github://external.cto` is rewritten to `https://raw.githubusercontent.com/external.cto`, and the test currently makes a live request through `global.fetch`. GitHub now returns 404 instead of the previously expected 400.

Changing the assertion to 404 fixes today's failure, but the test will remain dependent on GitHub's response for a malformed URL. Would it be better to stub `global.fetch` with the existing Sinon sandbox, return a controlled `{ ok: false, status: 404 }` response, and keep the full error assertion? That would preserve coverage of the default loader path while making the test deterministic and offline-safe.

I also noticed the PR currently includes unrelated dependency-pinning changes across multiple package files and `package-lock.json`; those appear to come from the earlier commit on the branch. The DCO and Snyk checks are failing as well. Rebasing or recreating the branch from current `main` with only the focused test change should make the PR much easier to review.

mttrbrts:
## pr-autopilot review

This PR updates a single test assertion in `modelmanager.js`, changing the expected HTTP error status from 400 to 404 for the "bad URL" scenario in `updateExternalModels()`.\n\nThe change itself is correct and minimal, and the commit message follows the required `type(scope): description` convention. The one thing worth calling out: the test this fixes performs a real, unmocked network request to `raw.githubusercontent.com`, so its outcome depends on GitHub's current behavior rather than on Concerto's own code. That's presumably why this test broke in the first place, and hardcoding the new observed status code doesn't remove that risk — a future change on GitHub's side could break it again. Mocking the HTTP call would make this test deterministic.\n\nThe diff also carries a large volume of dependency-pinning changes (lockfile and package.json updates) that don't appear related to the stated purpose of this PR; if those were already merged separately, it would be worth double-checking why they show up here.

**Risk breakdown**

- Blast radius: 0.20 (confidence 0.85)
- Security: 0.11 (confidence 0.66)
- API compatibility: 0.00 (confidence 0.99)
- Test adequacy: 0.51 (confidence 0.50)
- Description quality: 0.60 (confidence 0.81)
- Composite risk: 15.5/100 (high)

**Decision:** blocked -- not merge-ready: ci_green: CI state is 'FAILURE' (facts.ci_green=False); mergeable: mergeable is 'CONFLICTING' (facts.is_mergeable=False)

---
_Automated review by pr-autopilot (kind: test, policy: 2026.09.23-3+70e377528a92). This is not a substitute for human review._

PR #1293, opened 2026-08-20
fix(concerto-cto): escape string defaults, decorator strings, and regex patterns

rajanpanth:
`Printer.toCTO()` writes string-valued defaults straight into a double-quoted CTO literal with no escaping:

```ts
defaultString += ` default="${mm.defaultValue}"`;
```

So a default containing a double quote produces CTO that the project's own parser cannot read:

```js
const { Printer, Parser } = require('@accordproject/concerto-cto');

const cto = Printer.toCTO(modelWithDefault('he said "hi"'));
// namespace org.acme@1.0.0
// concept C {
//   o String s default="he said "hi""
// }

Parser.parse(cto);
// ParseException: Expected "-->", "@", "length", "o", "optional", "regex", "}",
// comment, end of line, or whitespace but "h" found. Line 4 column 32
```

Print → parse should round-trip, and here it doesn't. A backslash in a default has the same problem: it is emitted raw and then consumed as an escape introducer on the way back in.

## Fix

Both string-valued sites — `StringProperty`/`StringScalar` and `DateTimeProperty`/`DateTimeScalar` — now go through a small `toStringLiteral` helper that emits the quoted literal via `JSON.stringify`.

The CTO grammar's `DoubleStringCharacter` accepts `\\` followed by `EscapeSequence`, which covers `CharacterEscapeSequence`, `HexEscapeSequence` and `UnicodeEscapeSequence`. Everything `JSON.stringify` produces falls inside that set, so the escaped output parses back to the original value.

## Verification

- `packages/concerto-cto`: **154 passing**, including every existing `Should print <name>.json` golden-file comparison. `JSON.stringify` is byte-identical to the old output for defaults with no special characters, so no fixture changed.
- Two tests added: one asserting the escaping, one asserting `Parser.parse(Printer.toCTO(...))` recovers `he said "hi"` unchanged.
- To confirm the tests exercise the fix, I reverted `printer.ts` alone and re-ran: **21 passing, 2 failing**. Both new tests fail without the change.
- `eslint .` clean in the package.

I found this while round-tripping models through concerto-playground, which delegates its CTO output to this printer. I did not find an existing issue covering it; happy to open one first if you'd prefer that order.


rajanpanth:
Pushed a second commit extending this to the two sibling emission sites in the same file, found by probing the published package the same way:

1. **Decorator string arguments.** `decoratorArgFromMetaModel` emitted `"${value}"` unescaped, so `@Term("he said \"hi\"")` printed as `@Term("he said "hi"")`, which the parser rejects. Now uses the same `toStringLiteral` helper.

2. **Regex validators.** `regex=/${pattern}/` emitted the pattern verbatim, so a pattern containing `/` (for example `a/b`) or a line terminator ended the literal early and produced unparseable CTO. Patterns are now normalized the way JavaScript normalizes `RegExp.prototype.source` (which escapes exactly `/` and line terminators), with invalid patterns left untouched rather than throwing while printing.

Verification for the new commit: 25 printer tests pass (2 added), 156 across the package, eslint clean. Reverting only the printer change fails exactly the two new tests. All existing golden-file fixtures are unchanged.

Happy to split this into a separate PR if you would rather review the default-value fix alone.

mttrbrts (approved):

PR #1305, opened 2026-08-28
feat(concerto-cto): allow validators in any order in field declarations

muhabdulkadir:

Validators (regex, length, range, size) are no longer positionally fixed in the PEG grammar; they can appear in any order between default and optional

- Previously, `o String x length=[1,10] regex=/abc/` would fail to parse because `regex` had to come before `length`
- Adds per-type *Validators collection rules that match any valid validator for the field type in any order and bucket them by `$class`

### Author Checklist


mttrbrts (approved):
Nicely done, thanks Muh. Can you patch up the PR title, and check that the old issue is closed?

PR #1306, opened 2026-08-28
feat(concerto)!: native ESM migration for per-export tree-shaking (v5.0.0)

### LINKED ISSUE #1304: feat: native-ESM migration of concerto-core/concerto-util for per-export tree-shaking
## Summary

Follow-up to #1261 / PR #1263. Package-level tree-shaking now works (unused `@accordproject/*` packages are dropped from web bundles), but **per-export** tree-shaking inside `concerto-core` does not — importing one export vs. the whole library differs by ~0.14%.

This epic tracks migrating `concerto-core` (and `concerto-util`) source from TypeScript `export =` / `import X = require()` to **native ESM** so bundlers can drop unused exports.

## Root cause

The ESM build is `esbuild --bundle` (`bundle: true`), which produces a **CJS-in-ESM blob** (`__commonJS` / `__esm` wrappers). Bundlers cannot tree-shake inside that blob. Bundling is currently forced because **58 of 59 `concerto-core/src` files** use TS `export =` / `import X = require()`, which is illegal as native ESM (tsc `TS1203`).

## Spike measurement (GO — conditional)

A throwaway spike migrated one leaf module (`securityexception.ts`) to native ESM with `bundle: false`:

| Import | Bundle size |
|--------|-------------|
| 1 export — current (`bundle: true`) | 595,862 B (596 KB) |
| All exports — current | 596,715 B (597 KB) |
| **1 export — after native-ESM migration** | **22,908 B (22 KB)** |

**96.2% reduction (573 KB saved)** for a single leaf import. The win is real and the migration is mechanical.

**Condition:** full wins require `concerto-util` to co-migrate — it is *also* a CJS-in-ESM blob, so leaf imports still pull in ~all of `concerto-util` until it too is native ESM. With both migrated, a `SecurityException` import would be <2 KB.

## Migration shape

Bottom-up topological traversal of the import graph. Per-file pattern:

1. `const X = require('./foo')` → `import X from './foo.js'`
2. `const { X, Y } = require('@external')` → `import { X, Y } from '@external'`
3. `export = ClassName` → `export { ClassName }; export default ClassName;`
4. Migrate `globalize.ts` before any file that uses it.

Build-path change:
- Replace the `build:esm` esbuild `bundle: true` pass with an unbundled pass (`tsc --module ES2020 --moduleResolution bundler`, or esbuild `bundle: false` + `splitting: true`) emitting one `.mjs` per source file.
- Update `tsconfig.build.json`: `module` → `ES2020`, `moduleResolution` → `bundler`. Removing `export =` resolves `TS1203`.

File inventory (from spike):
- 4 pure leaf files (no internal `./` requires): `securityexception.ts`, `metamodelexception.ts`, `dayjs-setup.ts`, `serializer/validationexception.ts`
- ~38 files with only `const X = require('./internal')` patterns
- Complex hubs: `basemodelmanager.ts` (20 internal requires), `serializer/jsonpopulator.ts` (8+4), `factory.ts` (10)

## Recommended scope

1. Migrate `concerto-util` to native ESM (~1–2 days)
2. Migrate `concerto-core` to native ESM, top-down (~2–3 days)
3. Update `tsconfig.build.json` (module/moduleResolution) and replace the `build:esm` bundling step
4. Optional: per-module `exports` map entries for deep imports
5. Verify downstream packages (`concerto-analysis`, `concerto-vocabulary`, etc.) still build

**Estimate:** ~3–5 developer-days including test verification.

## Constraints / risk

- Must keep **all existing tests passing unchanged** — this is a larger refactor than #1263 and stresses that guarantee more, hence a separate PR.
- Non-breaking for existing consumers (the CJS `require` path and `./dist/*` deep-import escape hatch must keep working).

_Tracked internally as beads epic `concerto-334`; spike `concerto-08k`._

mttrbrts:
# Closes #1304

Converts the `@accordproject/concerto-*` packages to native ESM so bundlers can tree-shake per export, and reshapes the published artefacts to match. **This has to ship as 5.0.0** — it is a deliberate breaking change.

> **Release plan:** merge into `main`, then cut a **`v5.0.0`** release from it. Versions are not hand-edited here — `publish.yml` derives them from the release tag via `scripts/bump_version.js`, so edits in the branch would be overwritten. `scripts/tag.js` publishes prereleases under the `unstable` dist-tag, so a `v5.0.0-alpha` is available first if you want downstream consumers to verify against a real publish before `latest` moves.

## What changed

| | |
|---|---|
| **Source** | `export =` / `import X = require()` → native ESM across the packages |
| **Tests** | 219 deep source requires updated to named destructuring, across 100 files |
| **ESM emit** | one output module per source module instead of one flattened bundle — this is what actually delivers tree-shaking |
| **Targets** | separate node and browser ESM graphs; `process` bound in the browser graph |
| **Dependencies** | runtime deps external in the node build, matching what `tsc` already does for CJS |
| **UMD** | webpack browser bundles removed; browser e2e moved onto the ESM graph |
| **CI** | a Node ESM smoke test now runs as part of `npm run build` |
| **Types** (#1308) | 182 `: any` annotations → 29; real base-class types restored; `Decorated.ast` exported as `AstNode`; 3 latent bugs fixed |

194 files changed, +3,486 / −5,380 across 40 commits.

## Headline result

Minified bundle for a single-symbol import, measured with `esbuild --bundle --minify --format=esm`, importing **by package name** so resolution happens the way it does in a consumer's project:

| What you import | 4.2.0 | 5.0.0 | |
|---|---:|---:|---:|
| `{ Writer }` from `concerto-util` | 95,570 B | **1,050 B** | −98.9% |
| `{ SecurityException }` from `concerto-core` | 903,570 B | **4,064 B** | −99.6% |
| `{ ModelManager }` from `concerto-core` | 903,565 B | **419,416 B** | −53.6% |

4.2.0 ships no `exports` map, so a browser bundler takes its `browser` field — the UMD bundle — which is what browser consumers actually got. Bundling for Node (`--platform=node`, taking 4.2.0's `main` and 5.0.0's `import` condition) starts from a smaller baseline and lands in the same place: 28,647 → **863 B**, 559,234 → **3,831 B**, 559,229 → **426,781 B**.

`ModelManager` genuinely reaches most of the package, so it shrinks least — that is the floor, not a bug.

Downstream, `concerto-playground`'s production bundle drops 14.1%.

## For consumers

**Not breaking for the overwhelming majority.** Root-package imports are unchanged in both module systems. The breaks are deep imports into a package's compiled internals, and the removal of the UMD `<script>` bundles. [`MIGRATION.md`](https://github.com/accordproject/concerto/blob/esm/integration/MIGRATION.md) is the full guide.

## Types (#1308, merged in)

<details>
<summary><b>What #1308 covers</b></summary>

The ESM switch restored real base-class types — `require()` had been typing every base class as `any`, which switched property checking off across `concerto-core`. The ESM migration silenced the fallout by declaring each instance field `: any`, including fields the base class already declared. #1308 replaced them with real types while the `.d.ts` surface is breaking anyway. `: any` in the files it touched: **182 → 29**, and 17 of those 29 are pre-existing in `concerto-vocabulary`.

**Declared once, in the owning class.** `ClassDeclaration` no longer redeclares `modelFile`; `MapDeclaration` drops `modelFile`/`name`/`fqn` plus three overrides that were byte-identical to `Declaration`'s; `Property` and `ModelFile` drop fields that were declared but never assigned. Fields carry definite assignment assertions rather than initialisers — with `useDefineForClassFields` off, a subclass initialiser is emitted *after* the base constructor has run `process()`, and would clobber the value.

**Real types.** `Decorated.ast` is an exported `AstNode`; the introspect and model classes, `BaseModelManager`, `Factory`, `Serializer` and the serializer visitors carry concrete field and return types; the validators take their metamodel nodes (`ICollectionSizeValidator`, `IStringRegexValidator`/`IStringLengthValidator`, the three domain validators); `ModelFile.getDeclarations()` is generic, so `getAssetDeclarations()` and friends return their own type; `NullUtil.isNull` and `ModelUtil.isValidIdentifier` are type predicates. `concerto-util` now exports `WritableModelFile`/`WriteModelsOptions`, and `concerto-cto` types `parse()` as `IModel`.

**Three latent bugs it surfaced.** `Property` and `ModelFile` passed an always-`undefined` `this.modelFile` to `IllegalModelException`; `ScalarDeclaration` passed `this.ast.name.location` as that exception's *modelFile* argument, which was always `undefined` since `ast.name` is a string; and `ClassDeclaration.declarationKind()` answered `''` for any subclass missing an override, which `_resolveSuperType()` uses to decide whether an extends clause is legal. It throws now.

Runtime behaviour is unchanged apart from those three exception arguments. `Property.decorator` is gone (write-only, always `null`), and `Identifiable.$timestamp` is typed `Dayjs` — it always held one; only the JSDoc said `string`.

Two follow-ups left deliberately undone: `Decorator.getArguments()` could return `DecoratorLiteralUnion[]` straight from the AST instead of the projected shape it builds today (breaking change to a public method), and `MapDeclaration`'s constructor calls `this.process()` after `Declaration`'s constructor already did, so every map declaration builds a throwaway `MapKeyType`/`MapValueType` pair.
</details>

## Verification

- Full suite green: 1,679 mocha + 136 jest, zero failures.
- Browser e2e green against the browser ESM graph (3 tests, Chromium).
- A 10-check Node ESM smoke test runs as `postbuild`, so `npm run build` fails if the emitted `.mjs` is wrong. It caught two real regressions during this work: an extensionless `dayjs/plugin/utc` import that Node ESM rejects, and named imports that the CJS lexer could not see.
- Downstream `concerto-playground` and `template-engine` built against packed tarballs.


scripts/build-esm.js
```diff
@@ -0,0 +1,157 @@
...
+        }));
+    },
+};
+
+const commonBuildOptions = {
+    bundle: true,
+    format: 'esm',
+    platform: isNodeOnlyPackage ? 'node' : 'browser',
+    mainFields: isNodeOnlyPackage ? ['module', 'main'] : ['browser', 'module', 'main'],
+    target: 'es2020',
```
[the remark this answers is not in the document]
mttrbrts: Confirmed, and fixed in ae0c9df. This was a real runtime break, not a theoretical one — reproduced against the built output before the fix:

```
ModelWriter FAILED: TypeError - Cannot read properties of undefined (reading 'replace')
FileWriter FAILED: TypeError - fs.mkdirSync is not a function
```

The build now emits two ESM trees instead of one:

- `dist/esm` — node-targeted, builtins external, so `import` resolves to real `fs`/`path`.
- `dist/esm-browser` — keeps the stubbing, routed through a new `browser` export condition placed *before* `import` so a browser bundler matches it first.

Node-only packages (`concerto-linter`) get just the node build.

One knock-on worth flagging for reviewers: adding the `browser` condition means webpack now resolves sibling workspace packages to their ESM build when producing the UMD bundles. Webpack treats `.mjs` as fully specified, which breaks the extensionless request its `process` ProvidePlugin injects, so the three UMD configs got a `.mjs` rule turning that off. The sibling packages' own relative imports already carry extensions.

`concerto-core`'s `sideEffects` now also lists `./dist/esm-browser/dayjs-setup.mjs`, so the plugin registration survives shaking on both builds.

Verified after the fix: `ModelWriter`, `FileWriter` and `ModelLoader` all work through the node ESM entry; the browser build still carries no bare builtin imports; tree-shaking is unchanged on both (browser `SecurityException` 3,666 B against a 547,984 B whole-package ceiling, `Writer` 739 B); full monorepo suite and the UMD browser e2e tests pass.

---

mttrbrts:
## Downstream validation: `template-engine` and `concerto-playground`

Built `esm/integration` (383ee19), packed the four packages with `npm pack`, and installed the tarballs into both repos via `overrides` so every transitive consumer resolves them too. Baseline for every A/B below is `main` (8a4460d) built and packed the same way.

**Both repos pass on this branch as it stands.** One regression found while validating (Node ESM), one behavioural change worth a decision (dual-package hazard), both with proposed fixes in **[#1307](https://github.com/accordproject/concerto/pull/1307)** (based on `esm/integration`, so the diff is just the fix).

### Results

| | template-engine (`5ea28a7`) | concerto-playground (`785bf2c`) |
|---|---|---|
| Deep imports into `dist/`/`types/` | none in src; `webpack.config.js` aliases `concerto-core/dist/concerto-core.js`, still reachable via `./dist/*` | none into concerto packages |
| `tsc` build | ✅ clean | ✅ clean (`tsc --noEmit`) |
| Unit tests | ✅ 109/109, 13 suites, 41 snapshots | ✅ 328/328, 25 files |
| Bundle | ✅ webpack UMD, 0 errors | ✅ `vite build` |
| Browser e2e | — | ✅ 104/104 Playwright specs |

An `npm ls`-wide grep across both dependency trees (`cicero-core`, `concerto-codegen`, `markdown-*`, `concertino`) found **zero** deep subpath imports into any `@accordproject/concerto-*` package — everything already uses the root specifier with a named destructure, so the documented `export =` → `export default` break has no ecosystem exposure in these two trees.

Public API is unchanged. Comparing every root export against 4.2.0 — name, `typeof`, and own-property members of each class and namespace object — gives **no differences** in either direction, for the CJS build and the ESM build alike. Only addition is `Decorated`, as the description says.

**The playground's tree-shaking win is real.** Its `exports` map now routes Vite to the ESM graph rather than the UMD `browser` field:

| | 4.2.0 | `esm/integration` | |
|---|---:|---:|---|
| `dist/assets/index-*.js` | 5,626,640 B | **4,831,555 B** | −14.1% |
| `dist/` total | 20,906,439 B | **19,430,120 B** | −7.1% |

### 1. Node ESM consumers get the browser build, with `fs` stubbed 🔴

`scripts/build-esm.js` builds with `platform: 'browser'` and `stubNodeBuiltinsPlugin`, which maps every builtin to `module.exports = {}`. The exports map has only `import` and `require` conditions, so `import` **under Node** lands on that stubbed graph. Anything that reaches the file system breaks — while the same call works under `require()`:

```js
import { FileWriter } from '@accordproject/concerto-util';
new FileWriter(dir).openFile(path.join(dir, 'out.txt'));
// TypeError: fs.mkdirSync is not a function
```

```
FAIL  FileWriter openFile/write/closeFile      TypeError: fs.mkdirSync is not a function
FAIL  ModelWriter.writeModelsToFileSystem      TypeError: fs.writeFileSync is not a function
FAIL  ModelLoader.loadModelManager             TypeError: fs.readFileSync is not a function
PASS  ModelManager + Factory + Serializer roundtrip
```

Those are the only three modules in the four packages that import a builtin (`concerto-util/src/filewriter.ts`, `concerto-util/src/modelwriter.ts`, `concerto-core/src/modelloader.ts`), so the blast radius is small but it covers `FileWriter` and `ModelLoader` — the two APIs `concerto-cli` and the codegen path are built on. Neither of these two repos hits it today (template-engine is CJS; the playground's browser stubbing is correct), but the playground's own vitest suite already runs `environment: "node"` against the stubbed graph, so it is one fs-touching test away.

Neither repo caught this and neither would: the unit suites all run against `src/` through ts-node, so nothing here exercises the emitted `.mjs` at all.

**Fix in #1307:** emit the same entry points twice — `dist/esm` unchanged (browser-safe), `dist/esm-node` with real builtins — and put a `node` condition ahead of `import`:

```json
"node": { "import": "./dist/esm-node/index.mjs", "require": "./dist/index.js" },
"import": "./dist/esm/index.mjs",
"require": "./dist/index.js"
```

Browser bundlers don't set `node`, so they keep resolving `dist/esm`. Re-validated after the fix: all three probes pass; playground bundle **4,831,725 B** (+170 B of chunk-hash noise, so the tree-shaking win is intact), 328 unit + 104 e2e green; template-engine build + 109 tests + webpack green; monorepo suite unchanged at 1276 / 163 / 151 / 89 / 99 / 18 / 19.

Also adds `scripts/smoke-esm.mjs` (`npm run test:esm`, wired into `build.yml`) — it imports each package as a consumer does and asserts the fs-backed APIs actually touch disk. It fails on this branch and passes with the fix.

*(One dead end worth recording: routing browsers with a `browser` condition instead breaks `concerto-core`'s own UMD webpack build — it starts resolving `concerto-util` through that condition and hits `Can't resolve 'process/browser'`. The `node`-condition direction leaves the browser path untouched.)*

### 2. Dual-package hazard: `import` and `require` in one process 🟠

4.x had no `exports` map and no `module` field, so `import` and `require` both resolved to the same CJS file — one copy of every class per process. That is no longer true, and concerto's `instanceof` guards notice. Same script, both versions:

```js
const esmResource = new Factory(mm).newResource('t@1.0.0', 'A', 'a1');   // via import
cjsSerializer.toJSON(esmResource);                                       // via require
```
```
4.2.0            identical class? true    → {"$class":"t@1.0.0.A","id":"a1"}
esm/integration  identical class? false   → Error: "Serializer.toJSON" only accepts
                                             "Concept", "Event", "Asset", "Participant"
                                             or "Transaction".
```

`instanceof Resource` / `Identifiable` / `Relationship` appear in the serialization and validation paths, so this is not theoretical. It is inherent to shipping both module systems, so #1307 documents it in MIGRATION.md rather than trying to fix it in the build — but it is a real 5.0.0 behaviour change that isn't in the description yet, and it deserves a maintainer decision.

Worth noting for the ecosystem: `template-engine/src/index.ts` re-exports `ModelManager` from `concerto-core` with the comment *"so consumers share one concerto instance"*. That path stays safe (template-engine is CJS, so an ESM consumer importing from it gets the CJS copy through cjs-module-lexer). Consumers who import `@accordproject/concerto-core` directly *alongside* it are the ones who end up with two.

### 3. Minor / no action needed

- **`types/` is now unreachable.** Adding an `exports` map where there was none also removes subpaths that used to resolve — here that's `types/lib/{dcsconverter,dcsIndexWrapper,decoratormodel}.d.ts`, three files that still ship in the tarball. Deliberate per the description, and neither repo uses them; just noting the tarball still carries dead weight.
- **`browser` field is now bypassed for the root import.** `exports` takes precedence, so browser bundlers get the ESM graph rather than the UMD `dist/concerto-core.js`. That is where the playground's 14% comes from, and it verified clean through `vite build` + 104 e2e — but it is a resolution change for every browser consumer, not just a size win, and isn't called out in the description.
- **Type resolution is healthy.** `moduleResolution` `node16` / `bundler` / `node10` all resolve the four packages and their types correctly.
- **`dist/index.d.ts` has no top-level `types` field** to fall back on, same as 4.x — TS finds it adjacent to `main`. Fine, just fragile if `main` ever moves.

<details>
<summary>Reproduction</summary>

```bash
# concerto @ esm/integration
npm ci && npm run build
for p in concerto-core concerto-util concerto-cto concerto-vocabulary; do
  (cd packages/$p && npm pack --pack-destination /tmp/tarballs)
done

# in each downstream repo: point the four deps + an overrides block at the
# tarballs, npm install, then run that repo's own build and test scripts
```
Node 22.22.2, npm 10.9.7, Linux. Playground e2e run against `/opt/pw-browsers/chromium_headless_shell-1194`.
</details>

---

mttrbrts:
## Re-validated against `3cb7b87` — one new downstream break

Re-ran the full downstream validation after the force-push. `3cb7b87` fixes the Node-ESM stubbing I reported — confirmed, `FileWriter`, `ModelWriter.writeModelsToFileSystem` and `ModelLoader` all reach real `fs` through the `import` condition now. My own build-fix commit is dropped in favour of yours; [#1307](https://github.com/accordproject/concerto/pull/1307) is rebased onto `3cb7b87` and now leads with what the re-validation turned up.

### 🔴 New: `template-engine`'s webpack build fails on this branch

`npm run webpack` — part of its `build:dist` and `prepublishOnly`, so **publishing template-engine against 5.0.0 fails**:

```
ERROR in ./node_modules/@accordproject/concerto-util/dist/esm-browser/chunk-GG6UL5GV.mjs 12:28-35
Module not found: Error: Can't resolve 'process/browser'
BREAKING CHANGE: The request 'process/browser' failed to resolve only because it was
resolved as fully specified (probably because the origin is strict EcmaScript Module,
e.g. a module with javascript mimetype, a '*.mjs' file, ...)
```

This is the *same* failure the new `.mjs` `fullySpecified` rules fix inside this repo — it just escapes into consumers now that the `browser` condition ships the ESM graph to them. `dist/esm-browser` leaves `process` free (`process.env.TZ` in the serializer, `process.emitWarning` in `warning.ts`, `process.env.NODE_ENV` in bundled deps), template-engine's `ProvidePlugin` answers it with `process: 'process/browser'`, and webpack rejects that extensionless request from a `.mjs` origin. Its config aliases `concerto-core` to the UMD bundle but not `concerto-util`, so `concerto-util` goes through the new condition.

The three webpack configs you patched can't reach a consumer's config, so this needs fixing on the publishing side. #1307 injects a `process` shim into the browser esbuild build instead — the graph becomes self-contained, no consumer needs a polyfill or a resolver rule, and it defers to `globalThis.process` so an existing consumer polyfill still wins. With it, template-engine's webpack is back to 0 errors and everything else stays green. It also makes your three `.mjs` rules no longer load-bearing, if you'd rather drop them.

Worth noting the general shape: `Buffer` and `global` are left free in `dist/esm-browser` too. Neither breaks a build today — `global` is behind a `typeof` guard, and bare `buffer` resolves where `process/browser` doesn't — but `Buffer` (reached via `rfdc`'s `instanceof Buffer`) is a `ReferenceError` waiting for any browser consumer who doesn't polyfill it, where the UMD build always did. Not fixed in #1307; flagging it as the same class of thing.

### Everything else re-confirmed on `3cb7b87`

| | template-engine | concerto-playground |
|---|---|---|
| tsc build | ✅ | ✅ |
| Unit tests | ✅ 109/109 | ✅ 328/328 |
| Bundle | ❌ webpack, above → ✅ with #1307 | ✅ `vite build` |
| Browser e2e | — | ✅ 104/104 |

The playground is unaffected because `vite-plugin-node-polyfills` sets `globalThis.process`. Bundle size holds at **4,834,581 B** vs 5,626,640 B on 4.2.0 (−14.1%), so the tree-shaking win survives the browser-condition change. Public API still matches 4.2.0 member-for-member across CJS and both ESM graphs. This repo's own suite, the 3 UMD browser e2e tests and `licchk` are unchanged.

The **dual-package hazard** from my earlier comment still stands unchanged (`import` and `require` in one process → two class identities → `instanceof` guards reject valid objects). It's inherent to dual publishing; #1307 documents it in MIGRATION.md rather than trying to fix it, but it's still a 5.0.0 behaviour change that isn't in the description yet.

`scripts/smoke-esm.mjs` in #1307 now guards both regressions — verified to fail without each fix and pass with it — and runs in CI as `npm run test:esm`. Worth having: every unit suite runs against `src/` through ts-node, so nothing here has ever exercised the emitted ESM.

---

mttrbrts:
## #1307 landed here; one of its flags doesn't reproduce

Thanks — the downstream validation caught something this repo's own CI structurally cannot see, and the `template-engine` webpack break is exactly the kind of thing that would have shipped.

Rather than wait on #1307 to merge, I've fast-forwarded `esm/integration` onto its three commits (`27c05dc`, `36f50be`, `b1df6ff`) unchanged, so authorship is preserved and #1307's diff is now empty — it can be closed. Head is `b1df6ff`.

Re-validated here after taking them: `npm run build` clean, `npm run test:esm` 10/10, full monorepo suite unchanged (1276 / 163 / 151 / 89 / 99 / 18 / 19), UMD browser e2e 3/3, `licchk` clean. The `process` binding is confirmed — `dist/esm-browser` no longer carries a free `process` identifier in any package.

### On the `Buffer` flag: it isn't reachable

You flagged `Buffer` as "a `ReferenceError` waiting for any browser consumer who doesn't polyfill it". I went looking to fix it and couldn't make it fire. The reference lives in `rfdc`'s `copyBuffer`, and its only call sites are guarded:

```js
} else if (ArrayBuffer.isView(cur)) {
    a2[k] = copyBuffer(cur);       // ← the only way into the Buffer reference
}
```

`copyBuffer` is unreachable unless the value being cloned is a typed array. Everything concerto deep-clones through `rfdc` is model AST and decorator-command data — strings, numbers, booleans, arrays, plain objects — so the guard is never true and the `Buffer` identifier is never evaluated.

Checked empirically too: `DecoratorManager.decorateModels` against `dist/esm-browser` with `globalThis.Buffer` deleted completes normally.

So I've deliberately **not** shimmed it. A `Buffer` stub would be dead code that could only mask a genuine polyfill, and the UMD build's `ProvidePlugin({ Buffer: ['buffer', 'Buffer'] })` was likewise never load-bearing for this path. Worth revisiting only if a future change puts a typed array into something concerto clones. `global` is behind a `typeof` guard, as you noted, so the same reasoning applies.

### The two items for a maintainer

Both of your flags about the description are fair; I've added them:

- **Dual-package hazard.** Now called out in the description, not just MIGRATION.md. It's a real 5.0.0 behaviour change: `import` and `require` in one process yield two class identities, and concerto's `instanceof` guards reject the crossover. Inherent to dual publishing, documented rather than fixed — but it should be visible to whoever approves the major.
- **`browser` field bypassed for the root import.** Also now in the description. It's a resolution change for every browser consumer, not only the size win — `exports` takes precedence, so browser bundlers get the ESM graph instead of the UMD `dist/concerto-*.js`.

### One note on the `.mjs` webpack rules

I've kept the three `fullySpecified: false` rules rather than dropping them. You're right that the shim makes them non-load-bearing for `process` specifically, but they're free and they keep this repo's own UMD builds from being the first thing to break if another free identifier ever appears in the browser graph. Happy to drop them if you'd rather have the smaller diff.

---

plan-issue-1304-esm-migration.md
```diff
@@ -0,0 +1,354 @@
+# Plan & Handoff — Issue #1304: Native-ESM migration of `concerto-core` / `concerto-util`
```
DianaLease: Is this file intentionally comitted?

DianaLease (commented):
The description says, "This is the v5.0.0 branch", but the PR is pointing towards the main branch. Not sure if that was referring to a github branch or not. Do we want to create a v5 branch or are we planning to merge this into main and then cut a v5 release?

plan-issue-1304-esm-migration.md
mttrbrts: Good catch — it shouldn't have been. Removed in f9601c9.

It was a cross-session handoff artifact: it described itself as a "resumable checkpoint" and existed to carry WIP state between working sessions while the migration was paused on the Option A/B/C decision. That decision is made and the work is finished, so it's process scaffolding rather than repository content. Nothing referenced it, and it had to carry an appended licence header purely to satisfy `license-check-and-add` — which was itself a hint it didn't belong at the repo root.

What's worth keeping has a better home already: `MIGRATION.md` carries the consumer-facing guidance, and the decision rationale, the tree-shaking measurements and the two hypotheses I tested and rejected are all in the PR description.

If you'd rather keep it as a design record, `contrib-notes/` is excluded from the licence check and would be the natural place — happy to restore it there instead.

---

mttrbrts:
@DianaLease — sorry, that wording was ambiguous. "The v5.0.0 branch" meant *the work that has to ship as 5.0.0*, not a git branch. I've reworded the description. The PR targets `main` deliberately; there is no v5 branch and nothing here assumes one.

That said, your question is the real one, and it's a maintainer call rather than something I should decide. Here is what the tooling actually constrains, so the choice is on facts:

**Versions come from the release tag, not from the branch.** `publish.yml` fires on `release: published`, then runs:

```
node ./scripts/bump_version.js $TAG        # rewrites cross-package dep ranges
npm version --workspaces --include-workspace-root --no-git-tag-version --exact $TAG
npm publish --workspaces --access public $DIST_TAG
```

So nothing in the repo needs to say "5.0.0" — the tag decides. That's also why I deliberately left `package.json` versions at 4.2.0 in this branch: editing them would be overwritten and would churn `package-lock.json`.

**`scripts/tag.js` picks the npm dist-tag from the semver:** a prerelease publishes as `unstable`, anything else as `latest`. So `v5.0.0-alpha.1` would go out under `unstable` without disturbing `latest` — useful if you want downstream repos to try it before it becomes the default install.

**The two options, concretely:**

1. **Merge to `main`, cut `v5.0.0` from `main`.** Simplest, and it matches how 4.2.0 shipped. The cost is that `main` becomes 5.x immediately, so a 4.x patch afterwards needs a branch cut from the last 4.x tag.
2. **Create a `v5` branch, merge there, release from it.** Only worth the overhead if you expect to keep shipping 4.x from `main` while v5 settles. Note `publish.yml` has no branch restriction — it publishes whatever commit the release points at — so this works, but it does mean two active lines to keep in sync.

My read is that option 1 fits unless there's known 4.x maintenance coming, and that a `v5.0.0-alpha` release under `unstable` is worth doing first either way — the downstream validation on `template-engine` and `concerto-playground` was done against packed tarballs, and a real `unstable` publish would let other consumers check themselves before `latest` moves. But that's your call, and I'm happy to follow whichever you pick.

---

mttrbrts:
> The description says, "This is the v5.0.0 branch", but the PR is pointing towards the main branch. Not sure if that was referring to a github branch or not. Do we want to create a v5 branch or are we planning to merge this into main and then cut a v5 release?

@DianaLease my plan was to merge to main, then release a 5.0 release

mttrbrts:
## Runtime dependencies are now external in the node ESM build (`dceada5`)

While measuring the two ESM targets I noticed the builds disagreed about what a dependency is. `tsc` emits `require("dayjs")` for the CJS output; the ESM build marked only workspace packages external, so every runtime dependency was compiled in. `debug`'s source was sitting inside concerto-util, concerto-core **and** concerto-cto independently, `dayjs` across nine chunks of concerto-core, and Spectral inside concerto-linter's 8.4 MB `index.mjs`.

### Published `dist/esm`, `.mjs` bytes

| package | before | after | |
|---|---:|---:|---:|
| concerto-util | 75,726 | **40,887** | −46% |
| concerto-core | 958,029 | **441,228** | −54% |
| concerto-vocabulary | 349,284 | **87,639** | −75% |
| concerto-cto | 505,039 | 491,738 | −3% |
| **total** | **1,888,078** | **1,061,492** | **−44%** |

### What it does *not* buy

Consumer bundle sizes barely move — between −0.3% and −0.8% across single-symbol, whole-package and multi-package imports. I'd expected a dedup win there and it didn't materialise, because the dependency the packages actually share is `@accordproject/concerto-metamodel`, and that one has to stay bundled (below). An intermediate build with *everything* external showed −13% on a multi-package consumer, so the win is real but currently blocked upstream.

So this is a published-artifact and dependency-hygiene change: smaller tarballs, dependencies visible to `npm audit` and SBOM tooling, and security patches flowing through a consumer's lockfile instead of requiring a concerto republish.

### Two things that had to stay bundled or change

**Four packages stay bundled.** `@accordproject/concerto-metamodel` and the three `@stoplight/spectral-*` packages are CommonJS with no ESM build, and Node's `cjs-module-lexer` can't see their named exports. Externalising them emits `import { MetaModelUtil } from '@accordproject/concerto-metamodel'` and Node throws *"Named export 'MetaModelUtil' not found"* at load. Bundling is what esbuild's interop was already doing for them. The exception list in `build-esm.js` carries the one-liner to re-test each entry — **if concerto-metamodel ever ships an `exports` map or an ESM build, dropping it from that list should unlock the consumer-bundle win too.**

**Two source imports needed file extensions.** `dayjs` and `@stoplight/spectral-cli` publish no `exports` map, so Node ESM rejects an extensionless subpath. Bundling had been resolving `dayjs/plugin/utc` by extension inference; once the specifier survives into the output it has to be `dayjs/plugin/utc.js`. Both forms work unchanged under CommonJS.

The browser graph is untouched — it still bundles its dependencies (its Node-builtin stubbing only holds if those dependencies are resolved at build time rather than by the consumer), and `dist/esm-browser` is byte-identical to before.

### Verification

`npm run test:esm` 10/10 — and worth noting it earned its keep here: it caught **both** of the failures above before they could ship, neither of which any unit test would have seen. Full monorepo suite unchanged (1276 / 163 / 151 / 89 / 99 / 18 / 19), UMD browser e2e 3/3, lint and `licchk` clean.

---

mttrbrts:
## Types (#1308, merged in)

#1308 is now part of this branch, so here is what it covers. Worth folding into the description; I couldn't edit it directly without risking the sections below the fold.

The ESM switch restored real base-class types — `require()` had been typing every base class as `any`, which switched property checking off across `concerto-core`. This migration silenced the fallout by declaring each instance field `: any`, including fields the base class already declared. #1308 replaced them with real types while the `.d.ts` surface is breaking anyway. `: any` in the files it touched: **182 → 29**, and 17 of those 29 are pre-existing in `concerto-vocabulary`.

**Declared once, in the owning class.** `ClassDeclaration` no longer redeclares `modelFile`; `MapDeclaration` drops `modelFile`/`name`/`fqn` plus three overrides that were byte-identical to `Declaration`'s; `Property` and `ModelFile` drop fields that were declared but never assigned. Fields carry definite assignment assertions rather than initialisers — with `useDefineForClassFields` off, a subclass initialiser is emitted *after* the base constructor has run `process()`, and would clobber the value.

**Real types.** `Decorated.ast` is an exported `AstNode`; the introspect and model classes, `BaseModelManager`, `Factory`, `Serializer` and the serializer visitors carry concrete field and return types; the validators take their metamodel nodes (`ICollectionSizeValidator`, `IStringRegexValidator`/`IStringLengthValidator`, the three domain validators); `ModelFile.getDeclarations()` is generic, so `getAssetDeclarations()` and friends return their own type; `NullUtil.isNull` and `ModelUtil.isValidIdentifier` are type predicates. `concerto-util` now exports `WritableModelFile`/`WriteModelsOptions`, and `concerto-cto` types `parse()` as `IModel`.

**Three latent bugs it surfaced.** `Property` and `ModelFile` passed an always-`undefined` `this.modelFile` to `IllegalModelException`; `ScalarDeclaration` passed `this.ast.name.location` as that exception's *modelFile* argument, which was always `undefined` since `ast.name` is a string; and `ClassDeclaration.declarationKind()` answered `''` for any subclass missing an override, which `_resolveSuperType()` uses to decide whether an extends clause is legal. It throws now.

Runtime behaviour is unchanged apart from those three exception arguments. `Property.decorator` is gone (write-only, always `null`), and `Identifiable.$timestamp` is typed `Dayjs` — it always held one; only the JSDoc said `string`.

Two notes on the description itself:

- The count under **What changed** is now stale. "189 files changed, +2,656 / −4,960" — with #1308 merged the PR reads **194 files, +3,486 / −5,380** across 40 commits.
- #1308 targeted `esm/integration`, and the workflows here are gated on `pull_request: branches: [main, v4.0.0]`, so none of its changes had CI until they landed on this branch. They were verified locally — full monorepo build including the ESM smoke checks, every workspace's tests, `eslint` and `lint:deps` — but this PR's run is the first time CodeQL and the conformance suite see them.

Two follow-ups left deliberately undone, both worth a look for 5.0.0: `Decorator.getArguments()` could return `DecoratorLiteralUnion[]` straight from the AST instead of the projected shape it builds today (breaking change to a public method), and `MapDeclaration`'s constructor calls `this.process()` after `Declaration`'s constructor already did, so every map declaration builds a throwaway `MapKeyType`/`MapValueType` pair.

---

jamieshorten:
Did an independent review of the diff. The downstream validation and the Node-ESM/process-shim/dual-package discussion already cover the big items, so this is just the type-safety things I didn't see raised elsewhere.

**1. `as unknown as` double casts in `basemodelmanager.ts`**

Two casts force incompatible types through the escape hatch:

```ts
downloader.downloadExternalDependencies(
    this.getModelFiles() as unknown as (ModelFileInstance[] & ModelFileSource[]), ...);

ModelWriter.writeModelsToFileSystem(
    this.getModelFiles() as unknown as WritableModelFile[], ...);
```

The first one is the worse offender — `ModelFileInstance[] & ModelFileSource[]` is an intersection that no single element can actually satisfy. It compiles only because `as unknown as` bypasses all checking. If either `FileDownloader`'s generic API or `ModelFile`'s shape drifts, this breaks at runtime with no compile-time signal. Worth at least a `// SAFETY:` comment explaining why the cast is sound, or better, a shared interface that both sides agree on.

**2. `getAllDeclarations() as ClassDeclaration[]` in `compare.ts` and `vocabulary.ts`**

```ts
this.compareClassDeclarations(comparers, a.getAllDeclarations() as ClassDeclaration[], ...);
```

`getAllDeclarations()` returns `Declaration[]`, which includes `ScalarDeclaration` and `MapDeclaration`. The cast tells TypeScript they're all `ClassDeclaration`. I verified this is runtime-safe today — the comparers only call `Declaration`-level methods and `getDeclarationType()` does its own `instanceof` dispatch — but it's a maintenance trap: anyone adding a `ClassDeclaration`-specific method call in a comparer gets no type error and a runtime crash on map/scalar inputs. The existing comment explains the intent, but a type-level fix (accepting `Declaration[]` in `compareClassDeclarations`) would be more durable.

**3. Minor: `minMax` and `duration` dayjs plugins are registered but unused**

`dayjs-setup.ts` registers four plugins. `utc` and `quarterOfYear` are exercised by the e2e test and by concerto-core itself. `minMax` and `duration` have zero call sites anywhere in the codebase (`grep` confirms). They're in `sideEffects` so they survive tree-shaking, but if they were dropped no test would notice. Not a bug — just worth knowing whether they're there for downstream consumers or are vestigial from 4.x.

packages/concerto-core/src/basemodelmanager.ts
```diff
@@ -464,11 +463,11 @@ class BaseModelManager {
         Object.assign(originalModelFiles, this.modelFiles);
 
         try {
-            const externalModels = await downloader.downloadExternalDependencies(this.getModelFiles(), options);
+            const externalModels = await downloader.downloadExternalDependencies(this.getModelFiles() as unknown as (ModelFileInstance[] & ModelFileSource[]), options);
```
jamieshorten: `ModelFileInstance[] & ModelFileSource[]` is an intersection no single element can satisfy — `as unknown as` just silences the disagreement between `FileDownloader`'s generic API and `ModelFile`'s shape. If either side drifts this breaks at runtime with no compile-time signal.

A shared interface (or widening `FileDownloader` to accept `ModelFile[]` directly) would make this cast unnecessary. At minimum a `// SAFETY:` comment explaining *why* the cast is sound would help future readers.

packages/concerto-core/src/basemodelmanager.ts
```diff
@@ -498,7 +497,7 @@ class BaseModelManager {
      *  If true, external models are written to the file system. Defaults to true
      */
     writeModelsToFileSystem(path, options = {}) {
-        ModelWriter.writeModelsToFileSystem(this.getModelFiles(), path, options);
+        ModelWriter.writeModelsToFileSystem(this.getModelFiles() as unknown as WritableModelFile[], path, options);
```
jamieshorten: Same pattern — `as unknown as WritableModelFile[]` bypasses type checking entirely. Less risky than the one above since `WritableModelFile` is a simpler interface, but still worth a safety comment or a type-level fix so `ModelWriter` can accept `ModelFile[]`.

packages/concerto-analysis/src/compare.ts
```diff
@@ -139,7 +139,9 @@ export class Compare {
 
     private compareModelFiles(comparers: Comparer[], a: ModelFile, b: ModelFile) {
         comparers.forEach(comparer => comparer.compareModelFiles?.(a, b));
-        this.compareClassDeclarations(comparers, a.getAllDeclarations(), b.getAllDeclarations());
+        // every declaration is put through the class declaration comparers; map and
+        // scalar declarations are additionally compared by the calls below
+        this.compareClassDeclarations(comparers, a.getAllDeclarations() as ClassDeclaration[], b.getAllDeclarations() as ClassDeclaration[]);
```
jamieshorten: `getAllDeclarations()` returns `Declaration[]` which includes `ScalarDeclaration` and `MapDeclaration` — neither is a `ClassDeclaration`. Verified this is runtime-safe today (comparers only call `Declaration`-level methods and `getDeclarationType()` does its own `instanceof` dispatch), but it's a maintenance trap: anyone adding a `ClassDeclaration`-specific call in a comparer gets no type error and a runtime crash on map/scalar inputs.

Widening `compareClassDeclarations` to accept `Declaration[]` would close the gap — the method name is misleading anyway since it already handles all declaration kinds.

packages/concerto-core/src/dayjs-setup.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc.js';
+import quarterOfYear from 'dayjs/plugin/quarterOfYear.js';
+import minMax from 'dayjs/plugin/minMax.js';
+import duration from 'dayjs/plugin/duration.js';
+
+dayjs.extend(utc);
+dayjs.extend(quarterOfYear);
+dayjs.extend(minMax);
```
jamieshorten, line 23: `minMax` and `duration` have zero call sites anywhere in the concerto codebase (grep-confirmed). `utc` and `quarterOfYear` are exercised by the e2e test and by concerto-core itself, but if these two were dropped no test would notice. Are they here for downstream consumers, or vestigial from 4.x?

jamieshorten (approved):

packages/concerto-core/src/basemodelmanager.ts
mttrbrts: Agreed a type-level fix is out of reach without changing FileDownloader's public generic API or WritableModelFile — both are cross-package surfaces I don't want to touch unilaterally right before this ships as 5.0.0. Added a `// SAFETY:` comment in a6a5688 explaining why the cast holds: getExternalImports() only ever touches `.ast`, which both ModelFileInstance and freshly-downloaded ModelFileSource satisfy structurally, and a caller-supplied fileDownloader has to honour the same contract.

---

packages/concerto-core/src/basemodelmanager.ts
mttrbrts: Went with the safety comment here too rather than widening ModelWriter to accept ModelFile[] — that would pull concerto-core's ModelFile into concerto-util, which sits lower in the dependency graph and is meant to stay free of that. Added in a6a5688: every ModelFile reachable here was built by a ModelFileLoader, which always supplies fileName/definitions, even though ModelFile's constructor types them nullable for the general case.

---

packages/concerto-analysis/src/compare.ts
mttrbrts: Traced this through before widening: `compareClassDeclarations`/`compareClassDeclaration` are private and the change you're describing is safe on its own, but the a/b values ultimately reach the public `Comparer.compareClassDeclaration?: (a: ClassDeclaration | undefined, ...) => void` callback — the one custom comparer authors implement against. Widening the private plumbing to `Declaration[]` still leaves a cast (or a public API change) right at that dispatch boundary, since ClassDeclaration is the documented, narrower type external implementers write against today.

Rather than change that public callback's signature unilaterally this close to the 5.0.0 cut, I added a `// SAFETY:` comment on the cast in a6a5688 spelling out the same reasoning you verified — instanceof guards inside compareClassDeclaration, and the fact the public callback already has to tolerate this at runtime. Fixed the same pattern in concerto-vocabulary/vocabulary.ts directly, though, since that one only calls `.getName()` and doesn't cross a public plugin interface.

Happy to revisit widening `Comparer.compareClassDeclaration` itself as a deliberate, separate API decision if that's wanted — just didn't want to fold it into this PR unasked.

---

packages/concerto-core/src/dayjs-setup.ts
mttrbrts: Checked git blame on `main`: `minMax` and `duration` were already registered pre-migration, unconditionally, in both `datetimeutil.ts` and `jsonpopulator.ts` (`dayjs.extend(minMax)` / `dayjs.extend(duration)`), with the exact same "no internal call site" property they have now — `dayjs-setup.ts` just consolidates four call sites that all did this independently in 4.x. So not vestigial or new, and not a regression from this PR. My read is they're intentional for downstream consumers: concerto hands back live `Dayjs` instances (e.g. from `DateTimeUtil`/`JSONPopulator`), and since dayjs plugins are global once registered, a consumer can call `dayjs.min()/max()` or `.duration()` on those without registering the plugins themselves, as long as concerto already has. Leaving as-is.

---

packages/concerto-core/src/basemodelmanager.ts
mttrbrts: Revisited — since 5.0.0 is a breaking release, the API change is the right call after all, so this is now fixed properly in 1a002d8 and the cast is gone.

`FileDownloader` takes a second type parameter: `FileDownloader<TFile, TSeed = TFile>`. The seed files are genuinely not the type the bound `FileLoader` produces — `BaseModelManager` seeds with parsed `ModelFile`s and gets back raw `ModelFileSource`s — which is exactly why one parameter could only be satisfied by that impossible intersection. `getExternalImports` now takes `TFile | TSeed`, which is what it is actually handed at runtime. `TSeed` defaults to `TFile`, so existing single-parameter uses are untouched.

---

packages/concerto-core/src/basemodelmanager.ts
mttrbrts: Also fixed properly in 1a002d8, without pulling `ModelFile` into `concerto-util`.

The nullability turned out to be real information rather than noise: a `ModelFile` built from an AST alone genuinely has no file name and no CTO source, so it cannot be written to disk. `WritableModelFile` stays strict, and the call site narrows and throws an error naming the namespace it can't write — where previously it cast and let `fs.writeFileSync` raise a `TypeError` a few frames later. Managers loaded from CTO are unaffected.

---

packages/concerto-analysis/src/compare.ts
mttrbrts: You were right, and 5.0.0 is the release that can absorb it — done properly in 1a002d8.

`Comparer.compareClassDeclaration` now takes `Declaration | undefined`, and the private plumbing along with it, so the cast is gone. The compiler now enforces what the runtime always required: narrow before reaching for `ClassDeclaration` members. None of the built-in comparers needed changing — they only call `getName()` and `getDeclarationType()`, and the latter already took a `Declaration`.

One extra simplification it enabled: `compareClassDeclaration()`'s two negative guards (`instanceof MapDeclaration`, `instanceof ScalarDeclaration`) collapse into a single positive `instanceof ClassDeclaration` check. Both extend `Declaration` directly, so the positive form covers them — and any declaration kind added later, instead of needing a new guard each time.

Left the `class-declaration-*` rule keys and the `compareClassDeclaration` name alone: they're user-configurable and appear in downstream changelog tooling, so the `// todo rename to declaration.ts` in comparers/class-declarations.ts is a separate, purely cosmetic change. Documented all of this in MIGRATION.md under breaking changes.

---

PR #1307, opened 2026-08-28
fix(build): bind process in the browser ESM build

### LINKED ISSUE #1304: feat: native-ESM migration of concerto-core/concerto-util for per-export tree-shaking

mttrbrts:
# Closes #1304 (partially — proposed fixes for #1306)

Proposed on top of [#1306](https://github.com/accordproject/concerto/pull/1306), from validating that branch against `accordproject/template-engine` and `accordproject/concerto-playground`. Rebased onto `3cb7b87`, so the diff is only what's still additive on top of it.

**Update after `3cb7b87`.** My original first commit fixed the Node-ESM stubbing; `3cb7b87` fixes the same bug independently and is now the base, so that commit is dropped. Re-validating the new base against both downstream repos surfaced a new break, which is what this PR now leads with.

### Changes

- **`fix(build)`: bind `process` in the browser ESM build.** `3cb7b87` routes browser bundlers to `dist/esm-browser` through a new `browser` condition. That graph leaves `process` free — the serializer reads `process.env.TZ`, `warning.ts` reads `process.emitWarning`, bundled deps read `process.env.NODE_ENV`. Inside this repo the three UMD configs answer that with `ProvidePlugin`, and the new `.mjs` `fullySpecified` rule keeps them building. Downstream, neither applies:

  ```
  ERROR in ./node_modules/@accordproject/concerto-util/dist/esm-browser/chunk-GG6UL5GV.mjs
  Module not found: Error: Can't resolve 'process/browser'
  BREAKING CHANGE: ... resolved as fully specified
  ```

  That's `accordproject/template-engine`'s `npm run webpack`, which is part of its `build:dist` and `prepublishOnly` — so publishing it against 5.0.0 fails. Its config aliases `concerto-core` to the UMD bundle but not `concerto-util`, so `concerto-util` resolves through the new condition and webpack's `ProvidePlugin` injects an extensionless `process/browser` request into a `.mjs` module.

  The fix injects a `process` shim into the browser build, so the graph is self-contained and no consumer needs a polyfill or a resolver rule. It defers to a real `globalThis.process` when the page has one, so an existing consumer polyfill still wins; the fallback covers only what these packages read. (Once this lands, the three webpack `.mjs` rules from `3cb7b87` are no longer load-bearing — left in place, since removing them is your call.)

- **`test(esm)`: `scripts/smoke-esm.mjs`, run in CI as `npm run test:esm`.** Every unit suite runs against `src/` through ts-node, so nothing exercises the emitted ESM — which is why both of these regressions were invisible on this side. It imports each package the way a consumer does and asserts the fs-backed APIs touch disk, that ESM and CJS expose the same names, and that the browser graph leaves no unbound `process`. Verified to fail on both bugs and pass on both fixes.

- **`docs(esm)`: MIGRATION.md.** The exports reference still showed the pre-`3cb7b87` single-graph shape; it now shows the `browser` condition and which graph a given runtime selects. Plus a new section on the dual-package hazard (below).

### Flags

- **The dual-package hazard is documented, not fixed.** In 4.x `import` and `require` resolved to the same CommonJS file — one copy per process. From 5.0.0 they resolve to different files, so a process reaching a package both ways holds two sets of class identities and concerto's `instanceof` guards reject objects that are obviously the right type:

  ```
  Error: "Serializer.toJSON" only accepts "Concept", "Event", "Asset", "Participant" or "Transaction".
  ```

  Verified A/B: identical script, 4.2.0 succeeds, this branch throws. Inherent to shipping both module systems rather than something the build can fix, so MIGRATION.md explains how to avoid it — but it's a real 5.0.0 behaviour change that isn't in #1306's description, and worth a maintainer decision.

- **`Buffer` and `global` are still free in `dist/esm-browser`**, and are not addressed here. Neither breaks a build today — `global` sits behind a `typeof` guard, and a bare `buffer` request resolves where `process/browser` doesn't — but `Buffer` (reached via `rfdc`'s `instanceof Buffer`) is a `ReferenceError` waiting for a browser consumer who doesn't polyfill it, where the UMD build always did. Same class of problem as the `process` fix; worth deciding whether the shim should cover them too.

- **`mjs` added to the root `license-check-and-add` format list**, so the new script's header is checked with the same comment style as `.js`.

### Verification

- template-engine: `npm run webpack` back to 0 errors, `npm run build` clean, 109/109 tests
- concerto-playground: `tsc --noEmit` clean, `vite build` clean, 328/328 unit, 104/104 Playwright e2e
- this repo: 1276 / 163 / 151 / 89 / 99 / 18 / 19, `npm run test:esm` 10/10, 3/3 UMD browser e2e, `licchk` clean
- Node ESM: `FileWriter`, `ModelWriter.writeModelsToFileSystem` and `ModelLoader` all reach real `fs`; ESM and CJS export names identical to 4.2.0, member-for-member

`.github/workflows/build.yml` only triggers on PRs targeting `main`, so the unit-test and e2e jobs don't run here by design — the results above are from running them locally on this head. DCO is green.

### Related Issues

- Pull Request accordproject/concerto#1306
- Issue accordproject/concerto#1304

### Author Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,6 +47,12 @@ jobs:
           NODE_OPTIONS: "--max_old_space_size=4096"
         run: npm test
 
+      # The unit suites run against src/ through ts-node, so this is the only
+      # job that exercises what a consumer actually resolves: the emitted ESM,
+      # and specifically the dist/esm graph Node takes.
+      - name: ESM smoke test
+        run: npm run test:esm
+
       - name: Archive npm failure logs
         uses: actions/upload-artifact@v4
         if: failure()
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -156,6 +156,47 @@ why the packages deliberately do **not** widen their `exports` map to re-expose 
 `types/` tree. If a type you need is not reachable from the root, please open an issue
 asking for it to be exported rather than deep-importing the declaration file.
 
+### Mixing `import` and `require` in one Node process
+
+In 4.x a package had no `exports` map and no `module` field, so `import` and `require`
+both resolved to the same CommonJS file: one copy of every class, in one process. From
+5.0.0 they resolve to different files, and Node loads each independently. A process
+that reaches a package both ways therefore holds **two copies** of it, with two sets of
+class identities:
+
+```js
+import { ModelManager, Factory } from '@accordproject/concerto-core';
+const concerto = createRequire(import.meta.url)('@accordproject/concerto-core');
+
+ModelManager === concerto.ModelManager;   // 4.x: true — 5.0.0: false
+```
+
+Most of the API is duck-typed and survives this, but the `instanceof` guards do not.
+The failure is not subtle when it happens:
+
+```js
+const resource = new Factory(modelManager).newResource('org.example@1.0.0', 'Order', 'o1');
+new concerto.Serializer(...).toJSON(resource);
+// Error: "Serializer.toJSON" only accepts "Concept", "Event", "Asset",
+// "Participant" or "Transaction".
+```
+
+This is the standard dual-package hazard, and it is inherent to shipping both module
+systems. It only bites a process that loads a package twice, so:
+
+- **Pick one module system per process for a given package.** In an ESM application,
+  `import` the package; don't also `createRequire` it.
+- **Watch for a CommonJS dependency in an ESM app.** A library that still uses
+  `require('@accordproject/concerto-core')` internally has its own copy. Objects can
+  cross that boundary safely, but `Resource`s and `ClassDeclaration`s created on one
+  side must be handed back to that same side to be serialized or validated. Where a
+  library re-exports concerto types for you (`@accordproject/template-engine` re-exports
+  `ModelManager` for exactly this reason), import them from the library rather than from
+  `@accordproject/concerto-core` directly.
+- **Check for a duplicate before debugging anything stranger.** If an `instanceof` or a
+  `Serializer` guard fails on an object that is obviously the right type, run
+  `npm ls @accordproject/concerto-core` and confirm you aren't loading it twice.
+
 ## Not breaking
 
 Root-package imports are unaffected — this is true for both CommonJS and ESM, and for
@@ -236,9 +277,9 @@ To benefit from the tree-shaking this release enables:
 - `@accordproject/concerto-util` declares `"sideEffects": false`, so a compliant
   bundler can drop any export you don't import.
 - `@accordproject/concerto-core` declares `"sideEffects": ["./dist/dayjs-setup.js",
-  "./dist/esm/dayjs-setup.mjs"]` — everything else is safe to drop, but those two
-  modules run global setup code (`dayjs` plugin registration) as a side effect and are
-  always kept.
+  "./dist/esm/dayjs-setup.mjs", "./dist/esm-browser/dayjs-setup.mjs"]` — everything
+  else is safe to drop, but those modules run global setup code (`dayjs` plugin
+  registration) as a side effect and are always kept.
 
 The ESM build ships one output module per source module (with shared code hoisted into
 chunks), mirroring the CJS build, so your bundler can drop whole modules you never
@@ -278,6 +319,7 @@ Every package's `exports` map has the same shape (shown here for `concerto-core`
 "exports": {
   ".": {
     "types": "./dist/index.d.ts",
+    "browser": "./dist/esm-browser/index.mjs",
     "import": "./dist/esm/index.mjs",
     "require": "./dist/index.js"
   },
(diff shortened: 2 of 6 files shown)


mttrbrts:
### CI status

**DCO is red, and it isn't this PR's.** It lists 14 unsigned commits, all of them inherited from `esm/integration` and its ancestors — 4 from the `esm/*` branches, 10 earlier. My two commits are signed off and are not among them. This is the same failure [#1306 already flags](https://github.com/accordproject/concerto/pull/1306) ("DCO is red, and was already red before this push"). Clearing it means rewriting the base branch's history, which is the author's call, not something to do from a PR stacked on top of it — so I'm standing down on it rather than force-pushing over someone else's commits.

**Nothing else ran.** `.github/workflows/build.yml` triggers only on PRs targeting `main` (or `v4.0.0`), and this one targets `esm/integration`, so the unit-test and e2e jobs are skipped by design. Run locally against the branch head instead, all green:

- `npm run build` clean, all seven packages emitting CJS + both ESM graphs + UMD + `.d.ts`
- `npm test` — 1276 / 163 / 151 / 89 / 99 / 18 / 19, matching #1306's stated baseline exactly
- `npm run test:esm` — 9/9 (fails on `esm/integration` without the build fix)
- `npm run licchk` clean
- template-engine: build + 109 tests + webpack UMD, green
- concerto-playground: `tsc`, `vite build`, 328 unit, 104 Playwright e2e, green

Both of these clear once the change lands on a branch that targets `main`.

---

PR #1308, opened 2026-08-29
refactor(types)!: tighten introspect and model types, drop redeclared fields

mttrbrts:

Stacked on `esm/integration` (#1306). That PR replaced `require()` with `import`, which restored real base-class types across `concerto-core` and so switched property checking back on. It papered over the resulting errors by declaring every instance field as `: any` — including fields the base class already declared. Since 5.0.0 is a deliberate breaking release, this replaces those with real types while the `.d.ts` surface is already changing.

`: any` in the files touched here: **182 → 29**, and 17 of the 29 are pre-existing in `concerto-vocabulary`, which is outside this PR's scope.

### Changes

**Declared once, in the owning class**
- `ClassDeclaration` no longer redeclares `modelFile` — `Declaration` owns it.
- `MapDeclaration` no longer redeclares `modelFile`, `name` or `fqn`, and drops its `getModelFile` / `getName` / `getFullyQualifiedName` overrides, which were byte-identical to `Declaration`'s. Its `process()` no longer recomputes the `name` / `fqn` that `super.process()` has already set.
- `Property` dropped `modelFile`, which was declared but never assigned.
- `ModelFile` dropped `modelFile` and `ast`, likewise never assigned and already owned by `Decorated`.

Fields carry definite assignment assertions (`foo!: T`) rather than initialisers throughout. These classes are populated by a `process()` method that a *base* constructor invokes, and with `useDefineForClassFields` off a subclass initialiser is emitted after that constructor has run — it would clobber the value. A type-only declaration erases to nothing and is safe.

**Real types**
- `Decorated.ast` is an exported `AstNode` interface; `decorators` is `Decorator[]`; `getModelFile()` declares `ModelFile`, which removes the casts its callers needed.
- `Declaration`, `ClassDeclaration`, `ScalarDeclaration`, `MapDeclaration`, `MapKeyType`, `MapValueType`, `Property`, `Field`, `RelationshipDeclaration`, `Decorator`, `ModelFile`, `Introspector` and the four `Validator`s carry concrete field and return types. Nullable fields are `T | null`.
- `ModelFile.getDeclarations()` is generic, so `getAssetDeclarations()` and friends return their own declaration type.
- `NullUtil.isNull` and `ModelUtil.isValidIdentifier` are type predicates, so they narrow at their call sites instead of forcing casts.
- `BaseModelManager`, `Factory`, `Serializer`, the serializer visitors, the model classes, `DecoratorManager` and `DecoratorExtractor` take concrete field types. New shared interfaces in `types.ts`: `SerializerOptions`, `GenerateOptions`, `InstanceGeneratorParameters`, and the decorator command shapes mirroring the DCS model.
- `concerto-util` exports `WritableModelFile` / `WriteModelsOptions`; `concerto-cto` types `parse()` as `IModel`, gives `resolveExternal` a named options type, and the `as any` casts are gone from `FileDownloader`.

**Bugs the tightening surfaced**
- `Property` and `ModelFile` passed an always-`undefined` `this.modelFile` to `IllegalModelException`. They now pass the real `ModelFile`, so those two errors gain the file name and location they were meant to carry.
- `ScalarDeclaration` passed `this.ast.name.location` as the *modelFile* argument to `IllegalModelException` — `this.ast.name` is a string, so that was always `undefined`. It now passes `this.modelFile` and `this.ast.location`.
- `ClassDeclaration.declarationKind()` gained a `return ''` stub during the ESM migration, silently answering with an empty string for any subclass that forgot to override it — and `_resolveSuperType()` compares declaration kinds to decide whether an extends clause is legal. It throws now, matching `Decorated.getModelFile()`.

### Flags

- **No CI runs on this PR, by design.** `build.yml`, CodeQL and the conformance tests are all gated on `pull_request: branches: [main, v4.0.0]`, and this targets `esm/integration` to keep the diff to just the type work. Verified locally instead — see the footnote. It needs rebasing onto `main` once #1306 merges, and CI covers it from then on.
- **Runtime behaviour is unchanged apart from the three exception arguments above.** Everything else is types-only: no reordering, no renames, no removed fields.
- **`AstNode` keeps an index signature.** The introspect classes are deliberately duck-typed across node kinds — a `ClassDeclaration` is built from either a concept or an enum declaration — so narrowing `ast` to a precise metamodel interface per subclass would mean a type guard at every access site. That is a worthwhile follow-up, not something to fold into a types-only change.
- **`Property.decorator` is vestigial** — only ever assigned `null`, never read anywhere in `src` or `test`. It is typed and documented rather than removed, since removing a public field is beyond this PR. Worth deleting in 5.0.0 if you agree.
- **Downstream fallout was small**: `concerto-analysis` and `concerto-vocabulary` needed null guards and narrowing casts where core methods now admit `null`, and one `concerto-analysis` test builds deliberately incomplete ASTs (its own comment says the suite exists to reach branches a well-formed model cannot) which now go through an explicit cast.

### Screenshots or Video

n/a

### Related Issues
- Pull Request #1306

### Author Checklist

<sub>Verified locally: `npm run build` green across all packages including the 10 ESM smoke checks; `npm test` green in every workspace (concerto-core 1278 passing, coverage back above its 99% threshold; concerto-util 151; concerto-cto 163; concerto-analysis 99; vocabulary/concertino/linter all passing); `eslint` and `lint:deps` clean.</sub>

### DIFF
--- a/packages/concerto-core/src/decoratorextractor.ts
+++ b/packages/concerto-core/src/decoratorextractor.ts
@@ -498,7 +498,7 @@ class DecoratorExtractor {
     */
     processModels(){
         const processedModels = this.sourceModelAst.models.map(model =>{
-            if (model?.decorators?.length) {
+            if ((model?.decorators?.length ?? 0) > 0) {
                 this.constructDCSDictionary(model.namespace, model.decorators, {});
                 model.decorators = this.filterOutDecorators(model.decorators);
             }
--- a/packages/concerto-core/src/decoratorextractor.ts
+++ b/packages/concerto-core/src/decoratorextractor.ts
@@ -17,6 +17,12 @@ import ModelUtil from './modelutil';
 import yaml from 'yaml';
 import { MetaModelNamespace } from '@accordproject/concerto-metamodel';
 
+// Types needed for TypeScript generation.
+/* eslint-disable no-unused-vars */
+import type { IModels } from '@accordproject/concerto-metamodel';
+import type { DecoratorCommandTarget } from './types';
+/* eslint-enable no-unused-vars */
+
 // $class value for decorator arguments that carry a string value
 const DECORATOR_STRING_TYPE = `${MetaModelNamespace}.DecoratorString`;
 
@@ -26,14 +32,25 @@ const DECORATOR_STRING_TYPE = `${MetaModelNamespace}.DecoratorString`;
  * @memberof module:concerto-core
  * @private
  */
+/**
+ * A decorator collected from a model, keyed in the extraction dictionary by
+ * the namespace it was found in.
+ */
+interface ExtractedDecorator {
+    declaration: string;
+    property: string;
+    mapElement: string;
+    dcs: string;
+}
+
 class DecoratorExtractor {
-    extractionDictionary: any;
-    removeDecoratorsFromModel: any;
-    locale: any;
-    dcs_version: any;
-    sourceModelAst: any;
-    updatedModelAst: any;
-    action: any;
+    extractionDictionary: Record<string, ExtractedDecorator[]>;
+    removeDecoratorsFromModel: boolean;
+    locale: string;
+    dcs_version: string;
+    sourceModelAst: IModels;
+    updatedModelAst: IModels;
+    action: number;
     /**
      * The action to be performed to extract all, only vocab or only non-vocab decorators
      */
@@ -53,7 +70,7 @@ class DecoratorExtractor {
      * @param {int} [action=DecoratorExtractor.Action.EXTRACT_ALL]  - the action to be performed
      * @param {object} [options] - decorator extractor options
      */
-    constructor(removeDecoratorsFromModel, locale, dcs_version, sourceModelAst, action = DecoratorExtractor.Action.EXTRACT_ALL, options?) {
+    constructor(removeDecoratorsFromModel: boolean, locale: string, dcs_version: string, sourceModelAst: IModels, action: number = DecoratorExtractor.Action.EXTRACT_ALL, options?: Record<string, unknown>) {
         this.extractionDictionary = {};
         this.removeDecoratorsFromModel = removeDecoratorsFromModel;
         this.locale = locale;
@@ -208,7 +225,7 @@ class DecoratorExtractor {
      * @private
      */
     constructTarget(namespace, obj){
-        const target: any = {
+        const target: DecoratorCommandTarget & { $class: string } = {
             '$class': `org.accordproject.decoratorcommands@${this.dcs_version}.CommandTarget`,
             'namespace':namespace
         };
@@ -481,7 +498,7 @@ class DecoratorExtractor {
     */
     processModels(){
         const processedModels = this.sourceModelAst.models.map(model =>{
-            if ((model?.decorators?.length > 0)){
+            if (model?.decorators?.length) {
                 this.constructDCSDictionary(model.namespace, model.decorators, {});
                 model.decorators = this.filterOutDecorators(model.decorators);
             }
--- a/packages/concerto-core/src/introspect/collectionsizevalidator.ts
+++ b/packages/concerto-core/src/introspect/collectionsizevalidator.ts
@@ -15,6 +15,12 @@
 import { NullUtil } from '@accordproject/concerto-util';
 import Validator from './validator';
 
+// Types needed for TypeScript generation.
+/* eslint-disable no-unused-vars */
+import type { ValidatedElement } from './validator';
+import type { AstNode } from './decorated';
+/* eslint-enable no-unused-vars */
+
 const { isNull } = NullUtil;
 
 /**
@@ -24,8 +30,8 @@ const { isNull } = NullUtil;
  * @memberof module:concerto-core
  */
 class CollectionSizeValidator extends Validator {
-    minSize: any;
-    maxSize: any;
+    minSize: number | null;
+    maxSize: number | null;
 
     /**
      * Create a CollectionSizeValidator.
@@ -34,14 +40,14 @@ class CollectionSizeValidator extends Validator {
      *
      * @throws {IllegalModelException}
      */
-    constructor(field, validator) {
+    constructor(field: ValidatedElement, validator: AstNode) {
         super(field, validator);
         this.minSize = validator.minSize ?? null;
         this.maxSize = validator.maxSize ?? null;
 
         if (isNull(this.minSize) && isNull(this.maxSize)) {
             this.reportError(field.getName(), 'Invalid collection size, minSize and/or maxSize must be specified.');
-        } else if (this.minSize < 0 || this.maxSize < 0) {
+        } else if ((this.minSize ?? 0) < 0 || (this.maxSize ?? 0) < 0) {
             this.reportError(field.getName(), 'minSize and/or maxSize must be positive integers.');
         } else if (isNull(this.minSize) || isNull(this.maxSize)) {
             // this is fine and means that we don't need to check whether minSize > maxSize
@@ -57,7 +63,7 @@ class CollectionSizeValidator extends Validator {
      * @throws {IllegalModelException}
      * @private
      */
-    validate(identifier, value) {
+    validate(identifier: string | null, value: number): void {
         if(!isNull(this.minSize) && value < this.minSize) {
             this.reportError(identifier, `Collection must contain at least ${this.minSize} elements.`);
         }
@@ -70,7 +76,7 @@ class CollectionSizeValidator extends Validator {
      * Returns the minSize for this validator, or null if not specified
      * @returns {number} the min size or null
      */
-    getMinSize() {
+    getMinSize(): number | null {
         return this.minSize;
     }
 
@@ -78,7 +84,7 @@ class CollectionSizeValidator extends Validator {
      * Returns the maxSize for this validator, or null if not specified
      * @returns {number} the max size or null
      */
-    getMaxSize() {
+    getMaxSize(): number | null {
         return this.maxSize;
     }
 
@@ -90,7 +96,7 @@ class CollectionSizeValidator extends Validator {
      * @returns {boolean} True if this validator is compatible with the other
      * validator, false otherwise.
      */
-    compatibleWith(other) {
+    compatibleWith(other: Validator | null): boolean {
         if (!(other instanceof CollectionSizeValidator)) {
             return false;
         }
--- a/packages/concerto-core/src/introspect/collectionsizevalidator.ts
+++ b/packages/concerto-core/src/introspect/collectionsizevalidator.ts
@@ -18,7 +18,7 @@ import Validator from './validator';
 // Types needed for TypeScript generation.
 /* eslint-disable no-unused-vars */
 import type { ValidatedElement } from './validator';
-import type { AstNode } from './decorated';
+import type { ICollectionSizeValidator } from '@accordproject/concerto-metamodel';
 /* eslint-enable no-unused-vars */
 
 const { isNull } = NullUtil;
@@ -30,6 +30,7 @@ const { isNull } = NullUtil;
  * @memberof module:concerto-core
  */
 class CollectionSizeValidator extends Validator {
+    declare validator: ICollectionSizeValidator;
     minSize: number | null;
     maxSize: number | null;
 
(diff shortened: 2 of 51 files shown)


packages/concerto-core/src/introspect/collectionsizevalidator.ts
```diff
@@ -34,14 +40,14 @@ class CollectionSizeValidator extends Validator {
      *
      * @throws {IllegalModelException}
      */
-    constructor(field, validator) {
+    constructor(field: ValidatedElement, validator: AstNode) {
```
mttrbrts: Can the type of validator be stronger?

packages/concerto-core/src/introspect/decorator.ts
```diff
@@ -19,27 +19,43 @@ import IllegalModelException from './illegalmodelexception';
...
+import type Decorated from './decorated';
+import type { AstNode } from './decorated';
+import type { IDecorator } from '@accordproject/concerto-metamodel';
 /* eslint-enable no-unused-vars */
 
+/**
+ * A decorator argument that references a type, produced from a
+ * `DecoratorTypeReference` node in the metamodel AST.
+ */
+export interface DecoratorTypeReferenceArgument {
```
mttrbrts, line 31: Why can't we reuse a type from concerto-metamodel here?

packages/concerto-core/src/introspect/decorator.ts
```diff
@@ -53,8 +69,8 @@ class Decorator {
     * @param {*} [fileLocation] the file location
     * @private
     */
-    handleError(level, err) {
-        Logger.dispatch(level, err);
+    handleError(level: string | undefined, err: unknown): void {
```
mttrbrts: err: Error?

packages/concerto-core/src/introspect/mapdeclaration.ts
```diff
@@ -71,10 +71,9 @@ class MapDeclaration extends Declaration {
             throw new IllegalModelException(`MapDeclaration must contain valid MapValueType, for MapDeclaration ${this.ast.name}` , this.modelFile, this.ast.location);
         }
 
-        this.name = this.ast.name;
+        // name and fqn are already set by Declaration.process() above
```
mttrbrts: Double check that this is true

packages/concerto-core/src/introspect/property.ts
```diff
@@ -34,21 +36,26 @@ import type ModelFile from './modelfile';
...
+    // initialiser here would run before process() and be overwritten anyway.
+    name!: string;
+    /**
+     * Vestigial: only ever set to null, never read. Retained for compatibility.
+     */
+    decorator!: Decorator | null;
+    type!: string | null;
+    array!: boolean;
+    sizeValidator!: CollectionSizeValidator | null;
+    optional!: boolean;
```
mttrbrts: Consider removing

packages/concerto-core/src/introspect/stringvalidator.ts
```diff
@@ -41,7 +49,7 @@ class StringValidator extends Validator{
      *
      * @throws {IllegalModelException}
      */
-    constructor(field, validator, lengthValidator) {
+    constructor(field: ValidatedElement, validator: AstNode, lengthValidator?: AstNode) {
```
mttrbrts: Tighten validator type?

packages/concerto-core/src/model/identifiable.ts
```diff
@@ -25,8 +32,14 @@ import Typed from './typed';
...
 class Identifiable extends Typed {
-    $identifierFieldName: any;
-    $identifier: any;
+    $identifierFieldName: string;
+    // Set via setIdentifier(), called from this constructor rather than
+    // assigned directly, so TS can't see the assignment for definite-assignment
+    // analysis; the field is genuinely absent for non-identified instances.
+    $identifier!: string | undefined;
+    // Genuinely polymorphic at runtime: JSDoc'd as a string, but
+    // Factory#newResource stores a Dayjs instance (or null) here for
```
mttrbrts: Should it do this? I suspect that this is a bug, and we should instead store a string in Factory.newResource

packages/concerto-core/src/decoratorextractor.ts
```diff
@@ -481,7 +498,7 @@ class DecoratorExtractor {
     */
     processModels(){
         const processedModels = this.sourceModelAst.models.map(model =>{
-            if ((model?.decorators?.length > 0)){
+            if (model?.decorators?.length) {
```
mttrbrts: Is this safe? or necessary?

packages/concerto-core/src/introspect/collectionsizevalidator.ts
mttrbrts: Yes — done in f62d4a8. It now takes `ICollectionSizeValidator` from concerto-metamodel, so `validator.minSize` / `maxSize` are `number | undefined` rather than `any`.

The base `Validator.validator` is a union of the metamodel validator nodes:

```ts
export type NumberDomainValidatorAst = IIntegerDomainValidator | ILongDomainValidator | IDoubleDomainValidator;
export type ValidatorAst =
    | ICollectionSizeValidator
    | IStringRegexValidator
    | IStringLengthValidator
    | NumberDomainValidatorAst;
```

and each subclass narrows it with `declare validator: ICollectionSizeValidator`. `declare` emits nothing, so it can't clobber what the base constructor assigned — which is why it's safe here where a plain redeclaration would not be.

---

packages/concerto-core/src/introspect/stringvalidator.ts
mttrbrts: Done in f62d4a8:

```ts
constructor(field: ValidatedElement, validator?: IStringRegexValidator, lengthValidator?: IStringLengthValidator)
```

with `declare validator: IStringRegexValidator | undefined;` on the class, so `this.validator?.pattern` and `?.flags` in `compatibleWith` are `string` now instead of `any`. Both params are optional because the caller in `Field.process()` fires when *either* `ast.validator` or `ast.lengthValidator` is present, so `validator` genuinely can be undefined.

`minLength`/`maxLength` stay `number | null | undefined` — the metamodel marks both optional, and `NO_PARMS_IN_AST` in the existing tests passes them as explicit nulls, so all three states really do occur.

---

packages/concerto-core/src/introspect/decorator.ts
mttrbrts: I looked for one and there isn't a fit, because `process()` doesn't store the metamodel nodes — it projects them into a different shape:

```ts
if (thing.$class === `${MetaModelNamespace}.DecoratorTypeReference`) {
    this.arguments.push({ type: 'Identifier', name: thing.type.name, array: thing.isArray });
} else {
    this.arguments.push(thing.value);   // unwrapped to the raw literal
}
```

So the metamodel's `IDecoratorTypeReference` is `{ $class, type: ITypeIdentifier, isArray }`, whereas what lands in `arguments` is `{ type: 'Identifier', name, array }` — `type` means something different in each (a `TypeIdentifier` object vs. the literal string `'Identifier'`), `name` is flattened out of `type.name`, and `isArray` is renamed. Every other literal is unwrapped to its bare `.value`, discarding `$class` entirely, so `IDecoratorLiteral` / `DecoratorLiteralUnion` don't describe it either.

I did use the metamodel where the AST *is* stored as-is — `this.name = (this.ast as IDecorator).name` a few lines down.

Worth noting the alternative: `getArguments()` could return `DecoratorLiteralUnion[]` straight from the AST and let callers narrow on `$class`. That's a cleaner API and 5.0.0 is the moment for it, but it's a breaking change to a public method rather than a typing fix, so I've left it out of this PR. Say the word and I'll do it as a follow-up.

---

packages/concerto-core/src/introspect/decorator.ts
mttrbrts: `string | Error`, and changed to that in f62d4a8. The two call shapes differ:

- The eight calls inside `validate()` pass a message string they've just built — e.g. ``const err = `Decorator ${this.getName()} has too few arguments...` ``.
- The one call in the `catch` passes the caught exception, which is an `Error` (including the `IllegalModelException` an inner `handleError` may have thrown when the level is `'error'`).

`unknown` was me being lazy about that; `string | Error` is what the code actually does. The catch site needs `err as Error` since `useUnknownInCatchVariables` is on under `strict`.

---

packages/concerto-core/src/introspect/property.ts
mttrbrts: Removed in f62d4a8 — both the declaration and the `this.decorator = null;` in `process()`.

Confirmed nothing reads it: `grep -rn '\.decorator\b'` across every package's `src` and `test` (excluding `.decorators`) returns only that assignment and unrelated `command.decorator` hits in DecoratorManager/dcsconverter. It was write-only, and always `null`.

---

packages/concerto-core/src/introspect/mapdeclaration.ts
mttrbrts: Checked, and it holds. Parsing `map Dict { o String o String }` with the lines removed:

```
name: "Dict" | fqn: "org.acme@1.0.0.Dict"
getName(): Dict | getFullyQualifiedName(): org.acme@1.0.0.Dict
Declaration.process assigns name: true | assigns fqn: true
MapDeclaration.process assigns name: false | assigns fqn: false
```

The ordering works out because `Declaration`'s constructor sets `this.modelFile = modelFile` *before* it calls `this.process()`, so by the time `super.process()` computes `fqn` from `this.modelFile.getNamespace()` the model file is there. The two removed lines were byte-equivalent to what the base had already done (`this.ast.name` is what `this.name` was set from).

I've made the comment name the mechanism rather than just assert it.

While verifying this I noticed something adjacent I did **not** touch: `MapDeclaration`'s constructor is

```ts
constructor(modelFile, ast) {
    super(modelFile, ast);      // Declaration's ctor already runs this.process()
    this.modelFile = modelFile; // already set by super
    this.process();             // ...so this is the second run
}
```

`process()` runs twice, building a throwaway `MapKeyType`/`MapValueType` pair on the first pass. Harmless but wasteful, and dropping the last two lines looks safe. It's a runtime change rather than a typing one so I've left it alone — happy to do it here or as a follow-up if you'd like.

---

packages/concerto-core/src/model/identifiable.ts
mttrbrts: I dug into this and I don't think it's a bug — but the JSDoc that says `@param {string} timestamp` is wrong, and that's almost certainly what prompted the question. Fixed in f62d4a8.

Every DateTime value in the object model is a Dayjs, not just `$timestamp`. `JSONPopulator` builds one on the way in:

```ts
case 'DateTime': {
    ...
    result = dayjs.utc(json).utcOffset(this.utcOffset);
```

and `JSONGenerator` calls Dayjs methods on the way out:

```ts
case 'DateTime': {
    const objWithOffset = obj.utc().utcOffset(this.utcOffset);
```

Since `addTimestampField()` adds `$timestamp` as a `DateTimeProperty`, it goes through exactly that path. Storing a string in `Factory.newResource` would make it the only DateTime-valued field held as a string, and `toJSON()` on any transaction or event would throw on `obj.utc()`.

Round-tripping a transaction confirms it:

```
typeof $timestamp: object | is dayjs: true
toJSON: {"$class":"org.acme@1.0.0.Tx","txId":"a1","$timestamp":"2026-08-29T19:04:07.243Z"}
round-trip $timestamp: Dayjs
```

So the string form is the *serialized* representation, which is what `test/model/resource.js` asserts on, and the in-memory form is a Dayjs both before and after a round trip.

It's now typed `$timestamp?: Dayjs | null` rather than `any`, and the `@param`/`@return` on `Identifiable`, `Resource`, `Relationship` and `getTimestamp()` say `Dayjs`.

---

packages/concerto-core/src/decoratorextractor.ts
mttrbrts: Necessary, yes — and safe, but I've made it more obviously so.

Necessary because the original was `model?.decorators?.length > 0`, and `decorators` is optional in the metamodel, so `length` is `number | undefined`; comparing that with `>` is an error under `strictNullChecks`. It only compiled before because `model` was `any`.

Safe because the two forms agree on every input:

| `decorators` | `?.length > 0` | `?.length` (truthy) |
|---|---|---|
| `undefined` | false | false |
| `null` | false | false |
| `[]` | false | false |
| `[x]` | true | true |
| `[x,y]` | true | true |

(`undefined > 0` is false, and `0` is falsy.) Same for `model` itself being undefined.

That said, "no `> 0`" is a worse thing to have to reason about than the comparison it replaced, so f62d4a8 keeps the comparison explicit instead:

```ts
if ((model?.decorators?.length ?? 0) > 0) {
```

Same behaviour, and it reads as the original did.

---

PR #1309, opened 2026-09-02
fix(concerto-util): normalize identifiers with a leading continuation char

rajanpanth:
### What

`normalizeIdentifier` promises to turn arbitrary strings into valid Concerto identifiers, but it only guards a leading **digit**. Combining marks and connector punctuation are valid *continuation* characters (`ID_REGEX` allows `\p{Mn}`, `\p{Mc}`, `\p{Pc}` after the first char) while being invalid at the start, and a string that begins with one slips through every substitution untouched and dies on the function's own final check:

```js
normalizeIdentifier('́name');  // combining acute + "name"
// Error: Unexpected error. Not able to escape identifier '́name'.
normalizeIdentifier('‿foo');   // undertie (Pc) + "foo"
// Error: Unexpected error. Not able to escape identifier '‿foo'.
```

The message itself calls this path unexpected, and the existing test table shows the same characters are fully supported mid-string (`'foo‿bar'`, `'पः'` are no-op rows).

### The change

Step 1 now prefixes the underscore for any character that is a valid continuation but not a valid start: digits, combining marks, and connectors other than `_` itself, which must stay excluded so `'_class'` keeps round-tripping unchanged:

```diff
-        .replace(/^\p{Nd}/u, '_$&')
+        .replace(/^(?:\p{Nd}|\p{Mn}|\p{Mc}|(?!_)\p{Pc})/u, '_$&')
```

`$` is untouched (it is not `Pc`), leading ZWJ/ZWNJ already worked through step 2, and the empty-string throw asserted by the existing test is unchanged.

### Verification

Four rows added to the test table (leading undertie, leading combining mark, leading spacing combining mark, combining mark alone). Full `concerto-util` suite: 155 passing. Reverting only `src/identifiers.ts` fails exactly those 4 and leaves the other 151 passing, so the new rows isolate the change.

### DIFF
--- a/packages/concerto-util/src/identifiers.ts
+++ b/packages/concerto-util/src/identifiers.ts
@@ -43,9 +43,11 @@ export function normalizeIdentifier(identifier: string | null | undefined, trunc
         throw new Error(`Unsupported identifier type, '${typeof result}'.`);
     }
 
-    // 1. If the identifier begins with a number, add a leading underscore
+    // 1. If the identifier begins with a character that is only valid as a
+    // continuation, a digit, a combining mark or a connector other than _,
+    // add a leading underscore
     result = result
-        .replace(/^\p{Nd}/u, '_$&')
+        .replace(/^(?:\p{Nd}|\p{Mn}|\p{Mc}|(?!_)\p{Pc})/u, '_$&')
 
     // 2. Substitute Whitespace, and joiners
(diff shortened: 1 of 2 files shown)


mttrbrts (approved):
Great catch @rajanpanth, 🙏🏻 .

Can you fix the DCO sign-off?

Issue #302, opened 2021-08-04
Regular expressions do not work with flags


jeromesimeon:
# Bug Report 🐛

Regular expressions with flags e.g., `/^[a-zA-Z0-1]*^/u` do not work.

## Expected Behavior

Regular expressions should work with flags.

## Current Behavior

With the given CTO:

```
namespace org.test

concept Person {
	o String name regex=/^[a-zA-Z][a-zA-Z0-9]*$/u
	o Double age
}
```

And input data:
```
{
    "$class": "org.test.Person",
    "name": "John",
    "age": 33
}
```

The validation fails:

```
zsh-5.8$ concerto validate --model test.cto --input person.json
7:56:56 AM - INFO: Input is invalid
7:56:56 AM - ERROR: Validator error for field undefined org.test.Person.name: Value + 'John' failed to match validation regex: /^[a-zA-Z][a-zA-Z0-9]*$\//
Error: Validator error for field undefined org.test.Person.name: Value + 'John' failed to match validation regex: /^[a-zA-Z][a-zA-Z0-9]*$\//
    at StringValidator.reportError (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/introspect/validator.js:44:15)
    at StringValidator.validate (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/introspect/stringvalidator.js:56:22)
    at ResourceValidator.checkItem (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/serializer/resourcevalidator.js:327:42)
    at ResourceValidator.visitField (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/serializer/resourcevalidator.js:221:22)
    at ResourceValidator.visit (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/serializer/resourcevalidator.js:77:25)
    at Field.accept (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/introspect/decorated.js:65:24)
    at ResourceValidator.visitClassDeclaration (/usr/local/lib/node_modules/@accordproject/concerto-cli/node_modules/@accordproject/concerto-core/lib/serializer/resourcevalidator.js:174:26)
[...]

PR #1310, opened 2026-09-03
fix(concerto-core): reset lastIndex before testing a stateful validator regex

rajanpanth:
### Changes

- `StringValidator` gains a `matchesRegex(value)` method that resets `lastIndex` to 0 before calling `test`, and `validate()` uses it.
- `Factory.newResource`'s identifier-regex check uses the same method instead of calling `test` on the raw regex.

### Motivation

A regex with the `g` or `y` flag is stateful: `test()` advances `lastIndex`, so the same value alternates between matching and failing on consecutive calls. The CTO grammar parses `RegularExpressionFlags` (parser.pegjs), so this is reachable from any model:

```text
o String code regex=/^[A-Z]{3}$/g
```

Executed against main, serializing the same valid instance repeatedly:

```
toJSON #0 (code=ABC) -> OK
toJSON #1 (code=ABC) -> ERROR: Value 'ABC' failed to match validation regex: /^[A-Z]{3}$/g
toJSON #2 (code=ABC) -> OK
toJSON #3 (code=ABC) -> ERROR ...
```

`Factory.newResource` with a `g`-flagged identifier regex shows the same alternation on the id check.

### Tests

Four regression tests: repeated validation and repeated rejection with a `g` flag, repeated validation with a `y` flag, and three consecutive `newResource` calls with a `g`-flagged id regex. Reverting only the source while keeping the tests fails them with the exact alternating symptom; restoring passes the full 83-test stringvalidator plus factory suites.

### Related issues

Closest prior work is #302, which established that regex flags must be honored; nothing covers the statefulness of `g`/`y` under repeated validation.

Changes are backwards compatible: values accepted or rejected by a flagless regex are unaffected.

### DIFF
--- a/packages/concerto-core/src/factory.ts
+++ b/packages/concerto-core/src/factory.ts
@@ -127,7 +127,7 @@ class Factory {
                     idFullField = idFullField.getScalarField();
                 }
                 // if regex on identifier field & provided id does not match regex, throw error
-                if(idFullField?.validator?.regex && (idFullField.validator?.regex.test(id) === false)) {
+                if(idFullField?.validator?.regex && (idFullField.validator?.matchesRegex(id) === false)) {
                     throw new Error('Provided id does not match regex: ' + idFullField?.validator?.regex);
                 }
             }
--- a/packages/concerto-core/src/introspect/stringvalidator.ts
+++ b/packages/concerto-core/src/introspect/stringvalidator.ts
@@ -97,10 +97,27 @@ class StringValidator extends Validator{
                 this.reportError(identifier, `The string length of '${value}' should not exceed ${this.maxLength} characters.`);
             }
 
-            if (this.regex && !this.regex.test(value)) {
+            if (this.regex && !this.matchesRegex(value)) {
                 this.reportError(identifier, `Value '${value}' failed to match validation regex: ${this.regex}`);
             }
         }
+    }
+
+    /**
+     * Tests a value against the validation regex. Returns true when no regex is
+     * specified.
+     * @param {string} value the value to test
+     * @returns {boolean} true if the value matches the validation regex
+     */
+    matchesRegex(value) {
+        if (!this.regex) {
+            return true;
+        }
+        // `test` advances `lastIndex` when the regex has the global or sticky flag, so
+        // testing the same value twice would otherwise alternate between matching and
+        // not matching. Reset it to make every test independent of previous ones.
+        this.regex.lastIndex = 0;
+        return this.regex.test(value);
     }
 
     /**
--- a/packages/concerto-core/src/introspect/stringvalidator.ts
+++ b/packages/concerto-core/src/introspect/stringvalidator.ts
@@ -115,9 +115,14 @@ class StringValidator extends Validator{
         }
         // `test` advances `lastIndex` when the regex has the global or sticky flag, so
         // testing the same value twice would otherwise alternate between matching and
-        // not matching. Reset it to make every test independent of previous ones.
+        // not matching. Reset it before and after so that every test is independent of
+        // previous ones, and so that getRegex() never hands out a poisoned object.
         this.regex.lastIndex = 0;
-        return this.regex.test(value);
+        try {
+            return this.regex.test(value);
+        } finally {
+            this.regex.lastIndex = 0;
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


mttrbrts (commented):
This is an interesting find. Are there any other artifacts from regex.text that pollute shared variables?

Validation should be fully deterministic and idempotent.

rajanpanth:
Good question, so I went looking. Short answer: this validator is the only leak of its kind in the four packages, but it is fully reachable from a user model, and I found one more hole in my own fix, now pushed as f88a293.

**Scope of the audit.** I went through every regex site in `concerto-core`, `concerto-util`, `concerto-cto` and `concerto-vocabulary`. 31 carry a `g` or `y` flag or are built dynamically.

**The only leak is this one.** `StringValidator.regex` is the single regex in those packages that is both retained on an instance across calls and passed to `.test()`. Its two consumers, `stringvalidator.ts` and `factory.ts`, share the same object, so the pollution crosses subsystems.

Everything else is safe, and for concrete reasons rather than by assumption:

- The other 28 global-flagged sites (`identifiers.ts`, `label.ts`, `httpfileloader.ts`, `vocabularymanager.ts`, `basemodelmanager.ts`, the generated parser) are all **inline literals handed straight to `replace` or `split`**, never hoisted to a variable. A regex literal evaluates to a fresh object each time, and `String.prototype.replace` with a global regex resets `lastIndex` to 0 both before and after; I poisoned one to 99 and it still produced the correct result and ended at 0. `split` never consults it.
- `globalize.ts:29` builds a fresh `RegExp` per iteration and feeds it to `replace`.
- The shared module-level `ID_REGEX` in both `concerto-util/identifiers.ts` and `concerto-core/modelutil.ts` carries only the `u` flag. I set `lastIndex` to 12345 and `.test()` returned identical results three times running.
- In the generated `concerto-cto` parser, the distinct flag sets across all 30 `peg$rN` constants are exactly `(none)` and `i`. None carry `g` or `y`.
- Outside regexes, there are no module-level accumulators in the validation path, `Validator` is stateless, `ResourceValidator` is constructed per `toJSON` with its stack in per-call parameters, and `Math.random` appears only in sample generation, never in validation.

**Reachability, which I think is the important part.** `parser.pegjs:368` defines `RegularExpressionFlags = IdentifierPart*`, so flags are taken verbatim from the model and passed unfiltered to `new RegExp(pattern, flags)`. There is no allowlist. From a plain `.cto`:

```
concept P identified by id {
  o String id
  o String c regex=/^[a-z]+@[a-z]+.com$/g
}

call 1: PASS   call 2: FAIL   call 3: PASS   call 4: FAIL
```

So any modeller writing `regex=/.../g` hits this. The reason it went unnoticed is that **no test model in the repo uses `g` or `y`**; across all 206 `.cto` files only `u` appears. Two consequences worth knowing: a validator whose `default=` value triggers the constructor's own validation can be poisoned at model-load time, so `factory.newResource` then throws on its very first call for a model whose default matches its own regex; and two different valid records validated in sequence give a stable, non-flaky wrong answer where the second is always rejected.

**Idempotency spot-check.** Parsing all 206 test models three times each gave 189 stable successes and 17 stable failures (the negative fixtures) with zero non-determinism; 112 models introspected twice through `ModelManager` and 133 declarations serialized three times through `Serializer.toJSON({validate: true})` gave zero mismatches. The harness is not vacuously passing, since injecting a single `g` flag into one model made it fire immediately.

**What I changed in response.** The audit caught a hole in my own patch: `matchesRegex` reset `lastIndex` before the test but left it dirty afterwards, so a caller of the public `getRegex()` still got a poisoned object. Pushed f88a293 to reset in a `finally` as well, with a test asserting `getRegex().lastIndex === 0` after a validate; it fails with `expected 9 to equal +0` without the change.

Two things I did not do, happy to if you want them here rather than separately: adding a `.cto` fixture using `regex=/.../g` so this class stays covered, and deciding whether the parser should reject `g` and `y` outright, since neither flag is meaningful for a whole-value validation and rejecting them would remove the footgun at the source. One caveat on the `lastIndex` approach in general: the `regExp` ModelManager option lets callers swap in another engine, and a hypothetical engine holding state elsewhere would not be covered. XRegExp, which the tests use, produces real `RegExp` instances so it is fine.

mttrbrts:
I'm sensitive to changing the language scope (i.e. removing support for g & y outright). 

I'm happy to merge this as is, but it does reveal some wider issues:
1. We should more robustly ensure statelessness (even for custom regex engines). This will come with a performance tradeoff though, as instantiating RegExps can be expensive.
2. We should be explicit about the regexp modifiers that we support.

mttrbrts (approved):

PR #1311, opened 2026-09-03
fix(concerto-core): use the nested id field's regex when generating samples

rajanpanth:
### Changes

One line in `InstanceGenerator.getFieldValue`: the identifier-regex branch now reads `idField.validator.regex`, the variable its own guard just checked, instead of `fieldOrScalarDeclaration.validator.regex`, which is the outer field being visited.

### Motivation

Generating a sample for a concept that contains an identified type whose id carries a regex crashes on main:

```text
concept Inner identified by ssn { o SSN ssn }
scalar SSN extends String regex=/^\d{3}-\d{2}-\d{4}$/
concept Outer { o Inner inner }
```

```
factory.newConcept(ns, 'Outer', 'id', { generate: 'sample' })
ERROR: TypeError - Cannot read properties of undefined (reading 'regex')
```

The guard checks `idField?.validator?.regex` but the call then dereferences the outer field's validator, which is null here. When the outer field does have its own regex validator, the id is silently generated from the wrong regex instead of crashing.

After the fix the same call produces `"580-41-7163"`, matching the id's regex.

### Tests

Regression test added in `test/serializer/instancegenerator.js`: generating a sample for a nested identified type with a regex-constrained id yields an id matching that regex. Reverting only the source while keeping the test reproduces the exact `TypeError` above; restoring passes all 41 instancegenerator tests.

### Related issues

The code was introduced by #178 (issue #171); neither covers the nested case.

Changes are backwards compatible.

### DIFF
--- a/packages/concerto-core/src/serializer/instancegenerator.ts
+++ b/packages/concerto-core/src/serializer/instancegenerator.ts
@@ -172,7 +172,7 @@ class InstanceGenerator {
                     idField = idField.getScalarField();
                 }
                 if(idField?.validator?.regex){
-                    id = parameters.valueGenerator.getRegex(fieldOrScalarDeclaration.validator.regex);
+                    id = parameters.valueGenerator.getRegex(idField.validator.regex);
                 } else {
                     id = this.generateRandomId(declaration);
                 }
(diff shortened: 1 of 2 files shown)


rajanpanth:
Addressed: the id path now forwards the id field validator's minLength/maxLength to getRegex, mirroring the primitive-string call site above it. Pushed fcbebee with a repeated-sampling regression test (an id field with regex=/^[a-z]+$/ length=[2,4] stays within bounds across 15 samples; with the bounds dropped the test fails with values like length 20). Note this pairs with the reshaping fix in #1312, which is what makes getRegexString respect both facets at once.

mttrbrts (approved):
Lovely!

PR #1312, opened 2026-09-03
fix(concerto-core): stop mangling regex samples that already satisfy length

rajanpanth:
### Changes

`getRegexString` in `valuegenerator.ts` only reshapes the RandExp output when the value actually violates the declared length bounds, via a new `isLengthInRange` helper. Previously it always padded or truncated to a random length inside the range, even when the generated value was already compliant.

### Motivation

A scalar combining both facets produces samples that fail its own model:

```text
scalar Code extends String regex=/^[a-z]{3}$/ length=[1,10]
```

Executed against main, 20 sampled instances: 19 of 20 values failed the regex (`"etuhkbm"`, `"tfhn"`, `"iqtwhqirwx"`, ...) and 20 of 20 failed `toJSON` validation of the very instance the factory had just generated. After the fix: 0 of 20 on both counts.

The contract comes from the feature itself: #178 introduced this code as "Generate valid values for Regex and Range meta properties". For facets that genuinely conflict (a regex that cannot produce any in-range length), the previous best-effort reshaping still applies, so no existing behavior is lost.

### Tests

Two additions: a repeated-sampling test asserting a fixed-length regex stays valid across four length-bound combinations, and an end-to-end factory sample test in `test/serializer/instancegenerator.js`. Reverting only the source while keeping the tests fails deterministically across repeated runs; restoring passes all 67 tests in the two touched suites.

Changes are backwards compatible.

### DIFF
--- a/packages/concerto-core/src/serializer/valuegenerator.ts
+++ b/packages/concerto-core/src/serializer/valuegenerator.ts
@@ -134,6 +134,25 @@ const getString = (minLength, maxLength) => {
 
 
 /**
+ * Determine whether the length of a value is within the given bounds. A null or
+ * undefined bound is not enforced.
+ * @param {string} value the value to test.
+ * @param {number} minLength the lower bound on the range, inclusive.
+ * @param {number} maxLength the upper bound on the range, inclusive.
+ * @return {boolean} true if the length of the value is within the bounds.
+ * @private
+ */
+const isLengthInRange = (value, minLength, maxLength) => {
+    if (minLength !== null && minLength !== undefined && value.length < minLength) {
+        return false;
+    }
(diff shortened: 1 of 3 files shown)


rajanpanth:
Good catch on the zero bound. The truthiness gate was inherited from the original code and made length=[0,0] read as no bounds at all, so the out-of-range sample was returned untouched. Pushed 9c6d80b: the gate is now just `!isLengthInRange(...)`, since the helper already treats null and undefined as unenforced, plus a regression test that a regex scalar with length=[0,0] generates the empty string. Verified the new test fails with the old gate (sample stayed 37 chars long) and passes now; suite is 68 passing.

mttrbrts (approved):

PR #1314, opened 2026-09-03
fix(concerto-cto): reject duplicate validators instead of silently keeping the last

rajanpanth:
### Changes

- #1305 made validators order-independent by collecting `(validator)*` and bucketing by `$class` with plain assignment. A repeated validator of the same kind now parses and silently keeps only the last one: `regex=/a/ regex=/b/` drops `/a/`, `range=[0,10] range=[500,600]` keeps only `[500,600]`, and the same holds for `length` and `size`, on both fields and scalars. Before #1305 the grammar rejected all of these, and the metamodel has exactly one slot per validator kind, so last-wins silently discards a modeler's constraint and weakens validation. #1305's stated intent was ordering flexibility only.
- The grammar initializer gains a `bucketValidators(items, kindByClass)` helper that raises a located peggy `error('Duplicate <kind> validator')`, surfacing as a normal `ParseException`; the five reducer rules use it. `src/parser.js` is regenerated via `npm run parser`, matching the convention of committing the generated parser.

Executed comparison, old parser (HEAD~1 grammar) vs current main:

```
dup regex (field)   | old: rejected | main: ACCEPTED, kept pattern "b"
dup length (field)  | old: rejected | main: ACCEPTED, kept [30,40]
dup range (field)   | old: rejected | main: ACCEPTED, kept [500,600]
dup size (field)    | old: rejected | main: ACCEPTED, kept [9,10]
dup regex (scalar)  | old: rejected | main: ACCEPTED, kept pattern "b"
```

### Tests

Seven cases added under the `validator ordering` describe in `test/parserMain.js`: duplicate regex and length on a String field, duplicate range on Integer, Double and Long, duplicate size on an array, and duplicate regex on a scalar. With the grammar and generated parser reverted and the tests kept, all seven fail with `expected [Function] to throw an error`; restored, the concerto-cto suite passes 177/177 and every any-order valid case from #1305 still parses. eslint clean.

### DIFF
--- a/packages/concerto-cto/src/parser.js
+++ b/packages/concerto-cto/src/parser.js
@@ -996,49 +996,39 @@ function peg$parse(input, options) {
   };
   var peg$f88 = function(v) { return v; };
   var peg$f89 = function(items) {
-      const result = {};
-      for (const v of items) {
-        if (v.$class === `${metamodelNamespace}.StringRegexValidator`) { result.regex = v; }
-        else if (v.$class === `${metamodelNamespace}.StringLengthValidator`) { result.length = v; }
-      }
-      return result;
+      return bucketValidators(items, {
+        [`${metamodelNamespace}.StringRegexValidator`]: 'regex',
+        [`${metamodelNamespace}.StringLengthValidator`]: 'length',
+      });
    };
   var peg$f90 = function(v) { return v; };
   var peg$f91 = function(items) {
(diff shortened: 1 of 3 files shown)


mttrbrts (approved):
You're on a roll @rajanpanth . Thanks!

cc: @muhabdulkadir

PR #1318, opened 2026-09-04
fix(concerto-util): download a shared external model URI only once

rajanpanth:
### Changes

`FileDownloader.downloadExternalDependencies` builds its root-level jobs by flattening `Object.keys(externalImports)` across all files, with no deduplication, and without consulting the `downloadedUris` cache:

```ts
const jobs = flatten(files.map(file => {
    const externalImports = this.getExternalImports(file);
    return Object.keys(externalImports).map(importDeclaration => ({
        downloadedUris, url: externalImports[importDeclaration], options
    }));
}));
```

Several import declarations commonly share one URI, for example `import x.A from <uri>` and `import x.B from <uri>` in the same model file. Each declaration then produced its own download and its own entry in the returned array.

The intent is clear elsewhere in the same class: `runJob` deduplicates its recursive imports with `Array.from(new Set(...))` and keeps a `downloadedUris` cache commented "cache the URI, so we don't download it again". Only the root level skipped both.

Executed, before the change, with two import declarations on one URI plus a second file importing the same URI:

```
  loader.load -> https://models.example.com/ext.cto (call #1)
  loader.load -> https://models.example.com/ext.cto (call #2)
result       = ["EXT","EXT"]
result.length= 2 (expected 1)
load calls   = 2 (expected 1)
```

After: one call, `result = ["EXT"]`.

### Change

The URIs are deduplicated before jobs are created, mirroring the `Array.from(new Set(...))` idiom already used in `runJob`.

### Tests

`should download a shared external dependency only once` in `test/filedownloader.js`, following the existing `sinon.createStubInstance(GitHubFileLoader)` style. It asserts both the returned array and `load.withArgs(...).callCount === 1`.

Reverting only `filedownloader.ts` and keeping the test fails it with three entries where one is expected. Restored, the package passes 152/152 and eslint is clean.

### DIFF
--- a/packages/concerto-util/src/filedownloader.ts
+++ b/packages/concerto-util/src/filedownloader.ts
@@ -90,13 +90,21 @@ class FileDownloader<TFile = unknown, TSeed = TFile> {
 
         const downloadedUris = new Set<string>();
 
-        const jobs: Array<DownloadJob<TFile>> = flatten(files.map(file => {
-            const externalImports = this.getExternalImports(file);
-            return Object.keys(externalImports).map(importDeclaration => ({
-                downloadedUris: downloadedUris,
-                url: externalImports[importDeclaration],
-                options: options
-            }));
+        // several import declarations, in one file or across files, can share a
+        // URI, so the URIs are deduplicated before jobs are created for them
+        const importedUris = Array.from(
+            new Set(
+                flatten(files.map(file => {
+                    const externalImports = this.getExternalImports(file);
+                    return Object.keys(externalImports).map(importDeclaration => externalImports[importDeclaration]);
+                }))
+            )
+        );
+
+        const jobs: Array<DownloadJob<TFile>> = importedUris.map(url => ({
+            downloadedUris: downloadedUris,
+            url: url,
+            options: options
         }));
 
         return PromisePool
--- a/packages/concerto-util/test/filedownloader.js
+++ b/packages/concerto-util/test/filedownloader.js
@@ -106,6 +106,50 @@ describe('FileDownloader', () => {
                 });
         });
 
+        it('should download a shared external dependency only once', function() {
+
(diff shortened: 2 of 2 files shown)


rajanpanth:
Closing this one. The CI failure turned out to be informative rather than incidental, and it shows the change is not the clean win I presented it as.

The red legs are not in `concerto-util` at all. `concerto-cto` enforces 100% coverage and drops to 99.29%, exactly 2 statements and 2 branches:

```
Statements : 99.29% ( 280/282 )
Branches   : 98.99% ( 197/199 )
```

Those two statements are in `concerto-cto/src/external.ts`, in `updateModels`:

```ts
priors.forEach((priorModel) => {
    if (priorModel.namespace === newNamespace) {
        result.models.push(newModel);
        found = true;
    } else {
```

That branch handles a downloaded model whose namespace is already present, and the suite reaches it precisely because a shared URI is currently downloaded once per import declaration, so the same namespace arrives more than once. Deduplicating the downloads removes the only path the tests use to get there.

So the two layers are coupled: `downloadExternalDependencies` produces the duplicates that `updateModels` exists to absorb. Deduplicating in the downloader alone silently makes that defensive branch unreachable, and whether it should then be removed, kept for other callers, or whether the dedup belongs in `external.ts` instead is a design call that belongs to you rather than to a drive-by PR.

The underlying observation still stands, if you want it: `runJob` already dedupes with `Array.from(new Set(...))` and keeps a `downloadedUris` cache, while the root-level jobs array does neither, so N import declarations sharing one URI produce N downloads. Verified with a stubbed loader: two declarations on one URI gave `load` call count 2 and a result array of `["EXT","EXT"]`.

Happy to redo it as a coordinated change across both packages if you would like it, or to open an issue recording the duplicate downloads so it is not lost. I would rather not land something that quietly strands working code in another package.

Issue #1319, opened 2026-09-04
Decouple Serializer/Factory from BaseModelManager to enable tree-shaking for AST-only consumers (~25 KB gzip)


mttrbrts:
## Summary

`BaseModelManager` eagerly constructs a `Factory` and a `Serializer` in its
constructor. Because these are **static imports used by a retained class**, they
hard-link the entire instance/serialization layer into *every* model manager —
including `AstModelManager`, which AST-only consumers use purely for
introspection. This defeats the import-level tree-shaking that 5.0.0 was
designed to enable (#1304 / #1306): a consumer that only reads the model
structure still ships `urijs`, `dayjs`, `lorem-ipsum` and `randexp`.

Decoupling the serialization layer from the introspection path would remove
**~25 KB gzip (~36%)** from an AST-only consumer's Concerto footprint.

## Background

This came out of migrating a downstream widget to `concerto-core` 5.0.0-beta and
measuring the host bundle. Two wins were already identified:

1. Upgrading to 5.0.0's ESM build (tree-shakeable exports) took the concerto
   footprint from **161.1 KB → 102.3 KB gzip**.
2. Switching the consumer from `ModelManager` to `AstModelManager` (the manager
   never parses CTO, only feeds ASTs) dropped `concerto-cto` (the Peggy parser)
   entirely: **102.3 KB → 69.5 KB gzip**. This required exporting
   `AstModelManager` from the package root (small additive change).

This issue is about the **next** win, which is structural.

## The problem

For an AST-only consumer that imports just:

```ts
import {
  AstModelManager, ClassDeclaration, EnumDeclaration,
  Field, ModelFile, Property, ScalarDeclaration,
} from '@accordproject/concerto-core';
```

the reachable graph still pulls the full instance/serialization layer via:

```
AstModelManager → BaseModelManager (constructor)
    this.factory    = new Factory(this);
    this.serializer = new Serializer(this.factory, this, options);

Factory / Serializer → model/resource → model/resourceid   → urijs
Serializer → serializer/valuegenerator                     → lorem-ipsum, randexp
Serializer → serializer/jsonpopulator + model/* + datetimeutil → dayjs
```

`getFactory()` / `getSerializer()` are the only consumers of these modules on
the introspection path, and the sole *internal* use is `validateAst()`
(`this.getSerializer().fromJSON(...)`), which only runs under the optional
`metamodelValidation` flag.

## What I verified

[...]

PR #1320, opened 2026-09-04
refactor(concerto-core): replace urijs and lorem-ipsum with local variants

mttrbrts:
## Summary

Removes the `urijs` and `lorem-ipsum` runtime dependencies from `concerto-core` by inlining the small surface each was actually used for. This is a pure internal refactor: **no public API or behavior change**. It shrinks the browser bundle and reduces supply‑chain surface.

Analysis of actual usage showed each dependency had exactly one consumption site and used only a tiny slice of its API:

- **`urijs`** — only `URIJS.parse(uri)` in `model/resourceid.ts`, to parse the constrained `resource:ns.Type#id` scheme. The full RFC 3986 parser (punycode, IPv6, second‑level‑domain tables) was ~94 KB raw / ~20 KB gzip in the browser bundle.
- **`lorem-ipsum`** — only in the `getString` sample‑value helper in `serializer/valuegenerator.ts`, to produce throwaway random text for `{ generate: 'sample' }`. The output is non‑semantic, so a small local word‑list generator is equivalent.

`randexp` is **retained** — reverse‑generating strings from arbitrary user regexes is a genuinely specialised task and the wrong thing to reimplement.

## Changes

• Replaced `urijs` with a local `parseUri()` helper in `model/resourceid.ts` that reproduces exactly the fragment/query/scheme/authority parsing `ResourceId.fromURI` relies on, preserving all existing error messages (`Invalid URI`, `Invalid URI scheme`, `Invalid resource URI format`).
• Replaced `lorem-ipsum` with a local `generateSentence()` (1–5 words from a built‑in Latin word list) in `serializer/valuegenerator.ts`; `getString` min/max length behavior is unchanged.
• Dropped `urijs` and `lorem-ipsum` from `concerto-core` dependencies.

## Bundle impact (esm-browser)

• `urijs` chunk removed entirely (~94 KB raw / ~20 KB gzip).
• `lorem-ipsum` removed from the sample‑generator chunk (~54 KB → ~32 KB raw; the remainder is the retained `randexp`/`ret`/`drange`).
• Verified: `dist/esm-browser` now contains **zero** `urijs`/`lorem-ipsum` references; `randexp` remains.

## Flags

• No API or behavior change — the existing `resourceid`/`relationship`/`valuegenerator` test suites pass unchanged, including the tricky URI cases (ids containing `#`/`:`, non‑numeric port → `Invalid URI`, valid authority → `Invalid resource URI format`).
• Full `concerto-core` suite: **1283 passing**; coverage within thresholds (stmts 99.03%, branches 95.83%, funcs 99.33%, lines 99.04%). `tsc --noEmit` clean.

## Related Issues

• Related to #1319 (bundle‑size reduction for `concerto-core`) — complementary, unconditional dependency removal.

## Author Checklist

- [x] DCO sign‑off on all commits (`Signed-off-by`)
- [x] Conventional commit format (`refactor(concerto-core): …`)
- [x] Tests pass (`npm test`) and coverage thresholds met
- [x] No public API changes; no unintended whitespace changes

### DIFF
--- a/packages/concerto-core/src/model/resourceid.ts
+++ b/packages/concerto-core/src/model/resourceid.ts
@@ -12,11 +12,90 @@
  * limitations under the License.
  */
 
-import URIJS from 'urijs';
-
 import ModelUtils from '../modelutil';
 
 const RESOURCE_SCHEME = 'resource';
+
+/**
+ * Parse a URI into its component parts. Implements the subset of the
+ * generic URI parsing algorithm (RFC 3986) that ResourceId relies on:
+ * fragment, query, scheme, and authority (userinfo/host/port), leaving
+ * the remainder as the path.
+ * @param {String} uri - The URI to parse.
+ * @returns {Object} An object with protocol, username, password, port,
+ * query, fragment and path properties.
+ * @throws {Error} If the authority contains a non-numeric port.
+ * @private
+ */
+function parseUri(uri: string) {
+    let s = uri;
+    let fragment: string | null = null;
+    let query: string | null = null;
+    let protocol: string | null = null;
+    let username: string | null = null;
+    let password: string | null = null;
+    let port: string | null = null;
+
+    // fragment: split on the first '#'
+    const hashPos = s.indexOf('#');
+    if (hashPos > -1) {
+        fragment = s.substring(hashPos + 1) || null;
+        s = s.substring(0, hashPos);
+    }
+
+    // query: split on the first '?'
+    const qPos = s.indexOf('?');
+    if (qPos > -1) {
+        query = s.substring(qPos + 1);
+        s = s.substring(0, qPos);
+    }
+
+    // scheme: only recognised if the remainder does not start with '//'
+    if (s.substring(0, 2) !== '//') {
+        const colonPos = s.indexOf(':');
+        if (colonPos > -1) {
+            const candidate = s.substring(0, colonPos);
+            if (/^[a-z][a-z0-9.+-]*$/i.test(candidate)) {
+                protocol = candidate.toLowerCase();
+                s = s.substring(colonPos + 1);
+            }
+        }
+    }
+
+    // authority: only present if the remainder starts with '//'
+    if (s.substring(0, 2) === '//') {
+        s = s.substring(2);
+        const slashPos = s.indexOf('/');
+        const authority = slashPos > -1 ? s.substring(0, slashPos) : s;
+        s = slashPos > -1 ? s.substring(slashPos) : '';
+        let hostport = authority;
+        const atPos = authority.indexOf('@');
+        if (atPos > -1) {
+            const userinfo = authority.substring(0, atPos);
+            hostport = authority.substring(atPos + 1);
+            const uColon = userinfo.indexOf(':');
+            if (uColon > -1) {
+                username = userinfo.substring(0, uColon);
+                password = userinfo.substring(uColon + 1);
+            } else {
+                username = userinfo;
+            }
+        }
+        const pColon = hostport.lastIndexOf(':');
+        if (pColon > -1) {
+            const maybePort = hostport.substring(pColon + 1);
+            if (maybePort !== '') {
+                if (!/^[0-9]+$/.test(maybePort)) {
+                    throw new Error('Invalid port');
+                }
+                port = maybePort;
+            }
+        }
+    }
+
+    const path = s;
+    return { protocol, username, password, port, query, fragment, path };
+}
 
 /**
  * All the identifying properties of a resource.
@@ -72,7 +151,7 @@ class ResourceId {
     static fromURI(uri, legacyNamespace?, legacyType?) {
         let uriComponents;
         try {
-            uriComponents = URIJS.parse(uri);
+            uriComponents = parseUri(uri);
         } catch (err){
             throw new Error('Invalid URI: ' + uri);
         }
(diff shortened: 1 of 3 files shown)


packages/concerto-core/src/serializer/valuegenerator.ts
```diff
@@ -21,6 +20,23 @@ import dayjs from '../dayjs-setup';
...
+
+/**
+ * Generate a random lorem-ipsum-like sentence of one to five words.
+ * @return {string} a non-empty sentence.
+ * @private
+ */
+const generateSentence = () => {
+    const wordCount = Math.floor(Math.random() * 5) + 1;
+    const words: string[] = [];
+    for (let i = 0; i < wordCount; i++) {
```
jamieshorten, line 35: The old `loremIpsum({ units: 'sentences' })` returned sentences ending with a period (e.g. `"Lorem ipsum."`). `generateSentence()` doesn't append one.

This matters because `generateString` concatenates with `+=` (no separator). Before: `"Eu.Deserunt nostrud."` — ugly but word boundaries are visible. After: `"EuDeserunt nostrud"` — words at the join boundary merge into one token.

Adding a period to `generateSentence` would preserve the existing behavior:
```js
return sentence.charAt(0).toUpperCase() + sentence.substring(1) + '.';
```

packages/concerto-core/src/model/resourceid.ts
```diff
@@ -12,12 +12,91 @@
...
+    }
+
+    // scheme: only recognised if the remainder does not start with '//'
+    if (s.substring(0, 2) !== '//') {
+        const colonPos = s.indexOf(':');
+        if (colonPos > -1) {
+            const candidate = s.substring(0, colonPos);
+            if (/^[a-z][a-z0-9.+-]*$/i.test(candidate)) {
+                protocol = candidate.toLowerCase();
+                s = s.substring(colonPos + 1);
```
jamieshorten, line 60: Scheme detection skips if the string starts with `//`, which is correct for protocol-relative URIs. But consider `resource://host/path#id` — the `//` check fires first, so scheme detection is skipped entirely, and `resource` lands in the authority's userinfo/host parsing instead of being recognized as the protocol.

`urijs` handles this by extracting the scheme *before* checking for `//`:
```
resource://host/path → protocol: "resource", authority: "host", path: "/path"
```

The new parser would produce:
```
protocol: null, authority parsed from "resource://host" (wrong)
```

This doesn't affect `fromURI` today because `resource:` URIs don't use `//` — but the function's doc comment claims RFC 3986 subset compliance, and scheme-before-authority is the RFC's prescribed order. Swapping the two blocks (scheme first, then `//` authority) would fix it and is a one-line reorder.

jamieshorten (commented):
Clean PR — did an independent review. URI parser matches `urijs` on all the test-suite inputs I traced through (standard resource URIs, IDs with `#`/`:`, non-numeric ports, authority components). Two small things worth a look:

packages/concerto-core/src/serializer/valuegenerator.ts
```diff
@@ -113,22 +129,10 @@ const generateString = (seedString, minLength, maxLength, stringGenFunc) => {
...
+    let stringValue = generateSentence();
 
     if (minLength || maxLength) {
-        stringValue = generateString(stringValue, minLength, maxLength, () => loremIpsum({
-            count: 1                        // Number of words, sentences, or paragraphs to generate.
-            , units: 'sentences'            // Generate words, sentences, or paragraphs.
-            , sentenceLowerBound: lower         // Minimum words per sentence.
-            , sentenceUpperBound: upper         // Maximum words per sentence.
-        }));
+        stringValue = generateString(stringValue, minLength, maxLength, () => generateSentence());
```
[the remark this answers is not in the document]
mttrbrts: Good catch — fixed in `e96502f`. `generateSentence()` now appends a trailing period, so when `getString` concatenates multiple sentences to satisfy a length constraint the word boundaries stay visible (e.g. `"Lorem ipsum.Dolor sit."`), matching the prior `lorem-ipsum` output shape.

packages/concerto-core/src/serializer/valuegenerator.ts
mttrbrts: Agreed — applied exactly this in `e96502f`. Appended the trailing period to `generateSentence()` so `getString`'s `+=` padding no longer merges words at the join boundary. Thanks for the precise suggestion.

packages/concerto-core/src/model/resourceid.ts
mttrbrts: Thanks for the careful read — I looked into this and I believe the current ordering already handles it correctly. The `//` guard skips scheme detection only when the string *itself* starts with `//` (a protocol-relative URI that genuinely has no scheme). A `resource://host/path` string starts with `resource`, so the guard is false and the scheme block runs first, extracting `resource` and leaving `//host/path` for the authority block.

Verified against the exact `parseUri` logic:

```
resource://host/path#id => { protocol: 'resource', path: '/path', fragment: 'id' }
//host:8080/p           => { protocol: null, port: '8080', path: '/p' }
```

So scheme-before-authority (the RFC order) is preserved — the scheme block is physically first and runs for any string that isn't protocol-relative. The existing `resource://USER:PASSWORD@HOSTNAME:1567/...` test also depends on this: it only reaches the `Invalid resource URI format` branch because the scheme is stripped first and the remainder is then parsed into username/password/port.

If the doc comment read as implying the opposite order, I'm happy to reword it for clarity — let me know.

PR #1326, opened 2026-09-17
feat(concerto-core): expose structured validation details

### LINKED ISSUE #1325: Expose structured violation details on validation errors
## Use Case
When validating instances against Concerto models (e.g. during deserialization or validation), validation failures currently emit exception messages as formatted strings that can quote the submitted instance data.

In multi-tenant or API service environments (such as [accordproject/apap](https://github.com/accordproject/apap)), returning raw exception messages to callers can leak sensitive instance values that the caller's authorization scope should not see.

Downstream consumers are currently forced to resort to regex-parsing the error message strings to extract the violated property path and expected type while stripping the submitted values (see [accordproject/apap#253 (comment)](https://github.com/accordproject/apap/pull/253#discussion_r4005919474)):

```typescript
// Concerto violations arrive as a raw exception message that can quote the
// submitted instance, so only the JSON path and expected type are forwarded.
const CONCERTO_PATH_RX = /path `(\$[^\`]*)`/i;
const CONCERTO_TYPE_RX = /type `([^\`]+)`/i;

function sanitizeViolations(details?: Record<string, unknown>): Record<string, unknown> | undefined {
    const errors = details?.errors;
    if (!Array.isArray(errors)) return undefined;
    return {
        violations: errors.map((entry: any) => {
            const message = typeof entry?.message === 'string' ? entry.message : '';
            const path = CONCERTO_PATH_RX.exec(message)?.[1];
            const expectedType = CONCERTO_TYPE_RX.exec(message)?.[1];
            return { ...(path && { path }), ...(expectedType && { expectedType }) };
        }),
    };
}
```

## Possible Solution
Expose structured error/violation objects on validation exceptions (similar to Zod issues or JSON Schema errors):
- `path`: property path or JSON pointer (e.g., `$.address.city`)
- `expectedType`: expected type or constraint
- `code`: error code / category (e.g., `invalid_type`, `required`, `range`, `regex`)
- Separate structured metadata from the human-readable message so callers can inspect or forward validation facts without risking exposure of sensitive instance payloads.

## Context
Discussion in [accordproject/apap#253 (review comment)](https://github.com/accordproject/apap/pull/253#discussion_r4005919474).

Yyash64:
## Summary

Adds structured validation details to `ValidationException` so validation failures can expose machine-readable metadata without changing existing human-readable error messages.

## Changes

- Add optional `details` to `ValidationException`
- Add structured `path`, `code`, and `expected` details to relevant `JSONPopulator` validation failures
- Add structured details to relevant `ResourceValidator` validation failures
- Add tests covering structured validation details

## Compatibility

Existing validation error messages are unchanged.

## Testing

- `1302 passing`
- `8 pending`
- `git diff --check` passes

Closes #1325


Yyash64:
Hi @mon7792, Copilot reviewed #1326 and raised a few points around structured metadata, particularly propagating nested/indexed paths through ResourceValidator and adding paths to relationship/type-assignment violations. These seem potentially broader than the focused #1325 change.

Before I address those, I'd appreciate your thoughts on whether we should extend the scope to cover full nested/indexed paths now, or keep that aligned with the broader validation API work from #1239/#1240.

mon7792:
Hi @Yyash64 ,

Here are my thoughts:

From a correctness and ambiguity perspective, I think preserving the **_full nested/indexed path_** would be preferable, since it gives much more precise context about where the validation issue occurred.

That said, I’d also suggest looking at this from an agent-consumption perspective. If these validation errors are going to be consumed by coding agents, the key question is: what metadata gives the agent the most accurate and actionable context for identifying and fixing the issue?

`suggestion`: _only do this is you find any value_
One possible experiment would be to create a small CLI program with a Zod validator, introduce a few validation issues in nested input data, and then use a coding agent to diagnose them. Comparing the results with and without full nested/indexed paths could give us a practical signal on how useful this metadata is.

So overall, I’m in favor of nested paths from a correctness perspective. your call tough.

Yyash64:
Hi @mon7792, thanks for the detailed feedback. I agree that preserving the full nested/indexed path makes the structured metadata significantly more actionable, especially for agent consumption.

I’ll extend the current implementation to propagate the full validation path through nested/indexed values, while keeping the scope focused on structured validation details rather than pulling in the broader #1239/#1240 API.

I’ll also cover the relevant nested/indexed cases with regression tests. I’ll treat the Zod/agent experiment as optional and only pursue it if it provides useful signal during implementation.

Yyash64:
Hi @mon7792, I’ve pushed the follow-up changes in 79686ec.

The ResourceValidator now propagates full nested/indexed paths through validation, including relationship and type-assignment violations. I also added a regression test covering an indexed nested path (`$.containment[0]`).

The full concerto-core test suite passes locally: 1303 passing, 8 pending.

The remaining CI checks are currently awaiting maintainer approval for workflows from the fork.

mttrbrts:
## pr-autopilot review

This PR threads a path parameter through ResourceValidator/JSONPopulator/ValidatedResource and adds a structured details object (code, path, expected) to ValidationException, building on the prior commit that introduced structured details. The change is additive and backward compatible (new optional constructor param, new optional field), and is covered by new unit tests for both the date-type and nested-array-path cases.

A couple of gaps remain in the path-tracking: relationship-typed array fields don't get per-item [n] path segments the way plain object/enum arrays do, so violations inside a relationship array will report the parent field's path rather than the specific index. reportEmptyIdentifier also always reports path: '$' regardless of how deeply nested the resource with the empty identifier is. Neither breaks existing behavior since this is new functionality, but both undercut the "preserve full validation paths" goal for those two cases. There are also a couple of indentation inconsistencies in the new throw blocks that may trip the repo's linter.

**Risk breakdown**

- Blast radius: 0.70 (confidence 0.75)
- Security: 0.23 (confidence 0.69)
- API compatibility: 0.34 (confidence 0.99)
- Test adequacy: 0.73 (confidence 0.79)
- Description quality: 0.70 (confidence 0.85)
- Composite risk: 39.6/100 (medium)

**Decision:** needs-review -- gates failed: not_first_time_contributor: author association 'FIRST_TIME_CONTRIBUTOR' is not trusted; no_unresolved_threads: unresolved review threads; risk_band: risk band medium above low for kind feature; review_approves: verdict comment

---
_Automated review by pr-autopilot (kind: feature, policy: 2026.09.23-3+70e377528a92). This is not a substitute for human review._

Yyash64:
Thanks @mttrbrts. Addressed the two remaining path-tracking gaps:
Relationship arrays now include [index] in the path (e.g. $.owners[0]).
reportEmptyIdentifier now uses the current traversal path instead of hardcoding $.

Also threaded path into reportUndeclaredField, and fixed the indentation in the relationship loop. Added regression tests for both new cases.

Local suite: 1305 passing, 8 pending. Could you approve the awaiting workflows and take another look when you have a chance?

